PR #26, opened 2020-10-09
Refactor(cli): Review of initial version

ee7:
This PR is a review of the initial CLI implementation. The main things are:
- Allow a space between an option and its value - previously, the `--help` message showed incorrect syntax.
- Show error messages for unrecognised command-line input, rather than silently ignoring.
- Refactor internally to make things more idiomatic.

The commits are atomic - please review individual commits and see the commit messages.

And when we have this PR in an acceptable state, please merge without squashing so that we preserve the commit message granularity, and so that `git` can track the file rename.

Fixes: #30 
Fixes: #34
Fixes: #35
Fixes: #36
Fixes: #37


README.md
```diff
@@ -11,11 +11,11 @@ Usage: canonical_data_syncer [options]
 
 Options:
   -e, --exercise <slug>        Only sync this exercise
-  -c, --check                  Check if there missing tests. Doesn't update the tests. Terminates with a non-zero exit code if any one or more test cases are missing.
-  -d, --default <mode>         What to do with missing test cases. Allowed values: c[hoose], i[nclude], e[xclude]
+  -c, --check                  Check if there are missing tests. Doesn't update the tests. Terminates with a non-zero exit code if one or more tests are missing
```
ErikSchierboom, line 14: Good catch!

README.md
```diff
@@ -11,11 +11,11 @@ Usage: canonical_data_syncer [options]
...
+  -m, --mode <mode>            What to do with missing test cases. Allowed values: c[hoose], i[nclude], e[xclude]
```
ErikSchierboom, line 15: I've gone back and forth on this. Maybe we should add something that explains that setting this means things are handled automatically, whereas not setting it requires interaction.

src/cli.nim
```diff
@@ -0,0 +1,149 @@
...
+    optExercise, optCheck, optMode, optVerbosity, optHelp, optVersion
+
+  OptKey = tuple
+    short: string
+    long: string
+
+const
+  NimblePkgVersion {.strdefine.}: string = "unknown"
+
+  optKeys: array[Opt, OptKey] = [
```
ErikSchierboom, line 29: Ooh, I didn't know you could an `enum` as the size of the array!

src/cli.nim
```diff
@@ -0,0 +1,149 @@
...
+    ("c", "check"),
+    ("m", "mode"),
+    ("o", "verbosity"),
+    ("h", "help"),
+    ("v", "version"),
+  ]
+
+  optsNoVal = {optCheck, optHelp, optVersion}
+
+func short(opt: Opt): string =
```
ErikSchierboom, line 40: When should a `func` be used over ` proc`? I think I remember it being about side-effects.

src/cli.nim
```diff
@@ -46,6 +46,18 @@ proc showVersion =
   echo &"Canonical Data Syncer v{NimblePkgVersion}"
   quit(0)
 
+proc showError(s: string) =
```
ErikSchierboom, line 65: I don't know if this idiomatic, but I would normally name the parameter `error` or `errorMessage` or something like that.

src/cli.nim
```diff
@@ -58,6 +58,11 @@ proc prefix(kind: CmdLineKind): string =
   of cmdLongOption: "--"
   of cmdArgument, cmdEnd: ""
 
+proc showErrorForMissingVal(kind: CmdLineKind, key: string, val: string) =
```
ErikSchierboom, line 77: Maybe update the name a bit to indicate that the error is shown conditionally?

```suggestion
proc showErrorIfMissingVal(kind: CmdLineKind, key: string, val: string) =
```

ErikSchierboom (approved):
Thanks so much!

src/cli.nim
ee7: (I've now removed this proc).

src/cli.nim
ee7: Yes - we can use `func` instead of `proc` whenever the proc doesn't have side effects. In my opinion, it increases readability if we consisently use `func` whenever possible in a codebase. I'll eventually do that in this repo as I refactor the code.

See also:
- https://nim-lang.org/docs/manual.html#pragmas-nosideeffect-pragma
- https://nim-lang.org/blog/2020/09/01/write-tracking.html

Issue #45, opened 2020-10-16
CLI: syntax given in --help is not recognized


coriolinus:
```bash
$ bin/canonical_data_syncer -m c -e sum-of-multiple
Fri 16 Oct 2020 11:35:43 AM CEST
Error: '-m' was given without a value

Usage: canonical_data_syncer [options]

Options:
  -e, --exercise <slug>        Only sync this exercise
  -c, --check                  Check if there are missing tests. Doesn't update the tests. Terminates with a non-zero exit code if one or more tests are missing
  -m, --mode <mode>            What to do with missing test cases. Allowed values: c[hoose], i[nclude], e[xclude]
  -o, --verbosity <verbosity>  The verbosity of output. Allowed values: q[uiet], n[ormal], d[etailed]
  -h, --help                   Show this help message and exit
  -v, --version                Show this tool's version information and exit
```

The CLI option parser appears to require an `=` sign between the flag and its value:

```bash
$  bin/canonical_data_syncer -m=c -e=sum-of-multiple
Fri 16 Oct 2020 11:40:21 AM CEST
Syncing exercises...
Cloning into '/mnt/c/Users/coriolinus/Documents/Projects/exercism/xrust/.problem-specifications'...
remote: Enumerating objects: 540, done.
remote: Counting objects: 100% (540/540), done.
remote: Compressing objects: 100% (512/512), done.
remote: Total 540 (delta 27), reused 149 (delta 17), pack-reused 0
Receiving objects: 100% (540/540), 247.13 KiB | 605.00 KiB/s, done.
Resolving deltas: 100% (27/27), done.
All exercises are synced!
```

ee7:
Thanks. We should clean this up, or at least document it better.

Currently, I believe the situation is:
- long options can be separated by a (space/equals/colon)
- short options can only be separated by (space/colon/nothing)

That is, these work:
```
./canonical_data_syncer --mode choose 
```

```
./canonical_data_syncer --mode c
```

```
./canonical_data_syncer -mc
```

But these don't:
```
./canonical_data_syncer -m choose
```

```
./canonical_data_syncer -m c
```

ee7:
I will implement the fix for this issue.

ErikSchierboom:
[Version v0.12.0](https://github.com/exercism/canonical-data-syncer/releases/tag/v0.12.0) has been released today and it includes this fix. Thanks for reporting @coriolinus and for fixing it @ee7!

PR #54, opened 2020-10-19
Fix(cli): Support the syntax of e.g. `-m include`

ee7:
This PR adds support for separating a short option key from its value with a space (which the help message implied was possible). For example, running either of:
```
    canonical_data_syncer -m include
    canonical_data_syncer -m i
```
would previously produce an error:
```
    Error: '-m' was given without a value
```

This is because the `parseopt` module in Nim's standard library only supports separating a short option key from its value with an equals sign, a colon, or nothing. That is, only these work:
```
    canonical_data_syncer -m=include
    canonical_data_syncer -m:include
    canonical_data_syncer -m=i
    canonical_data_syncer -m:i
    canonical_data_syncer -mi
```

But a long option key and its value _can_ be separated by a space (or an equals sign, or a colon). For example:
```
    canonical_data_syncer --mode=include
    canonical_data_syncer --mode:include
    canonical_data_syncer --mode include
    canonical_data_syncer --mode=i
    canonical_data_syncer --mode:i
    canonical_data_syncer --mode i
```

The simplest fix is to use cligen's `parseopt3` module, which is designed to be API-compatible with `parseopt`. `cligen` is well-known and mature. Now we can support all of the above syntax, at the cost of:
- One extra dependency at compile-time only.
- Increasing the release binary size by about 2% (the Linux release binary is exactly 4 KiB bigger - see [here](https://github.com/ee7/canonical-data-syncer/releases/tag/release-fix-cli-use-parseopt3) for the releases on my fork, using a branch that releases on every push, not a tag).

Fixes: #45

### DIFF
--- a/canonical_data_syncer.nimble
+++ b/canonical_data_syncer.nimble
@@ -11,3 +11,4 @@ backend       = "c"
 # Dependencies
 requires "nim >= 1.2.6"
 requires "parsetoml"
+requires "cligen"
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -1,4 +1,5 @@
-import std/[options, os, parseopt, strformat, strutils, terminal]
+import std/[options, os, strformat, strutils, terminal]
+import pkg/[cligen/parseopt3]
 
 type
   Action* = enum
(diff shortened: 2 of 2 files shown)


ErikSchierboom (approved):
Great! I don't mind the little size increase, it is already a very small tool anyways. I do wish we would have tests to verify this behavior at some point.

PR #59, opened 2020-10-19
Refactor(check): Remove duplicate loop

ee7:
This is more efficient, and allows removal of a `sequtils` import.


ErikSchierboom (approved):
I do wish we had some tests at this point :D

Issue #52, opened 2020-10-17
Make git clone still show up


SleeplessByte:
#49 is great, but it completely removes the git related info messages. Whenever the internet connection is slow or flaky, this looks as if the terminal is hanging.

I recommend adding a single line "checking out <repo> into temporary location" or something like that, just so you know it's not broken.

PR #49, opened 2020-10-16
Make git clone quiet

glennj:
Resolving a personal annoyance: get rid of the `git clone` output.

* before this change
    ```bash
    $ ../canonical-data-syncer/canonical_data_syncer -c
    Checking exercises...
    Cloning into '/Users/glennjackman/src/exercism/tcl/.problem-specifications'...
    remote: Enumerating objects: 540, done.
    remote: Counting objects: 100% (540/540), done.
    remote: Compressing objects: 100% (512/512), done.
    remote: Total 540 (delta 27), reused 149 (delta 17), pack-reused 0
    Receiving objects: 100% (540/540), 247.16 KiB | 2.06 MiB/s, done.
    Resolving deltas: 100% (27/27), done.
    [warn] list-ops: missing 1 test cases
    [warn] some exercises are missing test cases
    ```

* after
    ```bash
    $ ../canonical-data-syncer/canonical_data_syncer -c
    Checking exercises...
    [warn] list-ops: missing 1 test cases
    [warn] some exercises are missing test cases

    $ ../canonical-data-syncer/canonical_data_syncer -c -od
    Checking exercises...
    Taking a shallow clone of problem-specifications repo into /Users/glennjackman/src/exercism/tcl/.problem-specifications
    [skip] accumulate: does not have canonical data
    ...
    ```

PR #60, opened 2020-10-19
Fix: Show minimal `git clone` output by default

ee7:
PR #49 caused the program to produce no output during the `git clone` at the default verbosity setting. It was nice to remove the excessive `git clone` output, but total removal means that the program can appear to hang on a poor internet connection.

This commit makes the situation a little better for that case. In the future, we can try to improve the behaviour further.

Fixes: #52


ErikSchierboom (approved):

ErikSchierboom:
Thanks

PR #65, opened 2020-10-21
Feature: Add `--probSpecsDir` option

ee7:
See the commit messages for more information - the commits are atomic.

Some things like the `git` error messages could probably be cleaned up a bit, and there's some obvious tests to add, but it should be reasonably robust. I think the implementation here is quite strict. Let me know if I missed any edge cases.

WIP:
- [X] Don't `checkout` or `merge`, even if the working directory is clean.
- [X] Try not to assume things about the remotes in the given directory.

Closes: https://github.com/exercism/canonical-data-syncer/issues/47

### DIFF
--- a/src/probspecs.nim
+++ b/src/probspecs.nim
@@ -30,16 +30,6 @@ proc clone(repo: ProbSpecsRepo) =
   logNormal(&"Cloning the problem-specifications repo into {repo.dir}...")
   execCmdException(cmd, "Could not clone problem-specifications repo")
 
-proc grainsWorkaround(repo: ProbSpecsRepo) =
-  ## Overwrites the canonical data file for `grains` so that it no longer
-  ## contains integers that are too large to store in a 64-bit signed integer.
-  ## Otherwise, we get an error when parsing it as JSON.
-  let grainsPath = repo.dir / "exercises" / "grains" / "canonical-data.json"
-  let s = readFile(grainsPath).multiReplace(
-    ("92233720368547758", "92233720368547758.0"),
-    ("184467440737095516", "184467440737095516.0"))
-  writeFile(grainsPath, s)
-
 proc remove(repo: ProbSpecsRepo) =
   removeDir(repo.dir)
 
@@ -88,8 +78,20 @@ proc initProbSpecsTestCases(node: JsonNode): seq[ProbSpecsTestCase] =
     for childNode in node["cases"].getElems():
       result.add(initProbSpecsTestCases(childNode))
 
+proc grainsWorkaround(grainsPath: string): JsonNode =
+  ## Parses the canonical data file for `grains`, replacing the too-large
+  ## integers with floats. This avoids an error that otherwise occurs when
+  ## parsing integers are too large to store as a 64-bit signed integer.
+  let sanitised = readFile(grainsPath).multiReplace(
+    ("92233720368547758", "92233720368547758.0"),
+    ("184467440737095516", "184467440737095516.0"))
+  result = parseJson(sanitised)
+
 proc parseProbSpecsTestCases(repoExercise: ProbSpecsRepoExercise): seq[ProbSpecsTestCase] =
-  initProbSpecsTestCases(json.parseFile(repoExercise.canonicalDataFile))
+  if repoExercise.slug == "grains":
+    repoExercise.canonicalDataFile().grainsWorkaround().initProbSpecsTestCases()
+  else:
+    repoExercise.canonicalDataFile().parseFile().initProbSpecsTestCases()
 
 proc initProbSpecsExercise(repoExercise: ProbSpecsRepoExercise): ProbSpecsExercise =
   result.slug = repoExercise.slug
@@ -106,7 +108,6 @@ proc findProbSpecsExercises*(conf: Conf): seq[ProbSpecsExercise] =
   try:
     probSpecsRepo.remove()
     probSpecsRepo.clone()
-    probSpecsRepo.grainsWorkaround()
     probSpecsRepo.findProbSpecsExercises(conf)
   finally:
     probSpecsRepo.remove()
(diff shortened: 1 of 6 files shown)


tests/test_probspecs.nim
```diff
@@ -0,0 +1,61 @@
+# This module contains tests for `src/probspecs.nim`
+import std/[json, options, os, osproc, strformat, unittest]
+import cli, probspecs
+
+type
+  ProblemSpecsDir = enum
+    psFresh = "fresh clone"
+    psExisting = "existing dir (simulate the `--probSpecsDir` option)"
+
+proc main =
```
ErikSchierboom, line 10: So good to finally have some tests!

src/probspecs.nim
```diff
@@ -100,13 +102,70 @@ proc findProbSpecsExercises(repo: ProbSpecsRepo, conf: Conf): seq[ProbSpecsExerc
...
+      # upstream, even if it's possible to do a fast-forward merge.
+      if execCmd(&"git fetch --quiet {remoteName} {mainBranchName}") != 0:
+        showError(&"failed to fetch `{mainBranchName}` in problem-specifications directory: '{probSpecsDir}'")
+
+      # Allow HEAD being on a non-`master` branch, as long as it's up-to-date
+      # with `upstream/master`.
+      let (revHead, _) = execCmdEx("git rev-parse HEAD")
+      let (revUpstream, _) = execCmdEx(&"git rev-parse {remoteName}/{mainBranchName}")
+      if revHead != revUpstream:
+        showError(&"the given problem-specifications directory is not up-to-date: '{probSpecsDir}'")
```
ErikSchierboom, line 161: Nice! I would suggest moving this to a separate function, as the verification code now slightly obscures the flow of the function

src/probspecs.nim
```diff
@@ -100,13 +102,70 @@ proc findProbSpecsExercises(repo: ProbSpecsRepo, conf: Conf): seq[ProbSpecsExerc
...
+
+    withDir probSpecsDir:
+      if execCmd("git rev-parse") != 0:
+        showError(&"the given problem-specifications directory is not a git repository: '{probSpecsDir}'")
+
+      # Exit if the working directory is not clean.
+      if execCmd("git diff-index --quiet HEAD") != 0: # Ignores untracked files.
+        echo &"\nUnstaged changes in {probSpecsDir}:"
+        discard execCmd("git status --short --untracked=no")
+        showError(&"the given problem-specifications working directory is not clean: '{probSpecsDir}'")
```
ErikSchierboom, line 132: Maybe output this the other way around?

```suggestion
        showError(&"the given problem-specifications working directory has unstaged changes: '{probSpecsDir}'")
        discard execCmd("git status --short --untracked=no")
```

ErikSchierboom (approved):
Some minor comments! Excellent work.

src/probspecs.nim
ee7: Doesn't work, unfortunately. `showError` isn't the best name, but it causes an exit.

But I agree that it might be better to see the `git status` output after the error, rather than before.

ErikSchierboom: An alternative would be to not display the unstaged changes.

tests/test_probspecs.nim
ee7: Agreed.

src/probspecs.nim
ee7: Good idea. Done.

I also moved the remote-finding into its own function, but didn't do that for the rest of the validating. It think it might be better to put it into a new module if we wanted to do that.

src/probspecs.nim
ee7: Done.

src/probspecs.nim
```diff
@@ -100,13 +102,87 @@ proc findProbSpecsExercises(repo: ProbSpecsRepo, conf: Conf): seq[ProbSpecsExerc
...
+    # Exit if the given directory is not a git repo.
+    if execCmd("git rev-parse") != 0:
+      showError("the given problem-specifications directory is not a git " &
+                &"repository: '{probSpecsDir}'")
+
+    # Exit if the working directory is not clean.
+    if execCmd("git diff-index --quiet HEAD") != 0: # Ignores untracked files.
+      echo &"\nUnstaged changes in {probSpecsDir}:"
+      showError("the given problem-specifications working directory is not " &
+                &"clean: '{probSpecsDir}'")
```
ErikSchierboom, line 154:
```suggestion
      showError("the given problem-specifications working directory has" &
                &"unstaged changes: '{probSpecsDir}'")
```

src/probspecs.nim
```diff
@@ -100,13 +102,87 @@ proc findProbSpecsExercises(repo: ProbSpecsRepo, conf: Conf): seq[ProbSpecsExerc
...
+    let (revHead, _) = execCmdEx("git rev-parse HEAD")
+    let (revUpstream, _) = execCmdEx(&"git rev-parse {remoteName}/{mainBranchName}")
+    if revHead != revUpstream:
+      showError("the given problem-specifications directory is not " &
+                &"up-to-date: '{probSpecsDir}'")
+
+proc findProbSpecsExercises*(conf: Conf): seq[ProbSpecsExercise] =
+  if conf.probSpecsDir.isSome():
+    let probSpecsRepo = ProbSpecsRepo(dir: conf.probSpecsDir.get())
+    validateProbSpecsRepo(probSpecsRepo)
```
ErikSchierboom, line 179: The other calls in this function use `probSpecsRepo.<function>()`. Was there any particular reason not to do that here?

```suggestion
    probSpecsRepo.validateProbSpecsRepo()
```

Mayb even do:

```suggestion
    probSpecsRepo.validate()
```

ErikSchierboom (approved):

src/probspecs.nim
ee7: Done, thanks.

src/probspecs.nim
ee7: Thanks. Fixed.

Issue #50, opened 2020-10-16
Allow specifying the track repository directory


ErikSchierboom:
We should allow the user to specify the directory of the track to work with. It should default to the current directory if not specified. I'm thinking it should just be the last, non-named parameter.

The `configlet` tool has a similar feature BTW.

ee7:
(I've implemented this - will PR after #65 is merged and the PR for #66 is merged).

PR #72, opened 2020-10-23
Feature: Add `--trackDir` option

ee7:
With this PR, we can now establish a beginning state by specifying:
- a track directory
- a problem-specifications directory
- the offline mode

This allows us to start doing black-box testing of the release binary, running something like:
```
    configlet sync --track-dir /tmp/python --prob-specs-dir /tmp/my_prob_specs --offline
```
and then asserting that the changes made are as expected.

Closes: #50

WIP: 
- [X] Rebase this PR on `master` and reflect recent changes (e.g. `--trackDir` -> `--track-dir`)
- [X] In `sync` mode, make it write to the given `-t, --trackDir`, rather than relative to the current directory.
- [X] Merge #134 and rebase this PR on top.
- [x] Address review comments

Let's add tests in a later PR. We can leave `-t, --track-dir` undocumented in the first `configlet` releases if we feel it isn't tested well enough. 

Note that `--track-dir` is implemented as a global option, not an option for `sync` - the option will also be useful for e.g. `lint`.

This feature makes it much more ergonomic to implement tests for `lint`.

### DIFF
--- a/src/sync/sync.nim
+++ b/src/sync/sync.nim
@@ -64,9 +64,10 @@ proc syncDecision(testCase: ExerciseTestCase, mode: Mode): SyncDecision =
   of modeChoose:
     chooseSyncDecision(testCase)
 
-proc sync(exercise: Exercise, mode: Mode): Exercise =
+proc sync(exercise: Exercise, conf: Conf): Exercise =
   result = exercise
 
+  let mode = conf.action.mode
   case mode
   of modeInclude:
     logNormal(&"[info] {exercise.slug}: included {exercise.tests.missing.len} missing test cases")
@@ -100,13 +101,13 @@ proc sync(exercise: Exercise, mode: Mode): Exercise =
 
   result.tests = initExerciseTests(included, excluded, missing)
 
-  writeTestsToml(result)
+  writeTestsToml(result, conf.trackDir)
 
-proc sync(exercises: seq[Exercise], mode: Mode): seq[Exercise] =
+proc sync(exercises: seq[Exercise], conf: Conf): seq[Exercise] =
   for exercise in exercises:
     case exercise.status
     of exOutOfSync:
-      result.add(sync(exercise, mode))
+      result.add(sync(exercise, conf))
     of exInSync:
       logDetailed(&"[skip] {exercise.slug} is up-to-date")
     of exNoCanonicalData:
@@ -116,7 +117,7 @@ proc sync*(conf: Conf) =
   logNormal("Syncing exercises...")
 
   let exercises = findExercises(conf)
-  let syncedExercises = sync(exercises, conf.action.mode)
+  let syncedExercises = sync(exercises, conf)
 
   if syncedExercises.anyIt(it.status == exOutOfSync):
     logNormal("[warn] some exercises are still missing test cases")
--- a/src/sync/tracks.nim
+++ b/src/sync/tracks.nim
@@ -83,5 +83,9 @@ proc findTrackExercises(repo: TrackRepo, conf: Conf): seq[TrackExercise] =
       result.add(newTrackExercise(repoExercise))
 
 proc findTrackExercises*(conf: Conf): seq[TrackExercise] =
-  let trackRepo = newTrackRepo()
+  let trackRepo =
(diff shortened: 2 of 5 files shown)


ErikSchierboom (approved):
Great! Approving pending the fix for the writing of the tests.toml files relative to the track dir.

src/sync/sync.nim
```diff
@@ -100,13 +101,13 @@ proc sync(exercise: Exercise, mode: Mode): Exercise =
 
   result.tests = initExerciseTests(included, excluded, missing)
 
-  writeFile(result)
+  writeTestsToml(result, conf.trackDir)
```
ErikSchierboom, line 104: Is `conf.trackDir` set to the current directory if the `--track-dir` argument is not specified?

src/sync/tracks.nim
```diff
@@ -83,5 +83,9 @@ proc findTrackExercises(repo: TrackRepo, conf: Conf): seq[TrackExercise] =
       result.add(newTrackExercise(repoExercise))
 
 proc findTrackExercises*(conf: Conf): seq[TrackExercise] =
-  let trackRepo = newTrackRepo()
+  let trackRepo =
+    if conf.trackDir.len > 0:
+      TrackRepo(dir: conf.trackDir)
+    else:
+      newTrackRepo()
```
ErikSchierboom, line 90: I'm wondering if it would perhaps be easier to set `conf.trackDir` to the current directory if the `--track-dir` argument is specified.

src/sync/sync.nim
ee7: It was set to the empty string, which ends up meaning the current directory.

I pushed a new commit, do you prefer it?

src/sync/tracks.nim
ee7: > set `conf.trackDir` to the current directory if the `--track-dir` argument is specified.

Do you mean "set the current directory to `conf.trackDir` if `--track-dir` is specified"?

We could indeed change the current directory in the program somewhere, but I think that might be harder to debug - it's more implicit.

We could use the `withDir` template, but it might be easier to just pass `conf` or `conf.trackDir` around for now. In the long-term I want to do a refactoring pass on `exercises.nim`, `tracks.nim`, and `sync.nim` that should help some of these issues. I have an old branch for that, but I never finished it.

src/sync/tracks.nim
```diff
@@ -24,9 +24,6 @@ type
     tests*: TrackExerciseTests
     repoExercise: TrackRepoExercise
 
-proc newTrackRepo: TrackRepo =
-  result.dir = getCurrentDir()
-
```
ee7, line 29: I made this change partly because I find it more readable to inline it in `findTrackExercises`. 

But there's also a general issue that I'll try to fix everywhere eventually: I believe that procs that initialize non-`ref` objects should be named starting with `init`, not `new` - even if they contain a field of a type like `string`. See https://nim-lang.github.io/Nim/nep1.html#introduction-naming-conventions

And in my refactoring branch I think I changed a lot of this stuff to use `distinct string` rather than an one-field object anyway.

ErikSchierboom (approved):
LGTM! The only thing I found was that when using `--track-dir`, the `prob-specs` repo gets cloned to the current directory instead of the track's dir. We can fix that later.

ee7:
> when using `--track-dir`, the prob-specs repo gets cloned to the current directory instead of the track's dir. We can fix that later.

Yes, currently the temporary `.problem-specifications` folder is created in the current directory rather than the track dir. But that would only be visible to a user that does a keyboard interrupt or encounters an error, right?

 We should be careful that everything behaves correctly when both options are provided, regardless of the order:
- `./configlet sync -t mytrack -p myprobspecs`
- `./configlet sync -p myprobspecs -t mytrack`
- `./configlet -t mytrack sync -p myprobspecs`

ErikSchierboom:
> Yes, currently the temporary .problem-specifications folder is created in the current directory rather than the track dir. But that would only be visible to a user that does a keyboard interrupt or encounters an error, right?

Correct. It would be mostly for the error case that I would expect it to be located in the track dir.

PR #78, opened 2020-10-23
--probSpecsDir option: Cope with ssh-cloned repo

glennj:
Repo cloned with ssh does not match `github.com/exercism`
```bash
$ cd ../problem-specifications/
$ git remote -v
origin	git@github.com:exercism/problem-specifications.git (fetch)
origin	git@github.com:exercism/problem-specifications.git (push)
```

### DIFF
--- a/src/probspecs.nim
+++ b/src/probspecs.nim
@@ -111,7 +111,7 @@ template withDir(dir: string; body: untyped): untyped =
   finally:
     setCurrentDir(startDir)
 
-proc getNameOfRemote(probSpecsDir, location: string): string =
+proc getNameOfRemote(probSpecsDir, host, location: string): string =
   ## Returns the name of the remote in `probSpecsDir` that points to `location`.
   ##
   ## Raises an error if there is no remote that points to `location`.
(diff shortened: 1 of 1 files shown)


ee7 (commented):
Thanks.

We should update the doc comment (and probably the error message) to reflect that `location` no longer contains `github.com`, but otherwise I think this PR should be merged.

I'd like to improve the `scanf` pattern instead of using two `contains`, but I can do that in another PR. I see the possibilities as:
1. hardcode `github.com` (removing `host` as a parameter) or
2. pass `host` as a `static string` and then write something like:

```nim
  var remoteName, remoteUrl, x: string # `x` to ignore http(s)/git/www.
  for line in remotes.splitLines():
    if line.scanf(&"$s$w$s$+{host}$+(fetch)$.", remoteName, x, remoteUrl):
      if remoteUrl.continuesWith(location, start = 1): # ignore ':' and '/'
        return remoteName
```

ErikSchierboom (approved):
I've verified that it still works for HTTPS repos. The change looks looks to me. I'll wait for the approve of @ee7 to merge.

ee7 (approved):

ErikSchierboom:
Thanks @glennj!

Issue #79, opened 2020-10-26
Fix(cli): Make options insensitive to case, '-' and '_'


ee7:
In https://github.com/exercism/canonical-data-syncer/issues/76#issuecomment-715481977 @ee7 wrote:
> I was intending to eventually make it style-insensitive anyway. So any of these should be valid:

> ```
> --probSpecsDir
> --probspecsdir
> --prob-specs-dir
> --prob_specs_dir
> --PROBSPECSDIR
> --prO_b----sPe____CsDIR
> ```

This will allow `--prob-specs-dir` even if it isn't shown in the `--help` like that yet.

I'll do this soon.

PR #84, opened 2020-10-26
Fix(cli): Parse options insensitive to style

ee7:
Most importantly, these previously invalid options are now accepted:
```
    --probspecsdir
    --prob-specs-dir
    --prob_specs_dir
```
Before this PR, we could only use:
```
    --probSpecsDir
```

That is: `--kebab-case` is now supported, even if that fact or `--prob-specs-dir` itself aren't shown in the `--help` message yet.

Note that we already had case-insensitive parsing for:
  - the values for `mode`
  - the values for `verbosity`
  - arguments (we support `canonical_data_syncer HELP`)

Fixes: #79

### DIFF
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -117,6 +117,20 @@ proc initConf: Conf =
     verbosity: verNormal,
   )
 
+func normalizeOption(s: string): string =
+  ## Returns the string `s`, but converted to lowercase and without '_' or '-'.
+  result = newString(s.len)
+  var i = 0
+  for c in s:
+    if c in {'A'..'Z'}:
+      result[i] = toLowerAscii(c)
+      inc i
+    elif c notin {'_', '-'}:
+      result[i] = c
+      inc i
+  if i != s.len:
+    setLen(result, i)
+
 proc processCmdLine*: Conf =
   result = initConf()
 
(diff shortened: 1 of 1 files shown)


ee7:
Changing the help message's `--probSpecsDir` to `--prob-specs-dir` needs to wait for a later PR, after some refactoring.

It'll make more sense later, but I'm deliberately using `.toLowerAscii` instead of changing the string value of the `probSpecsDir` enum to `"prob-specs-dir"` or `"probspecsdir"`. It's better that it doesn't contain hyphens, and by keeping it `camelCase` I can reduce duplication and generate the help text in a later PR by using a `camelCase` -> `kebab-case` proc.

src/cli.nim
```diff
@@ -141,7 +155,7 @@ proc processCmdLine*: Conf =
       of optVerbosity.short, optVerbosity.long:
         showErrorForMissingVal(kind, key, val)
         result.verbosity = parseVerbosity(kind, key, val)
-      of optProbSpecsDir.short, optProbSpecsDir.long:
+      of optProbSpecsDir.short, optProbSpecsDir.long.toLowerAscii():
```
ErikSchierboom, line 158: After reading your reasoning I understand the reason behind this, but I found it slightly hard to understand. Let's look at the best way to do this later.

ErikSchierboom (approved):
As a side note: it would be great to have unit tests to verify this.

src/cli.nim
ee7: To others: it was just a temporary workaround until https://github.com/exercism/canonical-data-syncer/pull/86 was merged.

The current implementation on `master` is:
https://github.com/exercism/canonical-data-syncer/blob/ae310285943fb9c404f522ad27a78fc592f629e5/src/cli.nim#L185-L188

PR #85, opened 2020-10-26
Fix(cli): Align long options in help message

ee7:
I think this looks nicer, and a lot of my CLI tools do it too.


ErikSchierboom (approved):
I like this version better too!

PR #86, opened 2020-10-26
Refactor(cli): Reduce duplication

ee7:
~To-do~
- [x] Merge #84 (the commit is included in this PR)
- [x] Merge #85 (the commit is included in this PR)
- [x] Rebase this PR on `master`

This PR reduces the code duplication in `cli.nim`.

The commits are atomic - please review individually. But they can be squashed before merging, if we prefer, given that they work towards the same goal.

Motivations:
- We don't need separate `parseMode` and `parseVerbosity` functions -
  we can just have a generic `parseValue` proc, which is easily
  extended if we ever add more options that have values.
- There should be a single source of truth for the string values of
  options/values.
- This refactoring makes it easier to add support for unambiguous
  abbreviations of options.
- We don't need to specify all the short options, just the short options
  that we don't want to have.

### DIFF
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -6,10 +6,14 @@ type
     actSync, actCheck
 
   Mode* = enum
-    modeChoose, modeInclude, modeExclude
+    modeChoose = "choose"
+    modeInclude = "include"
+    modeExclude = "exclude"
 
   Verbosity* = enum
-    verQuiet, verNormal, verDetailed
+    verQuiet = "quiet"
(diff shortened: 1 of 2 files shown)


ErikSchierboom:
@ee7 Let me know when you want this to be reviewed.

ErikSchierboom (approved):
Very nice addition! Makes adding new options a lot easier. I do wish we had some unit tests to verify the behavior :)

PR #88, opened 2020-10-27
Refactor(cli): Be explicit in `initConf`

ee7:
This proc already didn't include `probSpecsDir`, and `offline`, so let's only set the value that's different from the default (which is `verbosity` - the default is `verQuiet` since we list it first).

### DIFF
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -101,6 +101,8 @@ proc initConf: Conf =
     exercise: none(string),
     mode: modeChoose,
     verbosity: verNormal,
+    probSpecsDir: none(string),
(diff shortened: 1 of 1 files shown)


ErikSchierboom (commented):
> This proc already didn't include probSpecsDir, and offline, so let's only set the value that's different from the default (which is verbosity - the default is verQuiet since we list it first).

I'd argue the other way around: adding the missing values :) I'd prefer the options to be explicit, as changing the order of the enum values would not result in different default being used. This seems to me not to be very intuitive.

ErikSchierboom (approved):

PR #89, opened 2020-10-27
Refactor: Change `Action` to bool `check`

ee7:
The option for using the check-only mode is `--check`, rather than `--action=check` so let's reflect that internally.

To-do:
- [x] Merge https://github.com/exercism/canonical-data-syncer/pull/88 then rebase this PR on `master`


ErikSchierboom (approved):

ErikSchierboom:
Thanks!

PR #90, opened 2020-10-30
Refactor: Convert some `Option[string]` to `string`

ee7:
There isn't much reason to use `Option[string]` for `exercise` and `probSpecsDir` - we can just test whether the string is empty.

This also allows us to consider setting the default value of `probSpecsDir` to `.problem-specifications` in the future, which better models the current behaviour.  

Note that (both before and after this commit) if we run any of:
```
canonical_data_syncer -p=
canonical_data_syncer -p
canonical_data_syncer -e=
canonical_data_syncer -e
```

Then we see an error message like:
```
Error: '-p' was given without a value
```

### DIFF
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -1,4 +1,4 @@
-import std/[options, os, strformat, strutils, terminal]
+import std/[os, strformat, strutils, terminal]
 import pkg/[cligen/parseopt3]
 
 type
@@ -13,11 +13,11 @@ type
     verDetailed = "detailed"
 
   Conf* = object
-    exercise*: Option[string]
+    exercise*: string
     check*: bool
     mode*: Mode
     verbosity*: Verbosity
-    probSpecsDir*: Option[string]
+    probSpecsDir*: string
     offline*: bool
 
   Opt = enum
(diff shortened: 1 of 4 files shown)


ErikSchierboom (approved):
I do feel that we're sacrifing readability a bit here. `conf.exercise.isNone` to me is more descriptive for what is happening than `conf.exercise.len == 0`. But I don't feel strongly about it.

ee7:
I actually prefer it this way.

I can agree that `.isNone` and `.isSome` are "more descriptive" than the corresponding `.len == 0` and `.len > 0`. But I personally find the latter more "readable" - it avoids the noise, and otherwise a reader might wonder whether there's some subtle behaviour that makes `Option` necessary. For example: they might ask: "is `some("")` possible here?"

It's also good enough for [Nim](https://github.com/nim-lang/Nim/blob/6fe2e8977d0227a454cfb606c821cc455a9d0e07/compiler/options.nim#L411-L465) and [Nimble](https://github.com/nim-lang/nimble/blob/b4f29540bc5ff8a64bc6e672bb5deb25b1959690/src/nimblepkg/options.nim#L22) (see in particular `nimbleDir`, which is a similar "option with a value").

There's also the argument of "if `probSpecsDir` is `Option[string]`, then `check` should be `Option[bool]`".

PR #92, opened 2020-10-30
Test: Add initial tests for the binary

ee7:
For now, this just tests some of the CLI parsing.

We should find a better way to structure the tests, because we want:
- a failing test of the binary should block a release (that doesn't happen with this PR)
- tests of the binary should run on `push` and `pull_request` (we can't just add it to `build.yml` without modification because that runs only on tagging)

https://github.com/exercism/canonical-data-syncer/blob/2ebc4f6d73adeeef4643a9e5b88a948962cf1e25/.github/workflows/build.yml#L3-L6

Maybe something like:
- move `test_binary.nim` out of `all_tests.nim`
- rename `all_tests.nim`
- make the release part of the `build.yml` workflow a separate job that depends on a testing job that runs on `push` and `pull_request`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,8 +36,19 @@ jobs:
         with:
           nim-version: "1.4.0"
 
+      - name: Install musl on Linux
+        if: matrix.target.os == 'linux'
+        run: ./.github/bin/linux-install-build-tools
+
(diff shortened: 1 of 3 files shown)


ErikSchierboom (approved):
Excellent! I'm fine with the way the tests are structured for now. We can always restructure or optimize later.

PR #94, opened 2020-11-27
Refactor(cli): Reduce duplication in option printing

ee7:
Just a small refactoring that I'd like to have in place - we'll define more error messages in the future, and it's otherwise easy to e.g. forget a quote.

There's probably a better name for this func, but I don't know what it is.

No need to create a release.

Note: the `runnableExamples` don't actually run if you run `nim doc cli.nim`, because the func isn't exported. But I've tested that they do run successfully if you add the `*` export marker to the func.

The alternatives that I'm aware of:
1. Remove the examples entirely, but then the documentation is less clear.
1. Turn the `runnableExamples` into doc comments, but then they're harder to test and thus keep in-sync with the code.
1. Move the examples into `when isMainModule` tests, and add `nim c -r cli.nim` to the CI workflow. This means that they're tested, but the documentation suffers because the examples aren't close to the func itself, and the test layout becomes more convoluted.
1. Move the examples into a new file in the `tests` directory. This requires using `include cli` (rather than `import`), or a short hack/macro to export non-exported procs for testing purposes.

### DIFF
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -110,11 +110,24 @@ proc showError*(s: string) =
   stdout.write("\n\n")
   showHelp(exitCode = 1)
 
-proc prefix(kind: CmdLineKind): string =
-  case kind
-  of cmdShortOption: "-"
-  of cmdLongOption: "--"
-  of cmdArgument, cmdEnd, cmdError: ""
+func formatOpt(kind: CmdLineKind, key: string, val = ""): string =
+  ## Returns a string that describes an option, given its `kind`, `key` and
+  ## optionally `val`. This is useful for displaying in error messages.
+  runnableExamples:
+    import pkg/[cligen/parseopt3]
+    assert formatOpt(cmdShortOption, "h") == "'-h'"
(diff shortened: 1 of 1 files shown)


src/cli.nim
```diff
@@ -110,11 +110,24 @@ proc showError*(s: string) =
...
 
-proc prefix(kind: CmdLineKind): string =
-  case kind
-  of cmdShortOption: "-"
-  of cmdLongOption: "--"
-  of cmdArgument, cmdEnd, cmdError: ""
+func formatOpt(kind: CmdLineKind, key: string, val = ""): string =
+  ## Returns a string that describes an option, given its `kind`, `key` and
+  ## optionally `val`. This is useful for displaying in error messages.
+  runnableExamples:
```
ErikSchierboom, line 116: I just checked the documentation and this is really nice! The docs state that these examples are compiled and tested as part of the documentation generator. Do we run the documentation generator in our CI pipeline? If not, it might be good to do so.

ErikSchierboom (approved):

PR #95, opened 2020-11-27
Refactor(cli): Generate help text

ee7:
This PR improves how we construct the `--help` string, but does not change the actual output of `--help`.

Previously, we had to be careful with the spacing so that the help message for each option is aligned correctly.

This PR also makes it clearer that we set the name of the application in the help message at runtime, rather than compile-time.


ErikSchierboom (approved):
I love this!

PR #96, opened 2020-11-27
Refactor: Rename to configlet_v3

ee7:
The two commits here are atomic, but we should probably squash them when merging. That's because the first commit has a strange state: a `configlet_v3` binary with the same CLI as the old `canonical_data_syncer` binary.

Decisions:
- Removed the old `fetch-canonical_data_syncer` scripts, keeping only the new `fetch-configlet_v3` scripts. I think we can update the already-existing scripts in track repos to download `v0.17.0`, but we don't need to keep them around in this repo?
- Make `verbosity` a global option, rather than one that's specific just to `sync`.
- Use a case object to implement the command-style CLI.
- Use `namedBin` rather than `bin` in the `.nimble` file, so that we don't have to rename the src file from `configlet_v3.nim` to `configlet.nim` later.
- Remove the name of the application from the `--version` output. This makes it slightly easier for a script to do something with the configlet version string, and again, helps with the rename later.

To-do:
- [X] Think a bit more about whether resetting to `0.1.0` rather than `0.18.0` is the best idea.
- [X] Make the new fetch scripts stay working later. The old approach would break when we rename the `configlet-v3` repo to `configlet`, because it hardcodes the archive filename to contain `configlet-v3`.


scripts/fetch-configlet_v3
```diff
@@ -1,8 +1,8 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
ee7, line 1: Here I've squeezed in an unrelated change. See https://github.com/exercism/configlet/commit/e13f7c74c1d6840ecf92e746aa32dec12cb5febb

ErikSchierboom (approved):
This is looking great! My only remark would be that we probably want to only show the `sync` commands options when the user has actually passed in `sync` as the command. If we get more commands, the output could get quite long.

README.md
```diff
@@ -1,31 +1,39 @@
-# Exercism Canonical Data Syncer
+# Configlet
```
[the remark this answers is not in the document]
ee7: Good idea. Will do.

configlet.nimble
```diff
@@ -0,0 +1,13 @@
+# Package
+version       = "0.1.0"
+author        = "ee7"
+description   = "A tool for managing Exercism language track repositories"
+license       = "AGPL-3.0-only"
+srcDir        = "src"
+
+namedBin["configlet"] = "configlet_v3"
```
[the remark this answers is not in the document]
ee7: Good idea - I'll add a comment to clarify `namedBin`, since it's probably unfamiliar to most readers.

But when we rename, we'll also need to update the below files/lines. The one in the `.nimble` file is one of the easier ones to remember, since that's the one that controls the binary name.

```
.github/bin/create-artifact
 6:  BINARY_NAME="configlet_v3"

.gitignore
 5:  configlet_v3

README.md
 1:  # Configlet
 3:  This is a development version of Configlet for use with v3 of Exercism.
13:    configlet_v3 [global-options] <command> [command-options]
35:  To use the application in your track, you can copy the [`scripts/fetch-configlet_v3`](./scripts/fetch-configlet_v3) and/or [`scripts/fetch-configlet_v3.ps1`](./scripts/fetch-configlet_v3.ps1) files to your track's repository. Running either of these scripts will download the latest version of the `configlet_v3` tool to your track's `bin` directory.
39:  Bug reports and pull requests are welcome on GitHub at https://github.com/exercism/configlet-v3.

configlet.nimble
 8:  namedBin["configlet"] = "configlet_v3"

scripts/fetch-configlet_v3
 5:  readonly LATEST='https://api.github.com/repos/exercism/configlet-v3/releases/latest'
35:  FILENAME="configlet_v3-${OS}-${ARCH}.${EXT}"
47:          curl --header "$HEADER" -s --location "$URL" -o bin/latest-configlet_v3.zip
48:          unzip bin/latest-configlet_v3.zip -d bin/
49:          rm bin/latest-configlet_v3.zip

scripts/fetch-configlet_v3.ps1
 2:      $latestUrl = "https://api.github.com/repos/exercism/configlet-v3/releases/latest"
17:  $fileName = "configlet_v3-windows-$arch.zip"

tests/test_binary.nim
 7:    binaryName = "configlet_v3" & binaryExt
```

ee7:
> My only remark would be that we probably want to only show the `sync` commands options when the user has actually passed in `sync` as the command. If we get more commands, the output could get quite long.

I want to consider this too. Are you happy to leave it for a later PR, after we add a couple more commands?

README.md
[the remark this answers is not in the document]
ee7: Actually, this change is potentially confusing, because the version string of "v2 configlet" starts with `v3` (the latest release is `v3.9.2` / `v3.9.10`).

Perhaps this?

```diff
- # Configlet v3
+ # Configlet for Exercism v3 
```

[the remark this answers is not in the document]
ErikSchierboom: Sure!

configlet.nimble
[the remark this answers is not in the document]
ErikSchierboom: Maybe add a todo to all those lines?

ErikSchierboom:
> I want to consider this too. Are you happy to leave it for a later PR, after we add a couple more commands?

 I'm happy to leave it as is for now.

README.md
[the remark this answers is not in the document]
ee7: Done.

scripts/fetch-configlet_v3
```diff
@@ -32,21 +32,20 @@ else
...
 fi
 
-FILENAME="canonical_data_syncer-${OS}-${ARCH}.${EXT}"
+SUFFIX="${OS}-${ARCH}.${EXT}"
 
 get_url () {
     curl --header "$HEADER" -s "$LATEST" |
-        awk -v filename=$FILENAME '$1 ~ /browser_download_url/ && $2 ~ filename { print $2 }' |
-        tr -d '"'
+        sed -n "s#\s*\"browser_download_url\": \"\(.*/download/.*/configlet.*${SUFFIX}\)\"\$#\1#p"
```
ee7, line 39:
```suggestion
        grep "^\s*\"browser_download_url\": \".*${SUFFIX}\"$" |
        cut -d'"' -f4
```
An alternative that uses `grep` and `cut` instead of `sed`. Note that the `sed` pattern doesn't have to include `configlet` either. But maybe it's a good idea to include it, in case we ever release more than one asset with the same `SUFFIX`.

Some goals here:
- Don't use `jq` - the user might not have it installed.
- Use only portable features of `grep`, `sed`, `awk` etc.

ErikSchierboom: > Here's an alternative that uses grep and cut instead of sed. Might be more readable to most people.

It definitely is better than the sed one. What was the issue with the `awk` variant? My sole goal with this bit of code is to make it as stable as possible. I don't think it is likely that github will change its API output, but it also doesn't hurt to be paranoid :) E.g., do we need the `^\s*` prefix for the browser download URL?

>  But maybe it's a good idea to include configlet, in case we ever release more than one asset with the same SUFFIX.

Agreed. This makes sense to me.

ee7: > What was the issue with the `awk` variant?

The key point is that the old fetch script hardcodes the filename of the asset. See:
https://github.com/exercism/canonical-data-syncer/blob/906ffde760b0c7dd1791ae2632fe39f497c75453/scripts/fetch-canonical_data_syncer#L35-L41

So we shouldn't do exact matching, which wouldn't work when we later rename the assets from e.g.
`configlet_v3-linux-64bit.tgz`
to
`configlet-linux-64bit.tgz`

Therefore this PR currently matches using `configlet.*${SUFFIX}` where SUFFIX is e.g. `linux-64bit.tgz`.

I went with `sed` because:
- It's less powerful than `awk`, and I generally favor the least powerful tool when two tools provide equal correctness. `grep` is even less powerful, but maybe it's more idiomatic to capture than to pipe to `cut`.
- Maybe the average reader/maintainer of this script is better at `sed` than `awk`.

ee7: Pinging @glennj, @IsaacG, @SaschaMann:

In this PR I'm changing the `fetch-canonical_data_syncer` script to work as the `fetch-configlet_v3` script. It will eventually become the new `fetch-configlet` script that lives in `exercism/configlet` and the track repos so that maintainers can download `configlet`.

I want to grab the value of the `browser_download_url` key from the GitHub API that corresponds to the user's platform. For the configlet repo, see the response here https://api.github.com/repos/exercism/configlet/releases/latest.

The previous version of the script hard-coded the filename, but we don't want to do that here; ideally, this script should stay working after this repo is renamed from `configlet-v3` to `configlet` (after being renamed from `canonical-data-syncer` to `configlet-v3`).

What's the most tasteful, portable command that doesn't require `jq` to be installed?

In a later PR I'll add the changes we recently made to the `fetch-configlet` script in `exercism/github-actions`.

SaschaMann: > What's the most tasteful, portable command that doesn't require jq to be installed?

If you don't like using grep/sed for it, perhaps include [JSON.sh](https://github.com/dominictarr/JSON.sh)? It's only 200ish lines

ee7: > If you don't like using grep/sed for it,

I do! I just want to check if there's a better way of doing portable pattern matching with substring extraction. Maybe it's better to do something that involves `case ... esac` or `=~`.

> perhaps include JSON.sh? It's only 200ish lines

Mmm. 200 lines might be short for a JSON parser, but it'd still make this script 4x longer - we want to distribute only one file. And it looks like we'd still need to pipe its output to `grep` etc anyway, at which point we're in a similar position?

```
$ curl -sSf https://api.github.com/repos/exercism/configlet/releases/latest | ./JSON.sh -b | grep browser
["assets",0,"browser_download_url"]	"https://github.com/exercism/configlet/releases/download/v3.9.2/configlet-linux-32bit.tgz"
["assets",1,"browser_download_url"]	"https://github.com/exercism/configlet/releases/download/v3.9.2/configlet-linux-64bit.tgz"
["assets",2,"browser_download_url"]	"https://github.com/exercism/configlet/releases/download/v3.9.2/configlet-linux-arm-v5.tgz"
["assets",3,"browser_download_url"]	"https://github.com/exercism/configlet/releases/download/v3.9.2/configlet-linux-arm-v6.tgz"
["assets",4,"browser_download_url"]	"https://github.com/exercism/configlet/releases/download/v3.9.2/configlet-mac-32bit.tgz"
["assets",5,"browser_download_url"]	"https://github.com/exercism/configlet/releases/download/v3.9.2/configlet-mac-64bit.tgz"
["assets",6,"browser_download_url"]	"https://github.com/exercism/configlet/releases/download/v3.9.2/configlet-windows-32bit.zip"
["assets",7,"browser_download_url"]	"https://github.com/exercism/configlet/releases/download/v3.9.2/configlet-windows-64bit.zip"
```

`shellcheck JSON.sh` also produces quite a few complaints.

IsaacG: If we're going to add dependencies, they should be managed as such, not copied in wholesale. It should be added in as a meta-repo.

If you're asking about grep vs sed vs awk, my personal preference would be something slightly more verbose and (for me) more readable: `| awk -F": " -v "suffix=$SUFFIX" '$1 ~ /browser_download_url/ && $2 ~ "releases/download/.*" suffix {gsub(/"/, "", $2); print $2}'`

If you want a pure-bash solution,

```
PATTERN="^ *\"browser_download_url\": \"(.*/releases/download/.*${SUFFIX})\""
curl ... | while read -r ; do [[ $REPLY =~ $PATTERN ]] && echo "${BASH_REMATCH[1]}"; done
```

JSON is becoming ubiquitous these days and `jq` is becoming (or already is) the standard tool for handling JSON on the CLI. How bad would it be to depend on having it?

ErikSchierboom: I don't want any dependencies :) 

> JSON is becoming ubiquitous these days and jq is becoming (or already is) the standard tool for handling JSON on the CLI. How bad would it be to depend on having it?

As far as I can tell, there is no guarantee that users will have `jq` installed. We want this script to work out of the box for virtually anyone.

SaschaMann: If we tell people, especially Windows or Mac users without package managers, to install jq, we might as well tell them to install configlet manually instead

configlet.nimble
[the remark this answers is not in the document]
ee7: I've added the comment about `namedBin`, but I lean against adding comments for the others.

That's because I don't want the list of todos to be the source of truth: when we do the rename later, we should grep for `configlet` rather than `todo` - it's easy to add more things that have to be changed at rename-time, and forget to add a todo.

scripts/fetch-configlet_v3
ee7: I don't want to add a `jq` dependency here. And in the next PR I'll add the headers that request a fixed version of the GitHub API, so we don't have to worry much about robustness.

Thanks for the `awk` and pure-bash solutions. But I'll just go with the `grep` version for now, hoping that it's the most readable for the average maintainer. We can bikeshed it more later if it's ever a problem. I preferred the `sed` command, but the `grep | cut` is fine (as long as we don't release an asset that contains `"` - and we won't).

I kept the `^` in the pattern because it communicates the intention a bit better. And the `[[:space:]]` is marginally more portable than the `\s`.

The main thing is that we avoid features that aren't available on all operating systems, like `grep -P` (see for example [OpenBSD grep](https://man.openbsd.org/grep)).

README.md
```diff
@@ -1,31 +1,39 @@
...
 ```
-Usage: canonical_data_syncer [options]
+Usage:
+  configlet_v3 [global-options] <command> [command-options]
+
+Commands:
+  sync
 
-Options:
+Options for sync:
```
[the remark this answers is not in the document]
ErikSchierboom: Or just:

```suggestion
sync options:
```

ErikSchierboom (approved):
Excellent! I'll let you do the honors of merging.

PR #106, opened 2020-12-16
Feature: Implement `uuid` command

ee7:
I included an `-n` feature, as I often wanted that.

### DIFF
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -1,4 +1,4 @@
-import std/[os, strformat, strutils, terminal]
+import std/[os, parseutils, strformat, strutils, terminal]
 import pkg/[cligen/parseopt3]
 
 type
@@ -15,6 +15,7 @@ type
   ActionKind* = enum
     actNil = "nil"
     actSync = "sync"
+    actUuid = "uuid"
 
   Action* = object
     case kind*: ActionKind
@@ -26,6 +27,8 @@ type
       mode*: Mode
       probSpecsDir*: string
       offline*: bool
+    of actUuid:
+      num*: int
 
   Conf* = object
     action*: Action
@@ -40,6 +43,7 @@ type
     optSyncMode = "mode"
     optSyncProbSpecsDir = "probSpecsDir"
     optSyncOffline = "offline"
+    optUuidNum = "num"
 
 func genShortKeys: array[Opt, char] =
   ## Returns a lookup that gives the valid short option key for an `Opt`.
@@ -107,6 +111,7 @@ func genHelpText: string =
         of optSyncExercise: "slug"
         of optSyncMode: "mode"
         of optSyncProbSpecsDir: "dir"
+        of optUuidNum: "int"
         else: ""
 
       let paramText = if paramName.len > 0: &" <{paramName}>" else: ""
@@ -128,6 +133,7 @@ func genHelpText: string =
                          "rather than cloning temporarily",
     optSyncOffline: "Do not check that the directory specified by " &
                     &"`{list(optSyncProbSpecsDir)}` is up-to-date",
+    optUuidNum: "Number of UUIDs to generate",
   ]
 
   result = "Commands:\n  "
@@ -199,6 +205,8 @@ func initAction*(actionKind: ActionKind, probSpecsDir = ""): Action =
     Action(kind: actionKind)
   of actSync:
     Action(kind: actionKind, probSpecsDir: probSpecsDir)
+  of actUuid:
+    Action(kind: actionKind, num: 1)
 
 func initConf*(action = initAction(actNil), verbosity = verNormal): Conf =
   result = Conf(
@@ -321,6 +329,17 @@ proc handleOption(conf: var Conf; kind: CmdLineKind; key, val: string) =
         setActionOpt(offline, true)
       else:
         discard
+    of actUuid:
+      case opt
+      of optUuidNum:
+        var num = -1
+        discard parseSaturatedNatural(val, num)
+        if num < 1:
+          showError(&"value for {formatOpt(kind, key)} is not a positive " &
+                    &"integer: {val}")
+        setActionOpt(num, num)
+      else:
+        discard
 
   if not isGlobalOpt and not isActionOpt:
     case conf.action.kind
@@ -350,3 +369,5 @@ proc processCmdLine*: Conf =
     if result.action.offline and result.action.probSpecsDir.len == 0:
       showError(&"'{list(optSyncOffline)}' was given without passing " &
                 &"'{list(optSyncProbSpecsDir)}'")
+  of actUuid:
+    discard
--- a/src/uuid/uuid.nim
+++ b/src/uuid/uuid.nim
@@ -9,8 +9,12 @@ proc outputUuids(n: Positive) =
   if n > maxNumUuids:
     logNormal &"The UUID output limit is {maxNumUuids}, but {n} UUIDs were requested."
     logNormal &"Outputting {maxNumUuids} UUIDs:"
-  for i in 1 .. min(n, maxNumUuids):
-    stdout.writeLine genUUID()
+  let numUuidsToGenerate = min(n, maxNumUuids)
+  var s = newStringOfCap(numUuidsToGenerate * 37)
+  for i in 1 .. numUuidsToGenerate:
+    s.add $genUUID()
+    s.add '\n'
+  stdout.write s
 
 proc uuid*(n: Positive) =
   outputUuids(n)
--- a/src/uuid/uuid.nim
+++ b/src/uuid/uuid.nim
@@ -7,7 +7,7 @@ proc outputUuids(n: Positive) =
   ## greater than 1000.
   const maxNumUuids = 1000
   if n > maxNumUuids:
-    logNormal &"The UUID output limit is 1000, but {n} UUIDs were requested."
+    logNormal &"The UUID output limit is {maxNumUuids}, but {n} UUIDs were requested."
     logNormal &"Outputting {maxNumUuids} UUIDs:"
   for i in 1 .. min(n, maxNumUuids):
     stdout.writeLine genUUID()
--- a/src/uuid/uuid.nim
+++ b/src/uuid/uuid.nim
@@ -1,0 +1,67 @@
+import std/[strformat]
+import pkg/[uuids]
+import ".."/[logger]
+
+proc uuid*(n: Positive) =
+  ## Writes `n` version 4 UUIDs to stdout. Outputs only 1000 UUIDs if `n` is
+  ## greater than 1000.
+  const maxNumUuids = 1000
+  if n > maxNumUuids:
+    logNormal &"The UUID output limit is 1000, but {n} UUIDs were requested."
+    logNormal &"Outputting {maxNumUuids} UUIDs:"
+  for i in 1 .. min(n, maxNumUuids):
+    stdout.writeLine genUUID()
+
+func isValidUuidV4*(s: string): bool =
+  ## Returns `true` if `s` is a valid version 4 UUID (compliant with RFC 4122)
+  ## in the canonical textual representation.
+  ##
+  ## This func is equivalent to using `re.match` with the below regex pattern:
+  ##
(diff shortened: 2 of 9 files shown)


ErikSchierboom (approved):
Looks good! One tiny nit

src/cli.nim
```diff
@@ -107,6 +111,7 @@ func genHelpText: string =
         of optSyncExercise: "slug"
         of optSyncMode: "mode"
         of optSyncProbSpecsDir: "dir"
+        of optUuidNum: "int"
```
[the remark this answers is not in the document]
ee7: That line sets the parameter name in the help message. My thinking is that the below is clearer with `<int>` than with `<num>`: 

```
Options for uuid:
  -n, --num <int>              Number of UUIDs to generate
```

because `<int>` better implies that a fraction of a UUID is meaningless. That is, 2.5, √2, π and 2 + 3i are all *numbers*, but we're trying to ask the user for a (positive) integer.

Do you think there's something better than `<int>`? And is there a better option name than `-n, --num`?

And I think it's okay to write `<int>`, even though we didn't write `<string>` in the options for sync. The values of `<slug>`, `<mode>` and `<dir>` can't be "any string" - they have to be a valid slug, mode or directory.
```
Options for sync:
  -c, --check                  Terminates with a non-zero exit code if one or more tests are missing. Doesn't update the tests
  -m, --mode <mode>            What to do with missing test cases. Allowed values: c[hoose], i[nclude], e[xclude]
  -p, --prob-specs-dir <dir>   Use this `problem-specifications` directory, rather than cloning temporarily
  -o, --offline                Do not check that the directory specified by `-p, --prob-specs-dir` is up-to-date
```

Trivia: in the current implementation, all of
```
configlet_v3 uuid -n 2.5
configlet_v3 uuid -n 2/7
configlet_v3 uuid -n 2.9999999999999999999999999999999999999999999999999999999999999999
```
just output 2 UUIDs without complaining.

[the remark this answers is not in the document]
ErikSchierboom: Aha, now I understand. It's fine

ee7:
Force-pushes:
1. Rebase on `master`
1. Add `isValidUuidV4` func ([ea90f8d..90b8cfe](https://github.com/exercism/canonical-data-syncer/compare/ea90f8d35da452b32d213cf8bba3210d14ef7bc9..90b8cfe54d15950a18e6a063b6503f7ca3ddd6a8)). We'll want to use this in the `configlet lint` implementation later, so we might as well use it in the test that this PR adds too.
1. Fix styleCheck problem in our code ([90b8cfe..5b697f7](https://github.com/exercism/canonical-data-syncer/compare/90b8cfe54d15950a18e6a063b6503f7ca3ddd6a8..5b697f7c93402fc30df055f394141c3c4687035b))
1. Workaround styleCheck problem in `uuid` package code ([5b697f7..b8b0715](https://github.com/exercism/canonical-data-syncer/compare/5b697f7c93402fc30df055f394141c3c4687035b..b8b0715a9010554a4192a7e10d6c3949e9dc13bf)). See failed CI run on Windows https://github.com/exercism/canonical-data-syncer/runs/1701155784

ErikSchierboom:
That's a weird-ass error for the WinBool value :D

src/uuid/uuid.nim
```diff
@@ -0,0 +1,67 @@
...
+func isValidUuidV4*(s: string): bool =
+  ## Returns `true` if `s` is a valid version 4 UUID (compliant with RFC 4122)
+  ## in the canonical textual representation.
+  ##
+  ## This func is equivalent to using `re.match` with the below regex pattern:
+  ##
+  ## `^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$`
+  ##
+  ## However, this func's implementation is faster and doesn't add a PCRE
+  ## dependency.
```
ErikSchierboom, line 24: The PCRE dependency is a good argument, the speed argument less so I feel as that won't be noticeable by the user most likely.

tests/test_binary.nim
```diff
@@ -104,5 +115,13 @@ proc main =
           outp.contains("'-o, --offline' was given without passing '-p, --prob-specs-dir'")
           exitCode == 1
 
+  suite "uuid":
```
ErikSchierboom, line 118: When does a test go in the `test_binary.nim` file and when in a separate file like `test_probspecs`? It feels like maybe you would want both here, with `test_uuid` containing uuid specific tests?

src/uuid/uuid.nim
```diff
@@ -0,0 +1,67 @@
...
+  result =
+    s.len == 36 and
+    s[0] in Hex and
+    s[1] in Hex and
+    s[2] in Hex and
+    s[3] in Hex and
+    s[4] in Hex and
+    s[5] in Hex and
+    s[6] in Hex and
+    s[7] in Hex and
```
ErikSchierboom, line 39: Did you not use slices here for performance? Lot of repetitive code. I don't necessarily mind that here though, just curious.

ErikSchierboom (approved):

ErikSchierboom (approved):

tests/test_uuid.nim
```diff
@@ -0,0 +1,140 @@
+import std/[unittest]
+import "."/[uuid/uuid]
+
+proc main =
```
ErikSchierboom, line 4: Maybe also a test here that the `uuid` function returns a valid `uuid`? I know it is duplicated in the `test_binary` file, but I feel like it should also be specified here. That other test is then more about that the right number of uuids is specified (with added validity checks).

ErikSchierboom (approved):

src/uuid/uuid.nim
ee7: > Did you not use slices here for performance?

indeed. This is faster than something involving `s[0..7]`, because slicing in Nim involves a copy. It unfortunately isn't an alias for `toOpenArray(s, 0, 7)`.

`toOpenArray` was introduced in Nim 0.19.0 (September 2018), but making no-copy more ergonomic is still being worked on.

I also prefer the PR implementation to something like
```Nim
const ValidUuidV4 = "01234567-9012-4567-9012-456789012345"
doAssert ValidUuidV4.len == 36

func uuidIndices: (seq[int], seq[int]) =
  for i, c in ValidUuidV4:
    if c == '-':
      result[0].add i
    else:
      result[1].add i

const (HyphenIndices, HexIndices) = uuidIndices()

func isValidUuidV4*(s: string): bool =
  const
    Hex = {'0'..'9', 'a'..'f'}
    Separator = '-'
    Version = '4'
    Variant = {'8', '9', 'a', 'b'}

  if s.len == 36 and s[14] == Version and s[19] in Variant:
    for i in HexIndices:
      if s[i] notin Hex:
        return false
    for i in HyphenIndices:
      if s[i] != Separator:
        return false
    result = true
  else:
    return false
```

It's easy to write a macro and reduce the duplication. But I'd just make it generate the `func` in this PR, and the code would be less obvious.

src/uuid/uuid.nim
ee7: Yes, the main argument is against PCRE and increasing the release binary size.

The speed difference is large, but far below what a user can perceive.

ErikSchierboom: Binary size is not a huge issue I think, but PCRE is essentially a dependency, right? As in: it has to be installed on the machine.

src/uuid/uuid.nim
ErikSchierboom: Yeah, let's just keep it this way

tests/test_uuid.nim
ee7: One problem is that I didn't write a UUID function that returns a string, but only one that writes to stdout (which is arguably a bad pattern).

I could add that proc and test it, but I don't know if that helps much - it doesn't test that `configlet uuid` actually uses that proc.

tests/test_binary.nim
ee7: Yes, that's a question we'll keep facing. It's partly a philosophical question of integration tests vs unit tests.

We can test our code with unit tests, but it doesn't guarantee that the binary actually uses the code that's being tested.

I've added a test of `genUUID` to `test_uuid.nim` though - any thoughts?

src/uuid/uuid.nim
```diff
@@ -2,8 +2,8 @@ import std/[strformat]
 import pkg/[uuids]
 import ".."/[logger]
 
-proc uuid*(n: Positive) =
-  ## Writes `n` version 4 UUIDs to stdout. Outputs only 1000 UUIDs if `n` is
+proc writeUuids(n: Positive) =
```
ErikSchierboom, line 5: Where are the uuids written to? Tiny suggestion:

```suggestion
proc outputUuids(n: Positive) =
```

tests/test_uuid.nim
ErikSchierboom: I think I would indeed separate the generating of the UUID and the writing of it

src/uuid/uuid.nim
ee7: Yes, but I meant that it's also possible to statically link PCRE. I haven't checked exactly how much that'd increase the binary size, but it'd be significant.

At the moment we use `--passL:static` on Linux and Windows:

https://github.com/exercism/canonical-data-syncer/blob/11ab015b4d6dcb692e353028b80882af60d91f3a/.github/bin/build#L6-L13

ErikSchierboom: Okay cool

tests/test_binary.nim
ErikSchierboom: I like it

src/uuid/uuid.nim
ee7: Done - see https://github.com/exercism/canonical-data-syncer/pull/106/commits/f78d103c60e9a0bb7316dfe6f20beab10f2a853a.

ee7:
For the remaining feedback, do you prefer:

1. The current state of the PR
1. To add  [this extra commit](https://github.com/ee7/exercism-canonical-data-syncer/compare/feat-configlet-add-uuid-command...uuid-extra-proc)? However, I don't like to export procs just for testing purposes - I think it's generally better to test the (minimal) public interface. Will we ever need to generate a UUID for something other than `configlet uuid`? 
1. Write `include uuid/uuid` in the test file, instead of `import`
1. Add `when isMainModule` tests at the bottom of the `uuid.nim` file (messy: these tests are awkward to add to CI, and our tests are no longer just in the `tests` directory)
1. Add a short macro that annotates procs that should be exported for testing only. See https://forum.nim-lang.org/t/3269

In `my_macro.nim`:
```Nim
import macros

macro tpub*(x: untyped): untyped =
  ## Adds the export marker to a proc when `-d:test` is defined.
  expectKind(x, RoutineNodes)
  when defined(test):
    let n = name(x)
    x.name = newTree(nnkPostfix, ident"*", n)
  result = x
```

In `foo.nim`:
```Nim
import my_macro

proc sayHi: string {.tpub.} =
  result = "hi"
```

In `tfoo.nim`:
```Nim
import foo

doAssert sayHi() == "hi"
```

Then run
```shell
nim c -d:test -r tfoo
```

But one problem here is that IDEs will typically complain when viewing `tfoo.nim`, saying that `sayHi` is undeclared.

ErikSchierboom:
> However, I don't like to export procs just for testing purposes - I think it's generally better to test the (minimal) public interface. Will we ever need to generate a UUID for something other than configlet uuid?

We originally considered automatically adding the uuid to config.json elements that required them, but were missing them. That said, I'm fine with postponing that to see if we actually want that (I'm not sure).

Let's get the current version merged. I don't like option 3/4/5 and option 2 is something we could do later if so desired.

ee7:
> > However, I don't like to export procs just for testing purposes - I think it's generally better to test the (minimal) public interface. Will we ever need to generate a UUID for something other than configlet uuid?
> 
> We originally considered automatically adding the uuid to config.json elements that required them, but were missing them. That said, I'm fine with postponing that to see if we actually want that (I'm not sure).
> 
> Let's get the current version merged. I don't like option 3/4/5 and option 2 is something we could do later if so desired.

OK - I'll merge.

ErikSchierboom:
Great! We now have two commands 🎉

PR #112, opened 2021-01-12
CI: Add shellcheck workflow

ee7:
This causes CI to indicate failure if `shellcheck` has any complaint
about any of our scripts.

Note that we can't write

```yml
      - name: Run shellcheck
        uses: ludeeus/action-shellcheck@v1
```

because the `v1` tag doesn't exist, even though the latest version is `1.0.0`.

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,13 @@
+name: shellcheck
+on: [push, pull_request]
+
+jobs:
+  shellcheck:
+    name: Run shellcheck on scripts
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f # 2.3.4
+
+      - name: Run shellcheck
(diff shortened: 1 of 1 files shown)


.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,13 @@
+name: shellcheck
+on: [push, pull_request]
+
+jobs:
+  shellcheck:
+    name: Run shellcheck on scripts
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
ErikSchierboom, line 10: Could you pin this to a specific SHA?

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,13 @@
...
+
+      - name: Run shellcheck
+        uses: ludeeus/action-shellcheck@master
```
ErikSchierboom, line 13: Could you pin this to a specific SHA?

.github/workflows/shellcheck.yml
ee7: Sure. Done.

.github/workflows/shellcheck.yml
ee7: Sure. Done.

ErikSchierboom (approved):

PR #117, opened 2021-01-13
Chore: Add dependabot

ee7:
Do we want this?

---
Dependabot will now create a PR on the first day of every month for any
outdated GitHub Action that we use in a workflow.

The current list of GitHub Actions that we use is:
- actions/checkout
- jiro4989/setup-nim-action
- ludeeus/action-shellcheck

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+
+updates:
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
I like this!

PR #120, opened 2021-01-15
fetch-configlet: update PowerShell version to match shell version

ErikSchierboom:
There were quite some differences between the shell version of fetch_configlet and the powershell version. This fixes those.

- fetch-configlet: add retry count
- fetch-configlet: preserve authorization header on redirects
- fetch-configlet: stop when any error occurs
- fetch-configlet: fix authorization header not passed
- fetch-configlet: pass parameters as splat
- fetch-configlet: Use shared request options
- fetch-configlet: use pipe to simplify JSON processing
- fetch-configlet: inline if-statements

### DIFF
--- a/scripts/fetch-configlet.ps1
+++ b/scripts/fetch-configlet.ps1
@@ -1,3 +1,5 @@
+$ErrorActionPreference = "Stop"
+
 Function DownloadUrl ([string] $FileName, $Headers) {
     $latestUrl = "https://api.github.com/repos/exercism/configlet-v3/releases/latest"
     $json = Invoke-RestMethod -Headers $Headers -Uri $latestUrl -MaximumRetryCount 3 -RetryIntervalSec 1 -PreserveAuthorizationOnRedirect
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
Please don't squash

ee7 (approved):
I don't have the PowerShell expertise to review this properly, but it seems reasonable to me.

The main difference that I noticed (which isn't really worth noting) is that the `curl` commands in the bash script have an exponential backoff in the retry, so that the 3 retries occur after 1, 2 and 4 second intervals (total of 7 seconds), whereas this version retries 3 times in 1 second intervals.

Can you confirm that it works if you change the URL from `exercism/configlet-v3` to `exercism/configlet` to download the current `configlet` release for Exercism v2?

Approved if so.

ErikSchierboom:
> The main difference that I noticed (which isn't really worth noting) is that the curl commands in the bash script have an exponential backoff in the retry, so that the 3 retries occur after 1, 2 and 4 second intervals (total of 7 seconds), whereas this version retries 3 times in 1 second intervals.

That's correct. There is no such functionality built-in. This is my best approximation.

> Can you confirm that it works if you change the URL from exercism/configlet-v3 to exercism/configlet to download the current configlet release for Exercism v2?

I've tested this with the canonical-data-syncer, and that worked. I should have used configlet instead 🤦

PR #125, opened 2021-01-19
[CI] Optimize workflows

ErikSchierboom:
This PR ensures that the shellcheck and tests worflows only run on pushed to the main or master branch. This prevents duplicate execution of the workflows when a push is made to pull request. The PR also introduces support for manually triggering the workflows.

- CI: Only run shellcheck and tests on pushes to main or master branch
- CI: Allow manually triggering workflow

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - "*"
+  workflow_dispatch:
 
 jobs:
   build:
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,5 +1,6 @@
 name: shellcheck
-on: [push, pull_request]
+
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
   shellcheck:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,6 +1,6 @@
 name: tests
 
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
   all_tests:


ee7:
I approve the commit that allows triggering the workflow manually (https://github.com/exercism/canonical-data-syncer/pull/125/commits/89c31b262d05b0fbf055b9c374f260f2978d440a).

But the other commit (https://github.com/exercism/canonical-data-syncer/pull/125/commits/68ed8edf95705aa94d35322ff3d197a4104067f9) is not a strict "optimization", as the builds aren't strictly "duplicate". See e.g.:
https://github.community/t/duplicate-checks-on-push-and-pull-request-simultaneous-event/18012/4

The `push` build runs on the most recent branch commit, but the `pull_request` build runs on the result of the merge of the PR branch and the main branch. Therefore it's possible for one build to fail, and the other build to pass.

To my understanding, these are the pros and cons of https://github.com/exercism/canonical-data-syncer/pull/125/commits/68ed8edf95705aa94d35322ff3d197a4104067f9:
1. Pro: when pushing to a branch that has a PR and is strictly ahead of the main branch, we prevent an extra `push` build. But if that build is on a user's fork, this doesn't consume Exercism CI minutes.
1. Con: a build won't occur on a branch (on a contributor's fork or on the main repo) before a PR is created for it. 
1. Con: when pushing to a branch that has merge conflicts, we won't get any build at all (because this PR suppresses the `push`, and the `pull_request` build is not triggered because there are merge conflicts). However, there are good reasons to modify a branch before rebasing it on the main branch. For example, to apply a previously suggested change that itself must be modified when rebasing on the main branch. I also might want to see a CI run for new changes on a branch before resolving merge conflicts.

Running CI without/before creating a PR is more important to me than suppressing the sometimes-superfluous build. I always run CI on a branch in my fork before making a PR, or often when testing to see if a PR is worthwhile. This commit means that I have to trigger the build manually, or temporarily push to the main branch on my fork. A contributor who doesn't know to do that might actually end up using more CI minutes, because they first have to fix the failing build when they open the PR. 

I think the condition I want for builds is something like: 
"run like `on: [push, pull_request]`, but don't run on `push` if the branch is strictly ahead of the main branch, and has an open PR."

But that's more tricky to implement.

Is Exercism running short of CI minutes? If so, maybe it's better to first try:
- Use `paths` or `paths-ignore` so that a workflows run only when certain files are changed. The shellcheck workflow only consumes about 7 billable seconds though.
- Encourage users to make PRs from their fork (unless they want to collaborate with another user who has write access to the upstream repo).
- Implement something closer to the above condition. To suppress `push` builds on the upstream repo (untested):
```
if: github.event_name == 'push' && (github.repository != 'exercism/configlet-v3')
```

ErikSchierboom:
> Pro: when pushing to a branch that has a PR and is strictly ahead of the main branch, we prevent an extra push build. But if that build is on a user's fork, this doesn't consume Exercism CI minutes.
>
> Con: a build won't occur on a branch (on a contributor's fork or on the main repo) before a PR is created for it.
>
> Con: when pushing to a branch that has merge conflicts, we won't get any build at all (because this PR suppresses the push, and the pull_request build is not triggered because there are merge conflicts). However, there are good reasons to modify a branch before rebasing it on the main branch. For example, to apply a previously suggested change that itself must be modified when rebasing on the main branch. I also might want to see a CI run for new changes on a branch before resolving merge conflicts.

> Running CI without/before creating a PR is more important to me than suppressing the sometimes-superfluous build.

So this all depends on who the person is that is contributing. In my experience, most people start with a PR and only worry about CI then. Your workflow is different, with testing on branches first. But I see the arguments. Let's have more people weigh in.

> Is Exercism running short of CI minutes?

There is not a fixed amount of minutes, but we do try to keep the running time at a minimum. I've seen builds become quite slow due to us overloading the system, so I assume that there is some mechanism that prevents repos/orgs from completely spamming the CI system. This repo is also one of the slower CI builds, as the macOS build takes quite a lot of time.

SleeplessByte:
Absolutely agreed with @ee7 here. I've commented a few times on other repos how these events work. 

That said:

I prefer more (unnecessary builds) including ones that are prone to catch issues than convoluted setups that remove all the unnecessary builds.

ErikSchierboom:
Okay, clear. I've updated.

ee7 (approved):

SaschaMann:
First to clear it up: CI minutes don't matter for public repos, they're not capped (at least not visibly in any way).

---

> Therefore it's possible for one build to fail, and the other build to pass.

Theoretically it's possible but I have yet to actually run into a situation where this actually happens _and_ where the push build is relevant on its own. All the arguments in favour of running both assume that this is a common scenario. It might be, I've just never experienced it myself. In the end the only thing that matters is that the merged build (i.e. the `pull_request` build) passes unless it's some kind of release or git-flow-style dev branch (which would require different setups anyway). Running hundreds of extra builds only for such an eventuality is just wasteful even if we don't have to pay for it directly.

If there's a situation where the PR build fails, but tests pass locally _and_ the branch is up-to-date, you can still trigger a branch build manually. Those situations are rare enough that the extra effort is negligible.

> Running CI without/before creating a PR is more important to me than suppressing the sometimes-superfluous build. I always run CI on a branch in my fork before making a PR, or often when testing to see if a PR is worthwhile. 

The build is in my experience not sometimes-superfluous, but pretty-much-always-superfluous. That aside, you can open a draft PR and it will run the build regardless, so the same workflow still works, except the CI results will actually be more meaningful as they don't ignore the state of the main branch.

> I prefer more (unnecessary builds) including ones that are prone to catch issues than convoluted setups that remove all the unnecessary builds.

I agree. Conditions that go beyond modifying the `on:` bit of a workflow can be hard to test and are somewhat fragile.

> Use paths or paths-ignore so that a workflows run only when certain files are changed. The shellcheck workflow only consumes about 7 billable seconds though.

This is a good idea in general. Changes to the README and such rarely require a full CI run.

PR #126, opened 2021-01-19
git: ignore tests binary

ErikSchierboom:
Currently, running the tests using `nim c-r ./tests/all_tests.nim` results in a binary file being produced. This PR updates the .gitignore file to ignore said tests binary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,8 @@
 # Application binary
 configlet
 
+# Tests binary
+all_tests
+
 # Temporary prob-specs repo
 .problem-specifications/


.gitignore
```diff
@@ -4,5 +4,8 @@
 # Application binary
 configlet
 
+# Tests binary
+all_tests
```
ee7, line 8:
```suggestion
# Ignore all files that lack an extension
*
!/**/
!*.*

# Except those in these directories
!.github/**
!bin/**
!scripts/**
```

ee7 (commented):
The current change also doesn't ignore `tests/all_tests`, which will be created if we run `nim c -r all_tests.nim` using `tests` as a working directory.

Furthermore, we may often run nim files in `tests` and `src` individually during development. So would it be better to just ignore all files without an extension that aren't in the `.github`, `bin` and `scripts` directories?

We could also change commands so that they run `nim r` rather than `nim c -r`, meaning that they don't produce a binary in the working directory. From `nim --help`:
```
    nim command [options] [projectfile] [arguments]

Command:
    compile, c              compile project with default code generator (C)
    r                       compile to $nimcache/projname, run with [arguments]
                            using backend specified by `--backend` (default: c)
    doc                     generate the documentation for inputfile for
                            backend specified by `--backend` (default: c)
```

.gitignore
ErikSchierboom: > The current change also doesn't ignore tests/all_tests, which will be created if we run nim c -r all_tests.nim using tests as a working directory.

This is not actually true, as there is no path specified anywhere so it'll happily ignore all files named `all_tests`.

> Furthermore, we may often run nim files in tests and src individually during development. So would it be better to just ignore all files without an extension that aren't in the .github, bin and scripts directories?

I feel like the solution listed above is very hard to read. The individual test files are all prefixed with `test_`, so what if we do:

```suggestion
# Application binary
configlet

# Tests binary
all_tests
test_*
```

I feel like running individual files from the `src` directory will be far less common.

> We could also change commands so that they run nim r rather than nim c -r, meaning that they don't produce a binary in the working directory. 

I like this a lot. We should definitely document this somewhere.

ee7: > This is not actually true, as there is no path specified anywhere so it'll happily ignore all files named `all_tests`.

Oops.

```
# Tests binary
all_tests
test_*
```

I think we don't want `test_*`, becaues this would ignore Nim files in the `tests` directory, e.g. `tests/test_lint.nim`.

> I like this a lot. We should definitely document this somewhere.

It was added in Nim 1.4.0, by the way. You can also pipe to it, like:

```shell
 printf 'import algorithm; echo [5, 2, 1, 3, 4].sorted' | nim --hints:off r - 
```

Although in this case it's faster if you skip the compilation step via `nim e`
```shell
 printf 'import algorithm; echo [5, 2, 1, 3, 4].sorted' | nim --hints:off e - 
```
Both output `@[1, 2, 3, 4, 5]`.

ErikSchierboom:
Closed in favor of https://github.com/exercism/canonical-data-syncer/issues/128

PR #127, opened 2021-01-19
lint: Add stub implementation

ErikSchierboom:
This PR is based off https://github.com/exercism/canonical-data-syncer/pull/106, so it should only be merged after that PR is merged.

In this PR, a minimal stub implementation of the `lint` command is added. My goals with this PR are as follows:

- Output a friendly error to tracks when they try to run `configlet lint` (which will be sooner rather than later)
- Let tracks know that we're busy implementing the linting rules and that they thus need to re-run the check regularly
- Provide a base for others to also start working on the linter. Admittedly, this doesn't help much at the moment, but it is _a_ start.

I've deliberately not added any actual linting rule implementations or tests.

### DIFF
--- a/src/lint/lint.nim
+++ b/src/lint/lint.nim
@@ -1,0 +1,5 @@
+proc lint*() =
+  stdout.writeLine "No linting rules have yet been implemented.\n" &
+                   "As we're in the process of implementing the linting rules, " &
+                   "please re-run this command regularly to see if your track " &
+                   "passes the latest linting rules."
--- a/src/lint/lint.nim
+++ b/src/lint/lint.nim
@@ -1,5 +1,5 @@
-proc lint*() =
-  stdout.writeLine "No linting rules have yet been implemented.\n" &
-                   "As we're in the process of implementing the linting rules, " &
-                   "please re-run this command regularly to see if your track " &
-                   "passes the latest linting rules."
+proc lint* =
+  echo "No linting rules have yet been implemented.\n" &
+       "As we're in the process of implementing the linting rules, please " &
+       "re-run this command regularly to see if your track passes the " &
+       "latest linting rules."
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -14,12 +14,15 @@ type
 
   ActionKind* = enum
     actNil = "nil"
+    actLint = "lint"
     actSync = "sync"
     actUuid = "uuid"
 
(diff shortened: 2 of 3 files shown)


src/lint/lint.nim
```diff
@@ -0,0 +1,5 @@
+proc lint*() =
+  stdout.writeLine "No linting rules have yet been implemented.\n" &
+                   "As we're in the process of implementing the linting rules, " &
+                   "please re-run this command regularly to see if your track " &
+                   "passes the latest linting rules."
```
ee7, line 5:
```suggestion
proc lint* =
  echo "No linting rules have yet been implemented.\n" &
       "As we're in the process of implementing the linting rules, please " &
       "re-run this command regularly to see if your track passes the " &
       "latest linting rules."
```

Nit: let's prefer `echo` when we want to write to stdout with a newline, and flush.

I use `stdout.write` when I want to write to stdout without a newline (useful when you want to print something to the same line later, possibly based on some condition). And `stdout.writeLine` when writing a lot of lines to stdout, avoiding the overhead of continual, needless flushing (but it can be better to instead fill a string and write that once).

ee7:
Approved, but we could add something about the WIP-status of `lint` to the README in the next PR.

The output of `./configlet --help` regarding the `lint` options is a bit weird, but fine for now:
```
Usage:
  configlet [global-options] <command> [command-options]

Commands:
  lint, sync, uuid

Options for lint:

Options for sync:
  -c, --check                  Terminates with a non-zero exit code if one or more tests are missing. Doesn't update the tests
  -m, --mode <mode>            What to do with missing test cases. Allowed values: c[hoose], i[nclude], e[xclude]
  -p, --prob-specs-dir <dir>   Use this `problem-specifications` directory, rather than cloning temporarily
  -o, --offline                Do not check that the directory specified by `-p, --prob-specs-dir` is up-to-date

Options for uuid:
  -n, --num <int>              Number of UUIDs to generate

Global options:
  -h, --help                   Show this help message and exit
      --version                Show this tool's version information and exit
  -v, --verbosity <verbosity>  The verbosity of output. Allowed values: q[uiet], n[ormal], d[etailed]
```

ee7 (approved):

ErikSchierboom:
Agreed. Will be good to add something to the README.

PR #129, opened 2021-01-21
tests: always build fresh binary for binary tests

### LINKED ISSUE #128: Allow running tests with `nim r`
It would nice to be able to run the test suite without a binary being compiled. This prevents the issues that we had in https://github.com/exercism/canonical-data-syncer/pull/126, where a binary file is produced when running the tests.

Currently, running `nim r --styleCheck:error ./tests/all_tests.nim` will fail as it depends on the `configlet` binary being produced in the `src` dir.

ErikSchierboom:
Closes #128

### DIFF
--- a/tests/test_binary.nim
+++ b/tests/test_binary.nim
@@ -11,7 +11,10 @@ proc main =
   let binaryPath = repoRootDir / binaryName
   const helpStart = &"Usage:\n  {binaryName} [global-options] <command> [command-options]"
 
-  discard execCmdEx("nimble build -d:release", workingDir = repoRootDir)
+  const cmd = "nimble build -d:release"
+  stdout.write(&"Running `{cmd}`... ")
+  stdout.flushFile()
+  discard execCmdEx(cmd, workingDir = repoRootDir)
 
   suite "help as an argument":
     test "help":
--- a/tests/test_binary.nim
+++ b/tests/test_binary.nim
@@ -7,9 +7,11 @@ const
   binaryName = "configlet" & binaryExt
 
 proc main =
-  let repoRootDir = getAppDir().parentDir()
+  const repoRootDir = currentSourcePath.parentDir().parentDir()
   let binaryPath = repoRootDir / binaryName
   const helpStart = &"Usage:\n  {binaryName} [global-options] <command> [command-options]"
+
+  discard execCmdEx("nimble build", workingDir = repoRootDir)
 
   suite "help as an argument":
     test "help":
--- a/tests/test_binary.nim
+++ b/tests/test_binary.nim
@@ -11,7 +11,7 @@ proc main =
   let binaryPath = repoRootDir / binaryName
   const helpStart = &"Usage:\n  {binaryName} [global-options] <command> [command-options]"
 
-  discard execCmdEx("nimble build", workingDir = repoRootDir)
+  discard execCmdEx("nimble build -d:release", workingDir = repoRootDir)
 
   suite "help as an argument":
     test "help":
--- a/tests/test_binary.nim
+++ b/tests/test_binary.nim
@@ -14,7 +14,12 @@ proc main =
   const cmd = "nimble build -d:release"
   stdout.write(&"Running `{cmd}`... ")
   stdout.flushFile()
-  discard execCmdEx(cmd, workingDir = repoRootDir)
+  let (buildOutput, buildExitCode) = execCmdEx(cmd, workingDir = repoRootDir)
+  if buildExitCode == 0:
+    echo "success"
+  else:
+    echo "failure"
+    raise newException(OSError, buildOutput)
 
   suite "help as an argument":
     test "help":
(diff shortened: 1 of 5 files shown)


tests/test_binary.nim
```diff
@@ -7,10 +7,12 @@ const
   binaryName = "configlet" & binaryExt
 
 proc main =
-  let repoRootDir = getAppDir().parentDir()
+  let repoRootDir = getCurrentDir()
```
ErikSchierboom, line 10: If I ran `nim r --styleCheck:error ./tests/all_tests.nim`, this would equal: `/home/erik/.cache/nim`

With this change, it resolved to the directory in which the `nim` command is ran, and running it like above correctly builds the binary.

I'm running Nim 1.4.2, not sure if that matters.

ee7:
Sorry, when I said
> Make `tests/test_binary.nim` always run `nimble build` first.`

I meant "run `nimble build` with the options we use to build a release". 

The end-to-end test should test the actual binary that we release,which is what happens with the current `tests.yml`:
https://github.com/exercism/canonical-data-syncer/blob/f29d9be74797e6141bc9609cb7434a20d474ec55/.github/workflows/tests.yml#L46-L54

So if we make `test_binary.nim` build the `configlet` binary, it should do it with the `.github/bin/build` options:

https://github.com/exercism/canonical-data-syncer/blob/f29d9be74797e6141bc9609cb7434a20d474ec55/.github/bin/build#L5-L15

not a plain `nimble build`.

ErikSchierboom:
Good catch. I'll update

tests/test_binary.nim
ee7, line 10:
```suggestion
  const repoRootDir = currentSourcePath.parentDir().parentDir()
```

See https://nim-lang.github.io/Nim/system.html#currentSourcePath.t

Does this work for you? During development, I like to be able to run the tests from any directory.

ErikSchierboom: Yes, that works nicely. It fixes the path issue

ErikSchierboom:
@ee7 Regarding the build arguments, is it possible to have the build arguments in a separate configuration file? That way, we would have only one place to define the options.

ee7:
> @ee7 Regarding the build arguments, is it possible to have the build arguments in a separate configuration file? That way, we would have only one place to define the options.

Yes. I'll try to suggest changes that let us remove `.github/bin/build` completely.

ErikSchierboom:
@ee7 Check out https://github.com/exercism/canonical-data-syncer/pull/129/commits/47545672d8684778d441da665ddbff1a18eaf8af. Is that the right approach? It works on my machine

tests/test_binary.nim
ee7: > If I ran `nim r --styleCheck:error ./tests/all_tests.nim`, this would equal: `/home/erik/.cache/nim`

Great catch. We should indeed not determine `repoRootDir` from the path of the output binary. The `getAppDir` approach also breaks if somebody writes something like `nim c --outDir:/tmp -r tests/all_tests.nim`. 

> Is it common to run tests from the root directory, or more common to do that from the tests directory?

I don't know, but I think it's best if they can be run from anywhere. So I think `currentSourcePath` is better than `getCurrentDir`.

> I'm running Nim 1.4.2, not sure if that matters.

It doesn't matter, except insofar that `nim r` was only introduced in 1.4.0. But Nim 1.4.0 was released 2 weeks before this code was committed.

ErikSchierboom: I also think `currentSourcePath` is indeed the best approach.

ee7:
> @ee7 Check out 4754567. Is that the right approach? It works on my machine

Looks good - that's roughly what I'd do.

One remaining thing is that `tests.yml` now builds the binary twice:

https://github.com/exercism/canonical-data-syncer/blob/f29d9be74797e6141bc9609cb7434a20d474ec55/.github/workflows/tests.yml#L43-L54

Would you approve me pushing [this commit](https://github.com/ee7/exercism-canonical-data-syncer/commit/4efdbec150967fd151793043004de0fea656a8d9) to this PR's branch?

It's still not ideal, because the `tests.yml` also duplicates the `nimble -y install --depsOnly` part of the `build` file.

In the long-term I think we should make some deeper change. For example, make the current job in `build.yml` contain only the "Create artifact" and "Publish release" steps, but depend on an earlier job that runs the tests. Currently, we assume that the `tests.yml` build process is identical to that in `build.yml` - the job in `tests.yml` does not test the exact binary that is released (even if it could be bit-identical).

But I think we should handle that in a later PR.

I'd also like to push the other commits [here](https://github.com/ee7/exercism-canonical-data-syncer/commits/build-binary-in-test) to this PR's branch - any thoughts?

ErikSchierboom:
Yes, those all look like good additions. Push them to this PR.

ee7 (approved):

PR #131, opened 2021-01-22
Docs(README): Improve

ee7:
There were a few similar sentences that I probably didn't combine particularly well.

It will be easier to review the commits individually (we should squash this PR, though).


ErikSchierboom (approved):
Excellent

PR #134, opened 2021-01-26
sync: refactor writing of `tests.toml`

ee7:
If this is approved, I'll squash this.

While working on #72 I didn't want to add a second parameter to a proc named `writeFile`, so I'd prefer to merge something like this PR first.

Feel free to bikeshed.

### DIFF
--- a/src/sync/exercises.nim
+++ b/src/sync/exercises.nim
@@ -89,7 +89,7 @@ proc toToml(exercise: Exercise): string =
     result.add(&"\n# {testCase.description}")
     result.add(&"\n\"{testCase.uuid}\" = {isIncluded}\n")
 
-proc writeFile*(exercise: Exercise) =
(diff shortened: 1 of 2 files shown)


src/sync/exercises.nim
```diff
@@ -89,8 +89,9 @@ proc toToml(exercise: Exercise): string =
...
-proc writeFile*(exercise: Exercise) =
-  createDir(parentDir(exercise.testsFile))
+proc writeTestsToml*(exercise: Exercise) =
+  let testsPath = testsFile(exercise)
+  createDir(testsPath.parentDir())
 
-  let file = open(exercise.testsFile, fmWrite)
-  write(file, exercise.toToml())
+  let contents = toToml(exercise)
+  writeFile(testsPath, contents)
```
ErikSchierboom, line 97: I like this much better

ErikSchierboom (approved):

PR #137, opened 2021-01-27
Fix fetch configlet script

ErikSchierboom:
While testing the `fetch-configlet` script, I found that it didn't work:

```
tar (child): -C: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now
```

This turned out to be due to the arguments being in the wrong order.

I've also added a second commit to introduce a variable for the output directory, which makes it more concistent with the PowerShell script and reduce duplication.


ee7 (approved):
Great catch, thanks.

PR #139, opened 2021-01-27
Fetch script now downloads script to execute

ErikSchierboom:
For the background/motivation, see: https://github.com/exercism/configlet/issues/181


scripts/fetch-configlet
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+script_url="https://raw.githubusercontent.com/exercism/configlet/master/scripts/fetch-configlet-script"
```
ee7, line 5: Should we rename the main branch from `master` to `main` in this repo and in `exercism/github-actions` in advance of doing it in all the other repos?

ErikSchierboom: We should definitely do it here. I'll update the PR. I don't know what happens if we rename the master branch on github-actions though. Does it still work when people use it with `@master`?

ee7: > Does it still work when people use it with `@master`?

I think so. According to [this GitHub blog post](https://github.blog/changelog/2020-07-22-github-actions-better-support-for-alternative-default-branch-names/):

> In addition, we have renamed the default branch for most of the GitHub-authored actions from `master` to `main`. All of your workflows that reference the old branch name will still work. However, you’ll see a prompt to change to the new name when you edit the workflow on the web.

But I think we can always try renaming from `master` to `main`, and if it stops working, temporarily create a `master` branch in `exercism/github-actions` that's even its `main`.

I notice that https://github.com/actions/checkout has both a `main` and `master` branch. But I don't know if this is why (it's also strange that `master` is 10 commits behind `main`...)

ErikSchierboom: See https://github.com/exercism/canonical-data-syncer/pull/140

ErikSchierboom: The rename of the github-actions is on our todo list BTW: https://github.com/exercism/v3/issues/3079#issue-792841783

ErikSchierboom:
At some point, it would be good to add a test too to verify that the `fetch-configlet` script actually works. But I think we can postpone that.

ee7:
> At some point, it would be good to add a test too to verify that the `fetch-configlet` script actually works. But I think we can postpone that.

Indeed. Let's keep track of that in https://github.com/exercism/canonical-data-syncer/issues/121

scripts/fetch-configlet
```diff
@@ -1,58 +1,36 @@
...
+# Download the script that will fetch the configlet binary
+script_url="https://raw.githubusercontent.com/exercism/configlet/main/scripts/fetch-configlet-script"
+script=$(curl --silent --show-error --fail --location --retry 3 "${script_url}")
+
+# Require confirmation from the user before executing the fetch script
+# if the --confirm or -c command line argument was passed
+if [[ "$*" == "--confirm" ]] || [[ "$*" == "-c" ]]; then
+    message=$(printf "\n\n%s\n\nDo you want the execute the above script (y/n)? " "${script}")
+    read -r -p "${message}" choice
+    if [[ "${choice}" != "y" ]] && [[ "${choice}" != "Y" ]]; then
```
ee7, line 30:
```suggestion
  if [[ "${choice}" != [yY] ]]; then
```

ee7 (commented):
Side note: I don't love `fetch-configlet-script` as a filename, but I dont have a better suggestion right now.

scripts/fetch-configlet
SaschaMann: > But I don't know if this is why (it's also strange that master is 10 commits behind main...)

They switched the branch names before the renaming tool was available and probably kept master around to not break workflows that use `@master` for whatever reason.

ErikSchierboom:
> Side note: I don't love fetch-configlet-script as a filename, but I dont have a better suggestion right now.

Yeah me neither. I'm open for suggestions

ErikSchierboom:
This PR seems to be controversial after having spoken to some people, so I'll close this PR. The strategy will instead be to automatically send PRs to tracks to update.

ee7:
> This PR seems to be controversial after having spoken to some people, so I'll close this PR. The strategy will instead be to automatically send PRs to tracks to update.

Sounds reasonable. I was gradually leaning that way too.

I had this reservation: the "fetch-fetch" approach is only a win overall if the "fetch-fetch" script needs updating less often than the plain fetch script. But the "fetch-fetch" approach is less mature and has had fewer eyes on it, so it's likely to require updates if it has any complexity at all, and some complexity is arguably merited.

scripts/fetch-configlet
```diff
@@ -1,58 +1,36 @@
...
-curl "${curlopts[@]}" --output "${output_path}" "${download_url}"
-
-case "${ext}" in
-  *zip) unzip "${output_path}" -d "${output_dir}"   ;;
-  *)    tar xzf "${output_path}" -C "${output_dir}" ;;
-esac
-
-rm -f "${output_path}"
+# Execute the fetch script
+bash -c "${script}"
```
glennj, line 36: Could `exec bash -c "${script}"` here to replace the running process. That's (kind of) the shell's "tailcall" mechanism.

glennj:
> > Side note: I don't love fetch-configlet-script as a filename, but I dont have a better suggestion right now.
> 
> Yeah me neither. I'm open for suggestions

`configlet-fetcher` ?

scripts/fetch-configlet
ee7: Great feedback, thanks. Sorry for the slow reply - still catching up on notifications.

We decided to stick with the `fetch-configlet` status quo for now, but I imagine we'll use your suggestion if we revisit the "fetch-fetch-configlet" idea.

ee7:
> > > Side note: I don't love fetch-configlet-script as a filename, but I dont have a better suggestion right now.
>
> > Yeah me neither. I'm open for suggestions
> 
> `configlet-fetcher` ?

Well, it doesn't have "script" in the filename, so I like it :)

I guess `fetch-fetcher` was possible too.

I think the clearest naming was `fetch-fetch-configlet` and `fetch-configlet`, but we wanted to keep the repo file named `fetch-configlet` for backwards-compatibility, even though it'd perform "fetch-fetch".

PR #140, opened 2021-01-27
Use main instead of master

ErikSchierboom:
This prepares for the renaming of the branches to main instead of master. We should hold off on merging this until the prob-specs repo's default branch has been renamed to `main.`

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Global options:
 
 ## Linting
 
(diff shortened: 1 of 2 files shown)


ee7 (approved):

ee7:
When merging, I'd suggest changing the body of the commit message to something like
```
The main branch of the `v3-docs` repo and the `problem-specifications`
repo is now named `main`, not `master`.
```
to make it clear that it already happened.

PR #145, opened 2021-01-28
build: remove `v` tag prefix

ee7:
I think I have a small preference for tagging without a `v` prefix.


ErikSchierboom (approved):
We used the `v` prefix for most other projects on exercism, but it doesn't really matter.

PR #147, opened 2021-01-28
fetch-configlet: don't show download progress in PowerShell script

ErikSchierboom:


ee7 (approved):
Looks good, as far as I can tell from [these docs](
https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.1#progresspreference).

Approved, assuming that you've tested that it works as expected.

We don't need anything involving `$global`, right? I saw [a comment on Stack Overflow](https://stackoverflow.com/questions/18770723/hide-progress-of-invoke-webrequest#comment108965067_18770794) saying something about that.

ErikSchierboom:
I've indeed tested this. We don't need `$global`.

PR #148, opened 2021-01-28
git: ignore configlet.exe

ErikSchierboom:
This is useful when checking the powershell script locally.


ee7 (approved):
This is fine by me - it makes the our ignoring more equal across platforms. 

I'll eventually suggest a couple of alternative approaches though.

PR #152, opened 2021-01-29
build: support pre-release semantic version strings

ee7:
Unfortunately, Nimble requires that the `version` in a .nimble file only
consists of numbers and the `.` character. Therefore we can't write a
pre-release semantic version string in the .nimble file like
```
version = "4.0.0-alpha.1"
```
because this would make our `nimble install` command fail. Note that
`nimble check` would fail too.

As a workaround during the pre-release phase, this commit adds a file
that we read at compile-time to set the `configlet --version` output; it
can contain an arbitrary version string.

Some rejected alternatives:
- Avoid `nimble install` by cloning the dependencies manually (then the
  `nimble build` command by itself can succeed without error).
- Tag commits that do more than bump a version string.
- Use a version string that is a valid Nimble version string, but not a
  valid semantic version string.

As a bonus, removing the `strdefine` in this commit means that
`configlet --version` no longer relies on building with Nimble or
manually passing `-d:NimblePkgVersion=x.y.z`. That is,
`configlet --version` now outputs the correct version (rather than
"unknown") after compiling with just `nim c src/configlet.nim`.


ErikSchierboom (approved):
Nice! Clean and simple.

ee7:
Note: please don't tag this commit.

I'll make another PR that bumps just the version string to 4.0.0-alpha.1

Then we can merge that and tag it when we want to try making a release.

PR #158, opened 2021-01-29
lint: add initial linting rules

ee7:
Extremely basic to start with.

Bad identifiers: `b1`, `b2`, `helpers`

### DIFF
--- a/src/lint/lint.nim
+++ b/src/lint/lint.nim
@@ -1,5 +1,55 @@
-proc lint* =
-  echo "No linting rules have yet been implemented.\n" &
-       "As we're in the process of implementing the linting rules, please " &
-       "re-run this command regularly to see if your track passes the " &
-       "latest linting rules."
+import std/[json, os, strformat, terminal]
+import ".."/[cli, helpers]
+
+template writeError(description: string, details: string) =
+  stdout.styledWriteLine(fgRed, description & ":")
+  stdout.writeLine(details)
+  stdout.write "\n"
+  result = false
+
+proc isValidTrackConfig(trackDir: string): bool =
+  result = true
+  let configJsonPath = trackDir / "config.json"
+  if fileExists(configJsonPath):
+    try:
+      let j = parseFile(configJsonPath)
+    except:
+      writeError("JSON parsing error", getCurrentExceptionMsg())
+  else:
+    writeError("Missing file", configJsonPath)
+
+proc conceptExerciseFilesExist(trackDir: string): bool =
+  const
+    conceptExerciseFiles = [
+      ".docs" / "hints.md",
+      ".docs" / "instructions.md",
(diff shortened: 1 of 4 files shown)


src/lint/lint.nim
```diff
@@ -1,5 +1,55 @@
-proc lint* =
-  echo "No linting rules have yet been implemented.\n" &
-       "As we're in the process of implementing the linting rules, please " &
-       "re-run this command regularly to see if your track passes the " &
-       "latest linting rules."
+import std/[json, os, strformat, terminal]
+import ".."/[cli, helpers]
+
+template writeError(description: string, details: string) =
```
ErikSchierboom, line 4: At some point we probably want to move this as I expect that we'll be writing errors in other places too, but it is fine for now.

src/lint/lint.nim
```diff
@@ -1,5 +1,55 @@
...
+  result = false
+
+proc isValidTrackConfig(trackDir: string): bool =
+  result = true
+  let configJsonPath = trackDir / "config.json"
+  if fileExists(configJsonPath):
+    try:
+      let j = parseFile(configJsonPath)
+    except:
+      writeError("JSON parsing error", getCurrentExceptionMsg())
```
ErikSchierboom, line 17: For future reference: does the output include the file name too?

src/lint/lint.nim
```diff
@@ -1,5 +1,55 @@
...
+
+  let conceptDir = trackDir / "exercises" / "concept"
+  result = true
+
+  if dirExists(conceptDir):
+    for dir in getSortedSubDirs(conceptDir):
+      for conceptExerciseFile in conceptExerciseFiles:
+        let path = dir / conceptExerciseFile
+        if not fileExists(path):
+          writeError("Missing file", path)
```
ErikSchierboom, line 38: We have to think whether we want to check concept exercises based on the directories, or based on the config.json file's contents. But this is fine for now

src/lint/lint.nim
```diff
@@ -1,5 +1,55 @@
...
+
+  let trackDir = conf.trackDir
+  let b1 = isValidTrackConfig(trackDir)
+  let b2 = conceptExerciseFilesExist(trackDir)
+
+  if b1 and b2:
+    echo """
+Basic linting finished successfully:
+- config.json exists and is valid JSON
+- Every concept exercise has the required .md files and a .meta/config.json file"""
```
ErikSchierboom, line 53: This is temporary, right?

ErikSchierboom (approved):
This is a great beginning! I've tested it on the F# track and it works well.

We should probably open an issue to discuss how we want to present any errors once we're implementing a lot of rules. 

Side note: I really want our error messages to be as explicit as possible, providing the user with as much guidance as needed as linking to official documentation. But that's for the future :)

Great work!

PR #161, opened 2021-01-29
lint: implement concept exercise files rule

ErikSchierboom:
Check if all concepts have their required files present: https://github.com/exercism/docs/blob/main/anatomy/tracks/configlet/linting.md#rule-required-files-are-present

### DIFF
--- a/src/lint/lint.nim
+++ b/src/lint/lint.nim
@@ -37,6 +37,24 @@ proc conceptExerciseFilesExist(trackDir: string): bool =
         if not fileExists(path):
           writeError("Missing file", path)
 
+proc conceptFilesExist(trackDir: string): bool =
+  const
+    conceptFiles = [
+      "about.md",
+      "introduction.md",
+      "links.json",
+    ]
+
+  let conceptsDir = trackDir / "concepts"
+  result = true
+
+  if dirExists(conceptsDir):
+    for dir in getSortedSubDirs(conceptsDir):
+      for conceptFile in conceptFiles:
+        let path = dir / conceptFile
+        if not fileExists(path):
+          writeError("Missing file", path)
+
 proc lint*(conf: Conf) =
   echo "The lint command is under development.\n"  &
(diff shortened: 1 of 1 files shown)


ee7 (commented):
The spec says

> Each concept listed in the config.json should have the following files:

> `concepts/<slug>/about.md`
> `concepts/<slug>/introduction.md`
> `concepts/<slug>/links.json`

but this PR doesn't parse the `config.json`. Will we change the spec here?

I interpreted the previous

> The Concept Exercise specific files that must be present are:

> `exercises/concept/<slug>/.docs/hints.md`
> `exercises/concept/<slug>/.docs/instructions.md`
> `exercises/concept/<slug>/.docs/introduction.md`
> `exercises/concept/<slug>/.meta/config.json`

to mean "don't parse `config.json` to determine the required concept exercise files, but do parse it to determine the required files for concepts".

ErikSchierboom:
Ah, that is definitely not an intentional difference as they should be treated the same way. There are a couple of options:

1. Iterate over the sub-directories and check them. This has the benefit of catching any exercises that are being developed but have not yet been added to the config.json. The downside is that it is no longer possible to have any other directories in that directory, but I guess that's okay.
2. Parse the config.json file and check based on its contents. This has the benefit that this corresponds to how things will be interpreted by the website, as it will only look at the config.json file's contents.
3. Do a combination of the above and warn when they are out of sync. This is likely the best option, as it is not uncommon for contributors to forget to add things to the config.json file.

So in my mind, we should do option 3 so the current PR's implementation allows for that, even though it doesn't yet implement the config.json parsing.

The spec needs updating.

ee7 (approved):

ee7:
When merging, I think you should change the commit message. It's currently
```
lint: implement concept exercise files rule
```

But this PR is about concepts, not concept exercises, right?

ee7 (approved):

PR #162, opened 2021-01-29
ci: only run tests when source changes

### LINKED ISSUE #141: Only run the tests when a source file has changed
To speed up CI builds, we should only run the `tests` workflow when a source file has changed. This includes:

- Any `.nim` file modification
- Any `.nimble` file modification
- Any `.nims` file modification
- Any `.github/bin/*` file modification
- Any `.github/workflows/tests.yml` file modification

Anything more?

ErikSchierboom:
To prevent the tests workflow from running unnecessarily (which is quite slow due to the slowness of the mac build), we'll only run it when either:

1. Source code changes: *.nim, *.nims, or *.nimble files
2. GitHub workflows change: .github/bin/* or .github/workflows/* files

I've verified this on my fork: https://github.com/ErikSchierboom/canonical-data-syncer/actions

Closes #141

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,6 +1,13 @@
 name: tests
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+    paths:
+      - "!**.md"
+  pull_request:
+    paths:
+      - "!**.md"
+  workflow_dispatch:
 
 jobs:
   all_tests:


.github/workflows/tests.yml
```diff
@@ -1,6 +1,21 @@
 name: tests
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+    paths:
+      - "**.nim"
```
ErikSchierboom, line 6: See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#patterns-to-match-file-paths for examples of what this pattern does

ee7 (commented):
`"**.cfg"` could be another candidate for inclusion too, but we don't have one - it's fine.

I have some reservations that this could become outdated easily: I can imagine myself adding a new file, or moving some around, and not noticing that the test check is skipped (especially if there are many PRs, or more jobs/workflows in the future).

Maybe it would be safer to have the reverse filter? That is, an explicit list of files to skip.

I think the list of files that are effectively skipped by this PR is:

```
.github/CODEOWNERS
.github/dependabot.yml
.gitignore
LICENSE
README.md
configlet.version
scripts/fetch-configlet
scripts/fetch-configlet.ps1
```

That list is a bit longer than the list in the PR currently, and it might grow more quickly. But the failure mode seems better: a build when you didn't want one, rather than a PR passing when it should be failing.

What do you think?

ErikSchierboom:
Yeah, there is a trade-off. Let's just simplify things and _not_ run the tests when just Markdown files have changed. I have to look into how to encode that in the workflow.

ErikSchierboom:
I'm closing this in favor of https://github.com/exercism/configlet/pull/170 due to me having deleted my fork, which I apparently submitted the PR from.

Issue #16, opened 2020-10-08
Add stylecheck to build


ErikSchierboom:

ee7:
`--styleCheck:error` is currently too inconvenient.

But the development version of Nim has a new command-line option that allows turning a hint into an error. So one way to resolve this issue is to wait until Nim 1.6.0 and then use `--hintAsError[Name]:on` to get the best of both worlds.

See https://github.com/exercism/configlet/pull/305.

PR #166, opened 2021-01-30
build(config): add `--styleCheck:hint`

ee7:
This PR tries to address https://github.com/exercism/configlet/issues/16, but doesn't do so completely. The latest release of Nimble (0.12.0, in Oct 2020) adds `--hints:off` by default, and that's set after `--hint[Name]:on` can be passed.

From https://github.com/nim-lang/nimble/blob/master/changelog.markdown#0120:
> Calls to the Nim compiler now display --hints:off output by default. The --verbose flag will print the full Nim command as well as regular compiler output.

So when running `nimble build` we can only see the styleCheck output if `--verbose` was passed. However, this commit is also useful to turn on styleCheck by default for a user who runs `nim c foo.nim` in our project. 

`--styleCheck:error` would be possible too, but it's currently too
aggressive: it would currently produce an error if one of our Nimble
dependencies has a style problem.

To-do:
- [X] Merge #164 and #165, and rebase this PR on top.

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -1,3 +1,6 @@
+switch("styleCheck", "hint")
+hint("Name", on)
+
 if defined(release):
   switch("opt", "size")
   switch("passC", "-flto")


config.nims
```diff
@@ -1,10 +1,14 @@
-switch("opt", "size")
-switch("passC", "-flto")
+switch("styleCheck", "hint")
```
ErikSchierboom, line 1: Too bad that we need it to be a hint at the moment, but better than nothing!

ErikSchierboom (approved):

PR #167, opened 2021-01-30
build(config): add `--experimental:strictFuncs`

ee7:
This PR means that the Nim compiler will enforce that we don't have code like
```Nim
type
  Foo = ref object
    x: int

func mutate(foo: Foo) =
  foo.x = 2

let foo = Foo(x: 1)
mutate(foo)
echo foo.x
```

Without strictFuncs, the above code compiles without complaint (despite `mutate` being a `func`, and `foo` being declared with `let`), and outputs `2`. But compiling with `nim c --experimental:strictFuncs`, it produces an error:
```
/tmp/foo.nim(5, 6) Error: 'mutate' can have side effects
an object reachable from 'foo' is potentially mutated
/tmp/foo.nim(6, 6) the mutation is here
```

The code should instead be written as

```diff
 type
   Foo = ref object
     x: int

-func mutate(foo: Foo) =
+func mutate(foo: var Foo) =
   foo.x = 2

-let foo = Foo(x: 1)
+var foo = Foo(x: 1)
 mutate(foo)
 echo foo.x
```

See:
- https://nim-lang.github.io/Nim/manual_experimental.html#strict-funcs
- https://nim-lang.org/blog/2020/09/01/write-tracking.html

To-do:
- [X] Merge #164, #165 and #166 and rebase this PR on top.

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -1,5 +1,6 @@
 switch("styleCheck", "hint")
 hint("Name", on)
+switch("experimental", "strictFuncs")
 
 if defined(release):
   switch("opt", "size")


ErikSchierboom (approved):
This is definitely something that I would want the compiler to catch. It being marked as experimental, is there any chance of false positives?

ee7:
> It being marked as experimental, is there any chance of false positives?

Our codebase currently compiles with this flag, and I hope there won't be a `strictFuncs` regression in the stable release of Nim; code that compiles with `--experimental:strictFuncs` in Nim 1.4.2 shouldn't produce a false positive in a later release. But any false positive would just give us an error at compile-time - it doesn't affect the end user of the binary.

I think it's safe to add this flag for now - we can always remove it.

Furthermore, I personally use the development version of Nim, so I should notice any new strictFuncs regression long before it affects this repo. And I intend to eventually add `devel` Nim to CI in this repo (probably just with Linux, if the macOS build times are still long), which would also give us a heads-up.

ErikSchierboom:
Sounds good!

PR #168, opened 2021-02-02
lint: refactor logic of finding missing files

ee7:
Also remove an unneeded import and fix trivial style issues.

### DIFF
--- a/src/lint/lint.nim
+++ b/src/lint/lint.nim
@@ -1,4 +1,4 @@
-import std/[json, os, strformat, terminal]
+import std/[json, os, terminal]
 import ".."/[cli, helpers]
 
 template writeError(description: string, details: string) =
@@ -18,7 +18,23 @@ proc isValidTrackConfig(trackDir: string): bool =
   else:
     writeError("Missing file", configJsonPath)
 
+proc subdirsContain(dir: string, files: openArray[string]): bool =
+  ## Returns `true` if every file in `files` exists in every subdirectory of
+  ## `dir`.
+  ##
+  ## Returns `true` if `dir` does not exist or has no subdirectories.
+  result = true
+
+  if dirExists(dir):
+    for subdir in getSortedSubdirs(dir):
+      for file in files:
+        let path = subdir / file
+        if not fileExists(path):
+          writeError("Missing file", path)
+
 proc conceptExerciseFilesExist(trackDir: string): bool =
+  ## Returns true if every subdirectory in `trackDir/exercises/concept` has the
+  ## required files.
   const
     conceptExerciseFiles = [
       ".docs" / "hints.md",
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):

ee7:
@ErikSchierboom can you review again? I tweaked it a bit to:
- Prefer `proc` over `template`
- Try to improve the function name
- Try to Improve the comments

src/lint/lint.nim
```diff
@@ -18,7 +18,23 @@ proc isValidTrackConfig(trackDir: string): bool =
...
+  ##
+  ## Returns `true` if `dir` does not exist or has no subdirectories.
+  result = true
+
+  if dirExists(dir):
+    for subdir in getSortedSubdirs(dir):
+      for file in files:
+        let path = subdir / file
+        if not fileExists(path):
+          writeError("Missing file", path)
```
ErikSchierboom, line 33: Should this be:

```suggestion
          writeError("Missing file", path)
          result = false
```

src/lint/lint.nim
```diff
@@ -28,16 +44,11 @@ proc conceptExerciseFilesExist(trackDir: string): bool =
...
   let conceptDir = trackDir / "exercises" / "concept"
-  result = true
-
-  if dirExists(conceptDir):
-    for dir in getSortedSubDirs(conceptDir):
-      for conceptExerciseFile in conceptExerciseFiles:
-        let path = dir / conceptExerciseFile
-        if not fileExists(path):
-          writeError("Missing file", path)
+  result = subdirsContain(conceptDir, conceptExerciseFiles)
```
ErikSchierboom, line 47: A general question: when should one assign to `result` instead of doing `return`?

ee7: That's a good question. It's partly a matter of personal taste.

The [official style guide](https://nim-lang.github.io/Nim/nep1.html#introduction-coding-conventions) contains:
> The `return` statement should ideally be used when its control-flow properties are required. Use a procedure's implicit `result` variable whenever possible. This improves readability.

That's a common style, which optimizes for making immediate `return` obvious. However, Nim style is still evolving: some people argue that we should use `return` more often, and use `result =` only when building up a return value over multiple lines.

We also have the option of omitting `result =` or `return` entirely here, writing just
```Nim
subdirsContain(conceptDir, conceptExerciseFiles)
```

The parent proc then returns the result of that call (the compiler enforces that it returns a `bool` as declared in the proc signature).

That style can improve readability/robustness in some places, but here I find it clearest to assign to `result` - we're doing the call because we're interested in its return value, rather than to perform some other action.

Let me know if I can clarify further.

You might like to read https://peterme.net/tips-and-tricks-with-implicit-return-in-nim.html - probably one of the best discussions on this topic.

ErikSchierboom: That _does_ help! Thanks

src/lint/lint.nim
ee7: No, because the current `writeError` is:
https://github.com/exercism/configlet/blob/a7f090efeef1409b7edb7c430a474445175774f2/src/lint/lint.nim#L4-L8

Due to how templates work, `writeError` is already setting the `result` of the `subdirsContain` proc to `false`. 

Try running `configlet lint` with this PR using some track dir that has missing files. You should see that we detect them, but the only place that can set `result` to `false` is inside the `writeError` template.

I might refactor this pattern later though. I haven't settled on the most expressive/maintainable way to write the linting code, but this is one alternative to writing something like:
```Nim
proc isValidConfigFile(...): bool =
  result = true
  if not isValidFoo(...):
    result = false
  if not isValidBar(...):
    result = false
```
everywhere, or passing `result` as a `var` parameter to each validator function.

See the use of `writeError` in https://github.com/exercism/configlet/pull/169/ - in particular, how the `result` of `isValidConceptExerciseConfig` can become `false`. I don't claim this is the best approach, though.

We could probably improve the `writeError` name - any ideas?

ErikSchierboom: Ah, now I understand. I think the original pattern is fine, but maybe with some rename as suggested. But what to use as the name... 🤔 The only thing I could come up with is `handleError`, but I don't know.

PR #170, opened 2021-02-03
ci: don't run tests when markdown files are changed

ErikSchierboom:
See https://github.com/ErikSchierboom/configlet/actions for a working example.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,6 +1,15 @@
 name: tests
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+    paths:
+      - "**"
+      - "!**.md"
+  pull_request:
+    paths:
+      - "**"
+      - "!**.md"
+  workflow_dispatch:
 
 jobs:
   all_tests:


ee7 (approved):
Nice.

LGTM, but I'd suggest we amend the commit message to something like
```
ci: don't run tests workflow for Markdown-only changes
```
with the body

```
To prevent the tests workflow from running unnecessarily we don't run it
when only Markdown files are changed
```

This adds the word "only" to both - the previous wordings imply that we don't run tests for any commit that includes a change to a .md file.

And nitpick: can we wrap the commit message body at 72 chars?

ErikSchierboom:
I've updated the commit message.

> And nitpick: can we wrap the commit message body at 72 chars?

With due respect, but I won't be doing this. We don't enforce this anywhere on Exercism and I personally don't feel that it offers much.

ee7:
> > And nitpick: can we wrap the commit message body at 72 chars?
> 
> With due respect, but I won't be doing this. We don't enforce this anywhere on Exercism and I personally don't feel that it offers much.

No problem.

By the way, no need to force-push an amended commit if there's zero code change - I'd be fine if you added the "only" while you clicked "squash + merge".

ErikSchierboom:
Ah okay.

PR #169, opened 2021-02-02
lint: begin linting concept exercise `config.json`

ee7:
WIP proof-of-concept with templates, rather than a macro as done previously.

The error messages are currently bad. I'll write more later.

PR #171, opened 2021-02-03
lint: add linting of concepts `links.json`

ee7:
This PR implements [the `concept/<slug>/links.json` part of the spec](https://github.com/exercism/docs/blob/6ff54c5db80c21cb021d28d65b2528a72d9374de/building/configlet/lint.md#rule-conceptsluglinksjson-is-valid):

>- The file must be valid JSON
>- The JSON root must be an array
>- The `"[].url"` property is required
>- The `"[].url"` value must be an URL
>- The `"[].description"` property is required
>- The `"[].description"` value must be a non-empty, non-blank string
>- The `"[].icon_url"` property is optional
>- The `"[].icon_url"` value must be an URL

To-do:
- [x] Merge #169 and rebase this PR on top
- [X] Call `isEveryConceptLinksFileValid` from `lint.nim`
- [X] Improve the URL checking URL validity slightly. We make it stricter later.
- [X] Rebase on top of the new non-template code
- [x] Merge #218 and rebase this PR on top.
- [X] Improve error messages.
- [X] Check the diff that this PR causes to the `configlet lint` output on every track

### DIFF
--- a/src/lint/concepts.nim
+++ b/src/lint/concepts.nim
@@ -3,15 +3,11 @@ import ".."/helpers
 import "."/validators
 
 proc isUrlLike(s: string): bool =
-  ## Returns true if `s` starts with `http://`, `https://` or `www`.
+  ## Returns true if `s` starts with `https://`, `http://` or `www`.
   # For now, this is deliberately simplistic. We probably don't need
   # sophisticated URL checking, and we don't want to use Nim's stdlib regular
   # expressions because that would add a dependency on PCRE.
-  if s.startsWith("http"):
-    if s.continuesWith("://", 4) or s.continuesWith("s://", 4):
-      result = true
-  elif s.startsWith("www"):
-    result = true
+  s.startsWith("https://") or s.startsWith("http://") or s.startsWith("www")
 
 proc isValidLinkObject(data: JsonNode, context: string, path: string): bool =
   ## Returns true if `data` is a `JObject` that satisfies all of the below:
--- a/src/lint/concepts.nim
+++ b/src/lint/concepts.nim
@@ -4,15 +4,15 @@ import "."/validators
 
 proc isUrlLike(s: string): bool =
   ## Returns true if `s` starts with `https://`, `http://` or `www`.
-  # For now, this is deliberately simplistic. We probably don't need
-  # sophisticated URL checking, and we don't want to use Nim's stdlib regular
-  # expressions because that would add a dependency on PCRE.
+  # We probably only need simplistic URL checking, and we want to avoid using
+  # Nim's stdlib regular expressions in order to avoid a dependency on PCRE.
   s.startsWith("https://") or s.startsWith("http://") or s.startsWith("www")
 
 proc isValidLinkObject(data: JsonNode, context: string, path: string): bool =
   ## Returns true if `data` is a `JObject` that satisfies all of the below:
   ## - has a `url` key, with a value that is a URL-like string.
-  ## - has a `description` key, with a value that is a non-empty, non-blank string.
+  ## - has a `description` key, with a value that is a non-empty, non-blank
+  ##   string.
   ## - if it has a `icon_url` key, the corresponding value is a URL-like string.
   if isObject(data, context, path):
     result = true
@@ -54,7 +54,8 @@ proc isEveryConceptLinksFileValid*(trackDir: string): bool =
             try:
               parseFile(linksPath) # Shows the filename in the exception message.
             except CatchableError:
-              result.setFalseAndPrint("JSON parsing error", getCurrentExceptionMsg())
+              result.setFalseAndPrint("JSON parsing error",
+                                      getCurrentExceptionMsg())
               continue
           if not isValidLinksFile(j, linksPath):
             result = false
--- a/src/lint/concepts.nim
+++ b/src/lint/concepts.nim
@@ -1,5 +1,80 @@
-import std/os
+import std/[json, os, strutils]
+import ".."/helpers
 import "."/validators
+
+proc isUrlLike(s: string): bool =
+  ## Returns true if `s` starts with `http://`, `https://` or `www`.
+  # For now, this is deliberately simplistic. We probably don't need
+  # sophisticated URL checking, and we don't want to use Nim's stdlib regular
+  # expressions because that would add a dependency on PCRE.
+  if s.startsWith("http"):
+    if s.continuesWith("://", 4) or s.continuesWith("s://", 4):
+      result = true
+  elif s.startsWith("www"):
+    result = true
+
+proc isValidLinkObject(data: JsonNode, context: string, path: string): bool =
+  ## Returns true if `data` is a `JObject` that satisfies all of the below:
+  ## - has a `url` key, with a value that is a URL-like string.
+  ## - has a `description` key, with a value that is a non-empty, non-blank string.
+  ## - if it has a `icon_url` key, the corresponding value is a URL-like string.
+  if isObject(data, context, path):
+    result = true
+
+    if checkString(data, "url", path):
+      let s = data["url"].getStr()
+      if not isUrlLike(s):
+        result.setFalseAndPrint("Not a valid URL: " & s, path)
+    else:
+      result = false
+
+    if not checkString(data, "description", path):
+      result = false
+
+    if data.hasKey("icon_url"):
+      if checkString(data, "icon_url", path, isRequired = false):
+        let s = data["icon_url"].getStr()
+        if not isUrlLike(s):
+          result.setFalseAndPrint("Not a valid URL: " & s, path)
+      else:
+        result = false
+  else:
+    result.setFalseAndPrint("At least one element of the top-level array is " &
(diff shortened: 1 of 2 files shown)


ee7:
@ErikSchierboom Just to confirm, what should `configlet lint` do if it encounters:
1. A `links.json` file that is empty? For example, [here](https://github.com/exercism/ruby/blob/61133cdf179b45b58af734888300393358896021/concepts/enumeration/links.json) - there are 4 of these in total across Exercism. Currently, this PR prints 
```
JSON parsing error:
./concepts/enumeration/links.json(1, 0) Error: { expected
```
In the future we could detect empty files and print a better error message.

2. A `links.json` file that is an empty array? For example, [here](https://github.com/exercism/csharp/blob/a27e910807e22bb7cb523296ce686200562412ab/concepts/const-readonly/links.json) there are 238 of these in total across Exercism. Right now this PR prints e.g.
```
Array is empty: root:
./concepts/const-readonly/links.json
```
but it should't complain about this, right?

ErikSchierboom:
> A links.json file that is empty?

That is an error.

> A links.json file that is an empty array?

That's fine.

ee7:
This PR causes the following diff to the `configlet lint` output, per track:

#### clojure
```diff
+File is empty, but must contain at least the empty array, `[]`:
+./concepts/basics/links.json
+
```

#### haxe
```diff
+Not an array: root:
+./concepts/abstract-types/links.json
+
+Not an array: root:
+./concepts/array-comprehension/links.json
+
+Not an array: root:
+./concepts/arrays/links.json
+
+Not an array: root:
+./concepts/basics/links.json
+
+Not an array: root:
+./concepts/bit-manipulation/links.json
+
+Not an array: root:
+./concepts/chars/links.json
+
+Not an array: root:
+./concepts/classes/links.json
+
+Not an array: root:
+./concepts/constructors/links.json
+
+Not an array: root:
+./concepts/datetimes/links.json
+
+Not an array: root:
+./concepts/do-while-loops/links.json
+
+Not an array: root:
+./concepts/enums/links.json
+
+Not an array: root:
+./concepts/equality/links.json
+
+Not an array: root:
+./concepts/exceptions/links.json
+
+Not an array: root:
+./concepts/final/links.json
+
+Not an array: root:
+./concepts/flag-enums/links.json
+
+Not an array: root:
+./concepts/floating-point-numbers/links.json
+
+Not an array: root:
+./concepts/for-loops/links.json
+
+Not an array: root:
+./concepts/foreach-loops/links.json
+
+Not an array: root:
+./concepts/generic-types/links.json
+
+Not an array: root:
+./concepts/if-statements/links.json
+
+Not an array: root:
+./concepts/indexers/links.json
+
+Not an array: root:
+./concepts/inheritance/links.json
+
+Not an array: root:
+./concepts/integral-numbers/links.json
+
+Not an array: root:
+./concepts/interfaces/links.json
+
+Not an array: root:
+./concepts/iterators/links.json
+
+Not an array: root:
+./concepts/lambda/links.json
+
+Not an array: root:
+./concepts/lists/links.json
+
+Not an array: root:
+./concepts/maps/links.json
+
+Not an array: root:
+./concepts/nullability/links.json
+
+Not an array: root:
+./concepts/numbers/links.json
+
+Not an array: root:
+./concepts/operator-overloading/links.json
+
+Not an array: root:
+./concepts/option-type/links.json
+
+Not an array: root:
+./concepts/optional-arguments/links.json
+
+Not an array: root:
+./concepts/pattern-matching/links.json
+
+Not an array: root:
+./concepts/properties/links.json
+
+Not an array: root:
+./concepts/randomness/links.json
+
+Not an array: root:
+./concepts/reflection/links.json
+
+Not an array: root:
+./concepts/regular-expressions/links.json
+
+Not an array: root:
+./concepts/rest-args/links.json
+
+Not an array: root:
+./concepts/string-buffer/links.json
+
+Not an array: root:
+./concepts/switch-expressions/links.json
+
+Not an array: root:
+./concepts/templates/links.json
+
+Not an array: root:
+./concepts/ternary-operators/links.json
+
+Not an array: root:
+./concepts/throw-expressions/links.json
+
+Not an array: root:
+./concepts/time/links.json
+
+Not an array: root:
+./concepts/typedefs/links.json
+
+Not an array: root:
+./concepts/using-statements/links.json
+
+Not an array: root:
+./concepts/while-loops/links.json
```

#### java
```diff
+Missing key: 'url':
+./concepts/for-loops/links.json
+
+Missing key: 'url':
+./concepts/foreach-loops/links.json
```

#### python
```diff
+Missing key: 'url':
+./concepts/numbers/links.json
```

#### ruby
```diff
+
+File is empty, but must contain at least the empty array, `[]`:
+./concepts/enumeration/links.json
+
+File is empty, but must contain at least the empty array, `[]`:
+./concepts/exceptions/links.json
+
+File is empty, but must contain at least the empty array, `[]`:
+./concepts/ostruct/links.json
```

#### swift
```diff
+
+Missing file:
+./concepts/characters/links.json
+
+Missing file:
+./concepts/classes/links.json
+
+Missing file:
+./concepts/initializers/links.json
+
+Missing file:
+./concepts/opaque-indices/links.json
+
+Missing key: 'url':
+./concepts/shorthand-arguments/links.json
+
+Missing key: 'description':
+./concepts/shorthand-arguments/links.json
+
+Missing file:
+./concepts/strings/links.json
+
+Missing file:
+./concepts/structs/links.json
```

src/lint/concepts.nim
```diff
@@ -1,6 +1,81 @@
...
+      let s = data["url"].getStr()
+      if not isUrlLike(s):
+        result.setFalseAndPrint("Not a valid URL: " & s, path)
+    else:
+      result = false
+
+    if not checkString(data, "description", path):
+      result = false
+
+    if data.hasKey("icon_url"):
```
ErikSchierboom, line 34: Do you need this call? Doesn't `checkString` gracefully handle optional values?

src/lint/concepts.nim
```diff
@@ -1,6 +1,81 @@
...
+proc isValidLinksFile(data: JsonNode, path: string): bool =
+  result = isArrayOf(data, "", path, isValidLinkObject, isRequired = false)
+
+proc isNonBlank(path: string): bool =
+  ## Returns true if `path` points to a file that has at least one
+  ## non-whitespace character.
+  let contents = readFile(path)
+  for c in contents:
+    if c notin Whitespace:
+      return true
```
ErikSchierboom, line 54: Why not use `isEmptyOrWhitespace`? https://nim-lang.org/docs/strutils.html#isEmptyOrWhitespace%2Cstring It is more readable I think to use `isEmptyOrWhitespace`

ErikSchierboom (approved):
Two minor nits

src/lint/concepts.nim
ee7: > Do you need this call?

With the current `checkString`, unfortunately yes.

The full context is:
```Nim
    if data.hasKey("icon_url"):
      if checkString(data, "icon_url", path, isRequired = false):
        let s = data["icon_url"].getStr()
```

`icon_url` is optional, so we call `checkString` with `isRequired = false`. That means that `checkString` returns `true` if `icon_url` is missing, which means the `data["icon_url"].getStr()` would throw an exception if we don't have the first line.

It is indeed pretty ugly, and we're performing `hasKey` twice.

In the future I'm thinking about splitting `checkString` into `hasString` and `isString` (and the same for `checkBoolean`, and `checkInteger`), similar to the previous split of `checkArray` into `hasArray` and `isArray`.

src/lint/concepts.nim
```diff
@@ -1,6 +1,81 @@
...
 import "."/validators
 
+proc isUrlLike(s: string): bool =
+  ## Returns true if `s` starts with `http://`, `https://` or `www`.
+  # For now, this is deliberately simplistic. We probably don't need
+  # sophisticated URL checking, and we don't want to use Nim's stdlib regular
+  # expressions because that would add a dependency on PCRE.
+  if s.startsWith("http"):
+    if s.continuesWith("://", 4) or s.continuesWith("s://", 4):
+      result = true
```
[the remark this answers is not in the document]
ee7: OK - added https://github.com/exercism/configlet/pull/171/commits/f83cd29f7787ce8abdebd1c4996a79c2c8ec17ef which handles the `startsWith("www")` case and makes it even more readable.

It turns out that about 97% of the URLs in `links.json` files so far start with `https://` anyway:
```
$ find -name links.json -exec grep '"url":\s"https://' {} \; | wc -l
924
$ find -name links.json -exec grep '"url":\s"http://' {} \; | wc -l
30
$ find -name links.json -exec grep '"url":\s"www' {} \; | wc -l
0
```

src/lint/concepts.nim
ee7: Good idea. Added https://github.com/exercism/configlet/pull/171/commits/ee6081c199fc6923359924e31d1485f5b9e4aea7.

I removed the `isNonBlank` proc partly because I'll refactor later so that we don't read the file twice.

PR #176, opened 2021-02-08
tests(sync): add smoke test

ee7:
This PR tries to demonstrate that the fix in #175 is correct.

We didn't have tests for `configlet sync` yet. One reason is that they were awkward to implement until `configlet` learned the `-t, --track-dir` option, which happened relatively recently. 

Todo:
- [x] Merge #175 and rebase this PR on top.

### DIFF
--- a/tests/test_binary.nim
+++ b/tests/test_binary.nim
@@ -39,19 +39,23 @@ proc testsForSync(binaryPath: string) =
     removeDir(psDir)
     removeDir(trackDir)
 
-    test "can clone problem-specifications for test setup":
+    # Setup: clone the problem-specifications repo
+    block:
       execAndCheck(0):
         cloneExercismRepo("problem-specifications", psDir)
 
-    test "can clone track repo for test setup":
+    # Setup: clone a track repo
+    block:
       execAndCheck(0):
         cloneExercismRepo("nim", trackDir)
 
-    test "can set the problem-specifications repo to a known state":
+    # Setup: set the problem-specifications repo to a known state
+    block:
       execAndCheck(0):
         execCmdEx(&"git -C {psDir} checkout ef9e1e17c84721ee6e9d5a65c8dd3ba2122eac91")
 
-    test "can set the track repo to a known state":
+    # Setup: set the track repo to a known state
+    block:
       execAndCheck(0):
         execCmdEx(&"git -C {trackDir} checkout 798a250a5baf44468ff39bf016fafc3c6a5375c2")
 
--- a/tests/test_binary.nim
+++ b/tests/test_binary.nim
@@ -7,6 +7,260 @@ const
     else: ""
   binaryName = "configlet" & binaryExt
 
+proc cloneExercismRepo(repoName, dest: string; isShallow = false): tuple[
+    output: string; exitCode: int] =
+  # Clones the Exercism repo named `repoName` to the location `dest`.
+  let opts = if isShallow: "--depth 1" else: ""
+  let url = &"https://github.com/exercism/{repoName}/"
+  let cmd = &"git clone {opts} {url} {dest}"
+  result = execCmdEx(cmd)
+
+template execAndCheck(expectedExitCode: int; body: untyped) {.dirty.} =
+  ## Runs `body`, and prints the output if the exit code is non-zero.
+  ## `body` must have the same return type as `execCmdEx`.
+  let (outp, exitCode) = body
+  if exitCode != expectedExitCode:
+    echo outp
+    fail()
+
+func conciseDiff(s: string): string =
+  ## Returns the lines of `s` that begin with a `+` or `-` character.
+  result = newStringOfCap(s.len)
+  for line in s.splitLines():
+    if line.len > 0:
+      if line[0] in {'+', '-'}:
+        result.add line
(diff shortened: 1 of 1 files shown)


tests/test_binary.nim
```diff
@@ -7,6 +7,260 @@ const
...
+  if exitCode != expectedExitCode:
+    echo outp
+    fail()
+
+func conciseDiff(s: string): string =
+  ## Returns the lines of `s` that begin with a `+` or `-` character.
+  result = newStringOfCap(s.len)
+  for line in s.splitLines():
+    if line.len > 0:
+      if line[0] in {'+', '-'}:
```
ErikSchierboom, line 31: Minor style suggestion: I prefer to reduce nesting when possible, or is that against the Nim style guide?

```suggestion
    if line.len > 0 and line[0] in {'+', '-'}:
```

tests/test_binary.nim
```diff
@@ -7,6 +7,260 @@ const
...
+      execAndCheck(0):
+        cloneExercismRepo("nim", trackDir)
+
+    test "can set the problem-specifications repo to a known state":
+      execAndCheck(0):
+        execCmdEx(&"git -C {psDir} checkout ef9e1e17c84721ee6e9d5a65c8dd3ba2122eac91")
+
+    test "can set the track repo to a known state":
+      execAndCheck(0):
+        execCmdEx(&"git -C {trackDir} checkout 798a250a5baf44468ff39bf016fafc3c6a5375c2")
```
ErikSchierboom, line 56: I assume tests are executed in order? I was slightly confused by these tests. It looks like we are testing executing git commands. I assume that you've done this in order to get nice error messages when a step fails?

ErikSchierboom (approved):
Great to have some tests for the syncing!

tests/test_binary.nim
ee7: There isn't an official style rule one way or the other. I'm fine with either way, but lately I've been leaning towards extra nesting.

There is at least [one (opinionated) Nim style guide](https://github.com/disruptek/disruptek/blob/decfd8711a1f49e5da279e65abcb9f3245721099/style.md#super-dominating-conditionals) that argues that nesting conditionals is more readable. I'll let you decide :)

A little related is a discussion between
```Nim
if not cond:
  return

foo1
foo2
foo3
```

vs

```Nim
if cond:
  foo1
  foo2
  foo3
```

where I can see the argument to prefer the latter - see Araq's post in https://forum.nim-lang.org/t/7130

tests/test_binary.nim
ee7: Yes, they're executed in order. I agree they're confusing - it does look like we're just testing `git`.

I'll try to refactor so it's clearer - I didn't love how they looked in the test output either:

```
[Suite] sync
  [OK] can clone problem-specifications for test setup
  [OK] can clone track repo for test setup
  [OK] can set the problem-specifications repo to a known state
  [OK] can set the track repo to a known state
  [OK] `sync --check` exits with 1 and prints the expected output
  [OK] `sync --mode=include` exits with 0 and includes the expected test cases
  [OK] `git diff` shows the expected diff
  [OK] after syncing, another `sync --mode=include` performs no changes
  [OK] after syncing, `sync --check` shows that exercises are up to date
  [OK] the `git diff` output is still the same
```

Probably using one setup step. Would you like to see it in the log like the below, or not at all? 
```
[OK] can prepare a state for a track repo and problem-specifications 
```

But yes, it's possible to have good error messages without showing all the setup steps in the log.

tests/test_binary.nim
ErikSchierboom: 🤷 I'm fine with keeping the indentation here. We'll just evaluate per case what the best nesting is.

tests/test_binary.nim
ErikSchierboom: > Would you like to see it in the log like the below, or not at all?

My natural tendency would be to omit it from the log as it is not really a test. If something goes wrong, there will be an error message anyway, right?

ee7: > My natural tendency would be to omit it from the log as it is not really a test. 

Sounds good.

>  If something goes wrong, there will be an error message anyway, right?

Yes.

ee7: Addressed feedback by replacing `test` with `block`.

In the current implementation, we need the new scope because `execAndCheck` is a `dirty` template so that `outp` is available afterwards, for example here: 
```Nim
    test "`git diff` shows the expected diff":
      execAndCheck(0):
        execCmdEx(diffCmd)

      check outp.conciseDiff() == expectedDiffOutput
```

The test output now look like this: https://github.com/exercism/configlet/runs/1936097992#step:6:158

```
[Suite] sync
  [OK] `sync --check` exits with 1 and prints the expected output
  [OK] `sync --mode=include` exits with 0 and includes the expected test cases
  [OK] `git diff` shows the expected diff
  [OK] after syncing, another `sync --mode=include` performs no changes
  [OK] after syncing, `sync --check` shows that exercises are up to date
  [OK] the `git diff` output is still the same
```

ErikSchierboom: Excellent@

ErikSchierboom:
LGTM!

PR #183, opened 2021-02-18
lint: add more linting rules for track config.json

ErikSchierboom:
This PR adds some more linting rules for the track's config.json.

I'm not really happy with our error messages, but we can improve that later on. At this stage, it is important to just have some validation.

- refactor: create module for track config.json validation
- lint: add validator to check boolean property
- lint: add validator to check integer property
- lint: validate basic property of track config.json file
- lint: valid tags property of track config.json file
- lint: update output of applied rules


ee7 (approved):
Non-blocking comment: Can we think of a better name than `track.nim`? I think:
- it's too close to `src/sync/tracks.nim`
- it sounds like it handles linting the whole track, when actually it just handles the root `config.json` file

I'm fine with something like these:
- `track_config.nim`
- `root_config.nim`

or whatever you like.

Otherwise LGTM - approved, conditional on squashing at merge-time.

src/lint/track.nim
```diff
@@ -38,42 +38,40 @@ let tags = @[
   "used_for/robotics",
   "used_for/scientific_calculations",
   "used_for/scripts",
-  "used_for/web_development"
-]
+  "used_for/web_development",
+].toHashSet()
```
ErikSchierboom, line 42: Doh. I should have come up with this myself 🤦

ErikSchierboom:
I'm fine with `track_config.nim`. Can you update?

ErikSchierboom:
Squashed and merged

PR #185, opened 2021-02-21
tests(probspecs): fix for `accumulate` as new first exercise

valentin-p:
- The first exercise of problem specs is now `accumulate`
- New problem spec has been added in https://github.com/exercism/problem-specifications/pull/1738 
- `nimble test` succeeds
- `nim c --styleCheck:error -r ./tests/all_tests.nim` succeeds

The error prior this fix was:
```
[Suite] findProbSpecsExercises: fresh clone
  [OK] can return the exercises
    configlet\tests\test_probspecs.nim(41, 24): Check failed: exercise.slug == "acronym"
    exercise.slug was accumulate
  [FAILED] the first exercise is as expected
```

### DIFF
--- a/tests/test_probspecs.nim
+++ b/tests/test_probspecs.nim
@@ -38,11 +38,34 @@ proc main =
         let exercise = probSpecsExercises[0]
 
         check:
-          exercise.slug == "acronym" # The first exercise with canonical data.
+          exercise.slug == "accumulate" # The first exercise with canonical data.
+          exercise.testCases.len >= 5 # Tests are never removed.
+
+      test "the first test case of first exercise is as expected":
+        let firstTestCase = probSpecsExercises[0].testCases[0].json
+        let firstTestCaseExpected = """{
+      "uuid": "64d97c14-36dd-44a8-9621-2cecebd6ed23",
+      "description": "accumulate empty",
+      "property": "accumulate",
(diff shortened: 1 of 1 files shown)


ee7 (approved):
Looks great. Thanks for the PR.

I probably would've just replaced the `acronym` case with the `accumulate` one, but this is fine by me too. There's probably some case where checking more than one exercise is helpful.

As an aside, regarding:

> - `nimble test` succeeds
> - `nim c --styleCheck:error -r ./tests/all_tests.nim` succeeds

I recommend the latter for running the tests. As an optimization, it compiles all the tests at once; `nimble test` is slower because each dependency like `system.nim` must be compiled multiple times (Nim might get better at this in Nim 1.6.0 - "incremental compilation" is being worked on).

---

Nitpick: Since we've had a good streak of uniform commit messages lately, I'll state a preference for changing the commit message at merge-time. I'd suggest a first line of something like
```
tests(probspecs): fix for `accumulate` as new first exercise (#185)
```

with a body of something like
```
Alphabetically, the first exercise with canonical data in
`exercism/problem-specifications` is now `accumulate`.

See:
- https://github.com/exercism/problem-specifications/pull/1738

```

ErikSchierboom (approved):

PR #195, opened 2021-02-23
lint: add runtime/wasmtime tag

ErikSchierboom:


SleeplessByte (approved):

ee7 (approved):
I don't have a strong opinion on the tag itself, but I do think it's useful for a student to filter for tracks that have good support for doing WebAssembly things. And this just keeps `configlet` in sync with the docs/website - approved.

---

As an aside: I guess we won't change the list of valid tags much after a certain point, but it might still be worth considering if we want a single source of truth for them across Exercism (rather than needing to keep them in-sync across repos).

I don't see an obvious best implementation, though. Some options:
1. Make some file in this repo the source of truth, and read/process it at compile-time. The other repos would use/refer to this file.
1. Make some file in the `exercism/docs` repo the source of truth, and add it in this repo as a git submodule/subtree/subrepo. Again, read/process it at compile-time.
1. Download a file from e.g. `exercism/docs` at compile-time (I'd vote against this: I don't want to make an internet connection a hard requirement for building `configlet`).
1. Make `configlet` download the list of valid tags at run-time, perhaps only when running in CI or after an opt-in. I wouldn't want to introduce an internet connection as a hard requirement just for linting tags, but if we do that for checking UUIDs Exercism-wide then this might make the most sense. 

But I'm fine with the current approach, at least for now.

ErikSchierboom:
> As an aside: I guess we won't change the list of valid tags much after a certain point, but it might still be worth considering if we want a single source of truth for them across Exercism (rather than needing to keep them in-sync across repos).

Yeah, I briefly thought about this too. We might even make it as simple as there running a GitHub actions workflow to detect changes to the list (presumably listed in a file in this repo) and then creating an issue to update in the other repo(s).

PR #203, opened 2021-02-27
git: ignore temporary repos used in binary tests

ErikSchierboom:
The binary tests temporarily clone the prob-specs repo and the Nim track repo. While these directories should be removed after running the tests, adding them to the .gitignore list will prevent them from being committed accidentally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,7 @@
 configlet
 configlet.exe
 
-# Temporary prob-specs repo
+# Temporary repos
 .problem-specifications/
+.test_binary_problem_specifications/
+.test_binary_nim_track_repo/


valentin-p (approved):

ee7 (approved):
> While these directories should be removed after running the tests

Just to clarify for others: with the current implementation, these dirs are removed only at the start of running the tests. See:

https://github.com/exercism/configlet/blob/f162e6161a1b9d974b6c514f33a2df6e53c98ac4/tests/test_binary.nim#L35-L40

I didn't make it also remove the directories at the end of the tests because:
- I might want to look at them if the tests fail
- It could be distracting to refactor to support "only remove directories at the end if all relevent tests passed"
- Several test files will require a prob-specs clone, and we could add code so that we only run `git clone` once for each repo when running every test.

I think this PR is an improvement regardless, but whenever convenient we might consider to:
- remove the temporary directories after the `test_binary` sync tests - even unconditionally is probably fine. 
- use a standard prefix for temporary directories, allowing us to use a pattern in `.gitignore` that covers temporary directories in other test files.

The other temporary directory in the current tests is here (it's removed after the test finishes):
https://github.com/exercism/configlet/blob/f162e6161a1b9d974b6c514f33a2df6e53c98ac4/tests/test_probspecs.nim#L10-L12

ee7:
I still have a strong preference for wrapping the commit message body at 72 chars, which is what we've done so far in this repo.

So if you don't hate it, could you please copy-paste the below at merge time? No diff apart from newlines.

```
The binary tests temporarily clone the prob-specs repo and the Nim track
repo. While these directories should be removed after running the tests,
adding them to the .gitignore list will prevent them from being 
committed accidentally.

```

ErikSchierboom:
> use a standard prefix for temporary directories, allowing us to use a pattern in .gitignore that covers temporary directories in other test files.

This would be a good to do at some point in the future.

PR #208, opened 2021-03-03
console: write more things to stderr

ErikSchierboom:
Rationale:
- When the user writes `--help`, the help message should go to stdout.
- When the user writes some incorrect options/arguments, the help
  message should go to stderr.

### DIFF
--- a/tests/test_binary.nim
+++ b/tests/test_binary.nim
@@ -271,13 +271,12 @@ proc main =
   const helpStart = &"Usage:\n  {binaryName} [global-options] <command> [command-options]"
 
   const cmd = "nimble --verbose build -d:release"
-  stdout.write(&"Running `{cmd}`... ")
-  stdout.flushFile()
+  stderr.write(&"Running `{cmd}`... ")
   let (buildOutput, buildExitCode) = execCmdEx(cmd, workingDir = repoRootDir)
   if buildExitCode == 0:
-    echo "success"
+    stderr.writeLine "success"
   else:
-    echo "failure"
+    stderr.writeLine "failure"
     raise newException(OSError, buildOutput)
 
   suite "help as an argument":
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -173,9 +173,12 @@ proc showHelp(exitCode: range[0..255] = 0) =
   const helpText = genHelpText()
   let appName = extractFilename(getAppFilename())
   let usage = "Usage:\n" &
-              &"  {appName} [global-options] <command> [command-options]\n\n"
-  stdout.write usage
-  echo helpText
+              &"  {appName} [global-options] <command> [command-options]\n"
+  let f = if exitCode == 0: stdout else: stderr
+  f.writeLine usage
+  f.writeLine helpText
+  if f == stdout:
+    f.flushFile()
   quit(exitCode)
 
 proc showVersion =


tests/test_binary.nim
```diff
@@ -271,13 +271,12 @@ proc main =
   const helpStart = &"Usage:\n  {binaryName} [global-options] <command> [command-options]"
 
   const cmd = "nimble --verbose build -d:release"
-  stdout.write(&"Running `{cmd}`... ")
-  stdout.flushFile()
+  stderr.write(&"Running `{cmd}`... ")
```
ErikSchierboom, line 274: Why is this writing to `stderr`?

tests/test_binary.nim
```diff
@@ -271,13 +271,12 @@ proc main =
...
   let (buildOutput, buildExitCode) = execCmdEx(cmd, workingDir = repoRootDir)
   if buildExitCode == 0:
-    echo "success"
+    stderr.writeLine "success"
```
ErikSchierboom, line 277: Shouldn't this be `stdout`?

```suggestion
    stdout.writeLine "success"
```

tests/test_binary.nim
ee7: The purpose of this line is to tell the user that something is happening (otherwise it appears like a hang), and I think it's standard for such messaging to go to `stderr`.

For example, if we make a request with `curl` and redirect stdout
```
curl https://releases.ubuntu.com/20.04.2.0/ubuntu-20.04.2.0-desktop-amd64.iso > /dev/null
```
we see that progress is printed to `stderr`.

Using `stderr` in the Nim code also avoids a `flushFile`, since `stderr` is unbufferred by default.

But it's somewhat arbitrary - everything else from running the tests is written to `stdout`. That is, with the latest commit on `main`, if you compile `tests/all_tests.nim` and then run
```shell
$ ./all_tests > /dev/null
``` 
nothing is printed to the screen even if the tests fail. That's because Nim's `unittest` module takes the position that all test results are "output" and so should appear on `stdout`.

By changing to `stderr` here, the user who writes
```shell
$ ./all_tests > /tmp/configlet_tests.txt
```
will no longer see the diagnostic message in that file, only the test results. I think that's a little better.

Can we just do `stderr` here for now?

tests/test_binary.nim
ee7: It's the continuation of the `"Running `{cmd}`... ` message, so it depends entirely on that decision.

tests/test_binary.nim
```diff
@@ -20,7 +20,7 @@ template execAndCheck(expectedExitCode: int; body: untyped) {.dirty.} =
   ## `body` must have the same return type as `execCmdEx`.
   let (outp, exitCode) = body
   if exitCode != expectedExitCode:
-    echo outp
+    stderr.writeLine outp
```
ee7, line 23: Actually I think I'll revert this one, given that all the other `unittest` output goes to stdout.

tests/test_binary.nim
ErikSchierboom: Yeah that makes sense. I was mostly just curious.

tests/test_binary.nim
ErikSchierboom: Right.

PR #212, opened 2021-03-11
tests: make `nimble test` faster

ee7:
By default, `nimble test` does this:
- find a Nim file beginning with `t` in the `tests` directory
- compile it
- run it
- repeat until all such files have been run.

But it's better to run the tests via `all_tests.nim`, which compiles
every test in one pass as an optimization. Individually compiling and
running each test file is slower because each dependency (most notably,
`system.nim`) must be compiled multiple times.

This optimization should become less important in the future - the
so-called "incremental compilation" is currently being worked on.

This commit overrides `nimble test` so that it runs the test suite the
faster way. We already ran `all_tests.nim` during CI.

Note that this commit doesn't make CI run `nimble test` for now - in CI
we currently pass `--styleCheck:error` instead of `--styleCheck:hint`.

### DIFF
--- a/configlet.nimble
+++ b/configlet.nimble
@@ -11,3 +11,6 @@ requires "nim >= 1.4.2"
 requires "parsetoml"
 requires "cligen"
 requires "uuids >= 0.1.11"
+
+task test, "Runs the test suite":
+  exec "nim r ./tests/all_tests.nim"


ErikSchierboom (approved):
Any idea how much this saves us?

ee7:
> Any idea how much this saves us?

Well, it only saves time for somebody who was writing `nimble test` - I usually type `nim r tests/a[Tab]` or similar instead, and CI was already running `all_tests.nim`.

But it does:
- Make the repo more accessible for newcomers.
- Make the output of `nimble test` a bit cleaner
- Allow me to use my `alias nt='nimble test'` shell alias...

In general it'll typically save about a second or two per test file. On a not-fast machine:

| command | no cache | second run |
| :- | -: | -: |
| `nim c -r all_tests.nim` | 30 s | 15 s |
| `nimble test` | 31 s | 20 s |

The remaining runtime comes almost entirely from:
- compiling configlet in release mode in `test_binary.nim`
- fetching repos

rather than any of the test code itself.

ErikSchierboom:
Excellent!

PR #213, opened 2021-03-11
refactor: prefer `func`

ee7:
This PR changes `proc` to `func` where possible.

I made separate commits here, but I'd prefer to squash when merging.

### DIFF
--- a/src/lint/validators.nim
+++ b/src/lint/validators.nim
@@ -1,7 +1,7 @@
 import std/[json, strutils]
 import ".."/helpers
 
-proc q(s: string): string =
+func q(s: string): string =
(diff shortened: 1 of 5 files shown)


ee7:
Force-pushed to fix CI failure on Windows, see: https://github.com/exercism/configlet/runs/2085280331

It turns out `fileExists` apparently has side effects on Windows, but not on Linux.

See [the implementation of os.fileExists](https://github.com/nim-lang/Nim/blob/ddc5f8fbcfd55aa74a2fc7b13b287262780ba58e/lib/pure/os.nim#L1137-L1155).

ErikSchierboom (approved):

PR #214, opened 2021-03-11
lint(validators): add missing context

ee7:
For a `.meta/config.json` containing a whitespace-only string:
```json
{
  "files": {
    "solution": [" "],
    "test": ["test_foo.nim"],
    "exemplar": [".meta/exemplar.nim"]
  }
}
```

Before this commit:
```
    Array contains whitespace-only string: 'solution'
```

With this commit:
```
    Array contains whitespace-only string: 'files.solution'
```


ee7:
(Ignore the force-push - I pushed to the wrong branch).

ErikSchierboom (approved):
Great catch

PR #215, opened 2021-03-11
lint(validators): refactor `len` conditions

ee7:
In our validation procs, there are conditions like
```Nim
if data.hasKey(key):
  if data[key].kind == JString:
```
where the condition being `true` represents the "happy path".

To improve readability, let's try to do this consistently for every
condition apart from the innermost one; it's easier to see the "happy
path" when it minimises the number of `else`.

This commit refactors the `len` conditions accordingly, so that we
prefer `len > 0` and use `len == 0` only when there is no `else`.

This commit also changes a `var` to a `let` to better convey that we
don't mutate.

### DIFF
--- a/src/lint/validators.nim
+++ b/src/lint/validators.nim
@@ -22,10 +22,11 @@ proc checkString*(data: JsonNode; key, path: string; isRequired = true): bool =
   if data.hasKey(key):
     if data[key].kind == JString:
       let s = data[key].getStr()
-      if s.len == 0:
+      if s.len > 0:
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
LGTM

ee7:
I've again verified that this commit keeps the output of `configlet lint` the same on every track. I intend to add that to CI eventually, although it can't be a "test" exactly - some commits will be expected to change the output of `configlet lint`.

PR #218, opened 2021-03-11
lint: support top-level array

ee7:
So far, none of the files we were linting were an array at the top level. But `links.json` is.

Some possible future refactorings:
- Use common logic for all the "missing key" conditions.
- Make the `arrayOfStrings` procs extend the `array` procs.
- Make the empty string unnecessary in `isObject(data, "", path)`.

This PR is still a bit WIP, and I'm not set on its approach yet. But it might soon be good enough for now.

I need to do the below before I merge this PR:
- [x] Fix the bug that makes `configlet lint` produce an error
- [x] Verify that this PR does not change the output of `configlet lint`

### DIFF
--- a/src/lint/concept_exercises.nim
+++ b/src/lint/concept_exercises.nim
@@ -23,12 +23,12 @@ proc checkFiles(data: JsonNode, context, path: string): bool =
     result = false
 
 proc isValidConceptExerciseConfig(data: JsonNode, path: string): bool =
-  if isObject(data, "root", path):
+  if isObject(data, "", path):
     result = true
-    if not checkArrayOf(data, "authors", path, isValidAuthorOrContributor):
+    if not hasArrayOf(data, "authors", path, isValidAuthorOrContributor):
       result = false
-    if not checkArrayOf(data, "contributors", path, isValidAuthorOrContributor,
-                        isRequired = false):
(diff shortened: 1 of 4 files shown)


src/lint/concept_exercises.nim
```diff
@@ -13,26 +13,26 @@ proc isValidAuthorOrContributor(data: JsonNode, context: string, path: string):
 proc checkFiles(data: JsonNode, context, path: string): bool =
   result = true
   if hasObject(data, context, path):
-    if not checkArrayOfStrings(data, context, "solution", path):
+    if not hasArrayOfStrings(data, context, "solution", path):
```
ErikSchierboom, line 16: I like the new name!

src/lint/validators.nim
```diff
@@ -38,47 +41,66 @@ func format(context, key: string): string =
   else:
     q(key)
 
-proc checkArrayOfStrings*(data: JsonNode; context, key, path: string;
-                          isRequired = true): bool =
+proc isArrayOfStrings*(data: JsonNode;
```
ErikSchierboom, line 44: Love the new name!

ErikSchierboom (approved):
Nice!

> Use common logic for all the "missing key" conditions.
> Make the arrayOfStrings procs extend the array procs.
> Make the empty string unnecessary in isObject(data, "", path).

Those are all nice refactorings for the future.

PR #199, opened 2021-02-26
lint: allow null for optional values

ErikSchierboom:
In JSON, `null` is used to denote a value as being absent. This PR allows tracks to use `null` as the value for optional (non-required) fields.

Closes https://github.com/exercism/configlet/issues/192

PR #224, opened 2021-03-19
lint: allow null for optional values (review)

ee7:
I think the case statement with
```Nim
of JNull
```
makes things more readable than
```Nim
elif data[key].kind != JNull or isRequired:
```

And maybe it's slightly better to have the ability to write targetted error messages for the `null` case. 

I've verified that this PR keeps the output of `configlet lint` the same as #199 on every track.

### DIFF
--- a/src/lint/validators.nim
+++ b/src/lint/validators.nim
@@ -144,7 +144,7 @@ proc checkBoolean*(data: JsonNode; key, path: string; isRequired = true): bool =
   if data.hasKey(key):
     case data[key].kind
     of JBool:
-      return
+      return true
     of JNull:
       if isRequired:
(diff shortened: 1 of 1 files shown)


src/lint/validators.nim
```diff
@@ -23,14 +23,19 @@ proc hasObject*(data: JsonNode; key, path: string; isRequired = true): bool =
 proc checkString*(data: JsonNode; key, path: string; isRequired = true): bool =
   result = true
   if data.hasKey(key):
-    if data[key].kind == JString:
+    case data[key].kind
+    of JString:
```
ErikSchierboom, line 27: I like this case/of statement better!

ErikSchierboom (approved):
LGTM

PR #226, opened 2021-03-19
lint: refactor conditions

ee7:
Maybe something like this.

I might change this to something else, but the `b1 and b2 and ...` thing was getting a bit silly.


ErikSchierboom (approved):
Nice. I was also getting slightly annoyed by the b1, b2.. :)

PR #227, opened 2021-03-19
lint: print link to the spec on failure

ee7:
Something like this could help the user fix the given linting errors.


ErikSchierboom (approved):
Good idea. In the future, we can link to specific sections in the documentation for individual errors, but this is already very helpful. I do need to update that document at some point 😱

PR #228, opened 2021-03-19
lint: refactor to use common json parsing helper

ee7:
Something like this.

### DIFF
--- a/src/lint/validators.nim
+++ b/src/lint/validators.nim
@@ -1,4 +1,4 @@
-import std/[json, os, strutils]
+import std/[json, os, streams, strutils]
 import ".."/helpers
 
 func q(s: string): string =
@@ -182,3 +182,24 @@ proc subdirsContain*(dir: string, files: openArray[string]): bool =
         let path = subdir / file
         if not fileExists(path):
           result.setFalseAndPrint("Missing file", path)
+
+proc parseJsonFile*(path: string, b: var bool, allowEmptyArray = false): JsonNode =
+  ## Parses the file at `path` and returns the resultant JsonNode.
+  ##
+  ## Sets `b` to false if unsuccessful.
+  if fileExists(path):
+    let contents = readFile(path)
+    if not isEmptyOrWhitespace(contents):
+      try:
+        result = parseJson(newStringStream(contents), path)
+      except JsonParsingError:
+        b.setFalseAndPrint("JSON parsing error", getCurrentExceptionMsg())
+    else:
+      let msg =
+        if allowEmptyArray:
+          ", but must contain at least the empty array, `[]`"
+        else:
+          ""
+      b.setFalseAndPrint("File is empty" & msg, path)
+  else:
+    b.setFalseAndPrint("Missing file", path)
--- a/src/lint/concept_exercises.nim
+++ b/src/lint/concept_exercises.nim
@@ -48,16 +48,15 @@ proc isEveryConceptExerciseConfigValid*(trackDir: string): bool =
         if not isValidConceptExerciseConfig(j, configPath):
           result = false
 
-proc conceptExerciseFilesExist*(trackDir: string): bool =
+proc conceptExerciseDocsExist*(trackDir: string): bool =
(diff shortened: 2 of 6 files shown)


src/lint/validators.nim
```diff
@@ -182,3 +182,20 @@ proc subdirsContain*(dir: string, files: openArray[string]): bool =
...
+  if fileExists(path):
+    let contents = readFile(path)
+    if not checkEmpty or not isEmptyOrWhitespace(contents):
+      try:
+        result = parseJson(newStringStream(contents), path)
+      except JsonParsingError:
+        b.setFalseAndPrint("JSON parsing error", getCurrentExceptionMsg())
+    else:
+      b.setFalseAndPrint("File is empty, but must contain at least the empty " &
+                         "array, `[]`", path)
```
ErikSchierboom, line 199: The array suffix not necessarily the right message, as some JSON files should be empty objects then. Maybe just make it a bit more general about the file not being allowed to be empty?

ErikSchierboom (approved):
LGTM Just some tiny nits

ee7:
Previously (with 8d7fd89 or ff5349b) this PR would cause the following incorrect diff to the `configlet lint` output:

#### clojure
```diff
+Missing file:
+./exercises/concept/interest-is-interesting/.meta/config.json
+
+Missing file:
+./exercises/concept/international-calling-connoisseur/.meta/config.json
+
+Missing file:
+./exercises/concept/squeaky-clean/.meta/config.json
+
```

#### dart
```diff
+Missing file:
+./exercises/concept/futures/.meta/config.json
+
+Missing file:
+./exercises/concept/numbers/.meta/config.json
+
+Missing file:
+./exercises/concept/strings/.meta/config.json
+
```

That is, we would warn a second time about missing files. It seems like a good idea to keep the missing file check in `parseJsonFile`, and not suppress it with some parameter. So let's check for the existence of JSON files only in the procs that validate them, which also allows us to simplify the linting success message.

With https://github.com/exercism/configlet/pull/228/commits/191a94d87c524a48677fbb92889da2007e741703, this PR now produces the below diff:
#### clojure
Some errors are moved until later:
```diff
-Missing file:
-./exercises/concept/interest-is-interesting/.meta/config.json
-
-Missing file:
-./exercises/concept/international-calling-connoisseur/.meta/config.json
-
-Missing file:
-./exercises/concept/squeaky-clean/.meta/config.json
-
+Missing file:
+./exercises/concept/interest-is-interesting/.meta/config.json
+
+Missing file:
+./exercises/concept/international-calling-connoisseur/.meta/config.json
+
+Missing file:
+./exercises/concept/squeaky-clean/.meta/config.json
+
```



#### swift
We fix a bug on `main`, where we warned about missing `links.json` files twice. This currently only affects `swift`.
```diff
-Missing file:
-./concepts/characters/links.json
-
-Missing file:
-./concepts/classes/links.json
-
-Missing file:
-./concepts/initializers/links.json
-
-Missing file:
-./concepts/opaque-indices/links.json
-
-Missing file:
-./concepts/strings/links.json
-
-Missing file:
-./concepts/structs/links.json
-
```

We were previously printing errors for missing `links.json` in both `isEveryConceptLinksFileValid` and `conceptFilesExist`:
https://github.com/exercism/configlet/blob/f24176d0cdc737e4686c526a81de0b26d11c0cb2/src/lint/concepts.nim#L65-L79

src/lint/validators.nim
ee7: Yeah, I know - but we don't have any JSON files where `{}` is valid.

I've tried to improve it in the [force-push diff](https://github.com/exercism/configlet/compare/8d7fd895f17ea16ece7a5d97743b6fd2e6dd4671..ff5349b828a5dbf32cbe2493c5ea10d85d3706e5) (a little noisy, sorry) - it allows us to print a special message for empty `links.json` files, while still printing a "file is empty" message for empty `.meta/config.json` files. Good enough for now?

ErikSchierboom (approved):
I'm fine with this for now. I still find the "allow empty array" slightly odd, but we can fix that later.

PR #244, opened 2021-03-26
lint: refactor to check value in `checkInteger`

ee7:
This simplifies the code at the call site and reduces duplication.

---

The spec contains
> The "online_editor.indent_size" value must be a positive integer (>= 0)

But do we want to limit the `indent_size` to, say, `0..8`?

This PR doesn't affect the `configlet lint` output. But the new messages would look like
```
The value of `version` is `2`, but it must be `3`:
./config.json

The value of `indent_size` is `9`, but it must be between 0 and 8 (inclusive):
./config.json
```

### DIFF
--- a/src/lint/validators.nim
+++ b/src/lint/validators.nim
@@ -172,12 +172,22 @@ proc checkBoolean*(data: JsonNode; key, path: string; isRequired = true): bool =
   elif isRequired:
     result.setFalseAndPrint("Missing key: " & q(key), path)
 
-proc checkInteger*(data: JsonNode; key, path: string; isRequired = true): bool =
+proc checkInteger*(data: JsonNode; key, path: string; isRequired = true;
+                   allowed: Slice): bool =
   result = true
   if data.hasKey(key):
     case data[key].kind
     of JInt:
-      return true
+      let num = data[key].getInt()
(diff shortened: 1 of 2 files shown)


ErikSchierboom (approved):
I like it.

src/lint/validators.nim
```diff
@@ -172,12 +172,22 @@ proc checkBoolean*(data: JsonNode; key, path: string; isRequired = true): bool =
   elif isRequired:
     result.setFalseAndPrint("Missing key: " & q(key), path)
 
-proc checkInteger*(data: JsonNode; key, path: string; isRequired = true): bool =
+proc checkInteger*(data: JsonNode; key, path: string; isRequired = true;
+                   allowed: Slice): bool =
```
ee7, line 176: I've kept `allowed` as a required argument, for now. All of the integers in the spec so far have a restricted value:

> The "version" value must be the integer 3
> The "online_editor.indent_size" value must be a positive integer (>= 0)
> The "exercises.practice[].difficulty" value must be an integer >= 0 and <= 10

ErikSchierboom: I'm fine with that. We can update later if needed.

PR #245, opened 2021-03-26
lint: move set membership check into `checkString`

ee7:
Similar to https://github.com/exercism/configlet/pull/244

Another refactoring is required before we can deduplicate the logic for the tags too.

---

This commit doesn't produce a diff to the current `configlet lint` output, but if it did it'd look like
```
The value of `indent_style` is `foo`, but it must be one of {"space", "tab"}:
./config.json
```

### DIFF
--- a/src/lint/validators.nim
+++ b/src/lint/validators.nim
@@ -1,4 +1,4 @@
-import std/[json, os, streams, strutils]
+import std/[json, os, sets, streams, strutils]
 import std/unicode except strip
 import ".."/helpers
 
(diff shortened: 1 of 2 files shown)


src/lint/validators.nim
```diff
@@ -34,14 +34,27 @@ proc hasValidRuneLength(s, key, path: string; maxLen: int): bool =
...
   result = true
   if data.hasKey(key):
     case data[key].kind
     of JString:
       let s = data[key].getStr()
-      if s.len > 0:
+      if allowed.len > 0:
+        if s notin allowed:
+          let msgStart = "The value of `" & key & "` is `" & s &
+                          "`, but it must be one of "
```
ee7, line 50: Currently this is losing some context - we'll need to refactor later to regain an error message like
> "The value of `online_editor.indent_style` ..."

In general the `context` handling should be tidied up everywhere.

ErikSchierboom: > In general the context handling should be tidied up everywhere.

Agreed. I'm fine with doing that later.

ErikSchierboom (approved):

PR #252, opened 2021-03-30
lint(validators): add `isString`, `isBoolean`, `isInteger`

ee7:
This commit also deduplicates some logic between `hasObject` and `isObject`.

After this, I'll:
- make the tag checking use `isString`
- deduplicate the "missing key" logic
- make `isArrayOfStrings` use `isString` internally


ErikSchierboom (approved):
I like it!

PR #254, opened 2021-03-31
lint(validators): use `strformat`

ee7:
This improves readability.

Note that we must add the export marker to the `q` func.

---

Later I'll make our error messages more consistent, using either backticks or single quotes (rather than both).


ErikSchierboom (approved):
This improves readability quite a bit I feel!

PR #255, opened 2021-03-31
lint(validators): use shared logic for missing keys

ee7:
The first commit here (e12a46141f8fa832ba280b5ae0a22cf6086d9937) is a proc-only implementation, but I don't love all the

```Nim
  elif not isRequired:
    result = true
```

and I didn't see a cleaner way to do it with e.g. a proc that takes a `var bool` parameter.

The second commit (f1bbea237f1943e4b7a95c1991220989692d3f31) is an alternative that uses a template instead.

If we make the template take a `context` parameter explicitly, I believe we can't set its default value to `""` because we'd get a compilation error like
```
template handleMissingKey(data: JsonNode; key, path: string; isRequired: bool;
                          context = ""; body: untyped)
  first type mismatch at position: 5
  required type for context: string
  but expression '
result = isObject(data[key], key, path)' is of type: typed
```

But maybe we could clean up that ugliness later with all the other `context` ugliness.

@ErikSchierboom which commit do you prefer? Can you see a better way to refactor this?

I think each implementation is an improvement over the pre-PR state, but I don't love either of them - I'm interested in what you find clearer. I'm leaning towards just sticking with the first commit for now.

### DIFF
--- a/src/lint/validators.nim
+++ b/src/lint/validators.nim
@@ -20,14 +20,17 @@ func format(context, key: string): string =
   else:
     q(key)
 
-proc hasKey(data: JsonNode; key, path: string; isRequired: bool;
-            context = ""): bool =
-  ## Returns true if `data` contains the key `key`. Otherwise, returns false
-  ## and, if `isRequired` is true, prints an error message.
+template handleMissingKey(data: JsonNode; key, path, context: string;
+                          isRequired: bool; body: untyped) =
+  ## Runs `body` if `data` contains the key `key`. Otherwise: prints an error
+  ## message if `isRequired` is true, and sets `result` to true if `isRequired`
+  ## is false.
   if data.hasKey(key):
-    result = true
+    body
   elif isRequired:
-    result.setFalseAndPrint(&"Missing key: {format(context, key)}", path)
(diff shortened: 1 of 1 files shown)


src/lint/validators.nim
```diff
@@ -20,14 +20,17 @@ func format(context, key: string): string =
...
+                          isRequired: bool; body: untyped) =
+  ## Runs `body` if `data` contains the key `key`. Otherwise: prints an error
+  ## message if `isRequired` is true, and sets `result` to true if `isRequired`
+  ## is false.
   if data.hasKey(key):
-    result = true
+    body
   elif isRequired:
-    result.setFalseAndPrint(&"Missing key: {format(context, key)}", path)
+    result.setFalseAndPrint("Missing key: " & format(context, key), path)
```
ee7: This change is required because we can't use strformat like that inside a template (due to strformat implementation details).

ErikSchierboom:
> I'm leaning towards just sticking with the first commit for now.

That would be my preference too. I think its actually not that bad, as the code is quite explicit and thus easy to read.

ee7:
OK - reverted.

ErikSchierboom (approved):

PR #274, opened 2021-04-09
lint: add kebab-case checks

ee7:
This commit implements these checks for a track `config.json`:
> - The `"slug"` value must be a non-empty, non-blank, lowercased string using kebab-case
> - The `"exercises.concept[].slug"` value must be a non-empty, non-blank, lowercased string using kebab-case
> - The `"exercises.concept[].concepts"` values must be non-empty, non-blank, lowercased strings using kebab-case
> - The `"exercises.concept[].prerequisites"` values must be non-empty, non-blank, lowercased strings using kebab-case
> - The `"exercises.practice[].slug"` value must be a non-empty, non-blank, lowercased string using kebab-case
> - The `"exercises.practice[].practices"` values must be non-empty, non-blank, lowercased strings using kebab-case
> - The `"exercises.practice[].prerequisites"` values must be non-empty, non-blank, lowercased strings using kebab-case
> - The `"exercises.foregone"` values must be non-empty, non-blank, lowercased strings using kebab-case
> - The `"concepts[].slug"` value must be a non-empty, non-blank, lowercased string using kebab-case

--- 

This PR currently causes the following diff to the output of `configlet lint`:

#### rust
```diff
+The `exercises.concept.prerequisites` array contains `&str`, but every value must be lowercase and kebab-case:
+./config.json
+
+The `concepts.slug` value is `&str`, but it must be a lowercase and kebab-case string:
+./config.json
+
+Configlet detected at least one problem.
+For more information on resolving the problems, please see the documentation:
+https://github.com/exercism/docs/blob/main/building/configlet/lint.md
```

What should the definition of "kebab string" be?
- ~A string that only contains characters in `{'a'..'z', '0'..'9', '-'}` (the definition in e8983241bb72757cc4a0d0f5aff96dfc47f69fa0)~
- A string that matches `^([a-z0-9]+)(-[a-z0-9]+)*$`
- ~A string that does not contain a character in `{'A'..'Z', '_', ' '}`~
- ~Something else~

I think we can either
1. Disallow special characters for all kebab strings, and tell the Rust track to write the concept name as something like "string-slices" instead.
1. ~Allow special characters in a concept name, so something like `&str` is allowed, but disallow them in the other kebab strings.~

See e.g.
- https://doc.rust-lang.org/std/str/index.html
- https://doc.rust-lang.org/book/ch04-03-slices.html?highlight=%26str#string-slices

### DIFF
--- a/src/lint/validators.nim
+++ b/src/lint/validators.nim
@@ -83,21 +83,18 @@ func isKebabCase*(s: string): bool =
   ## - Does not contain two adjacent `-` characters
   ## This corresponds to the regex pattern `^[a-z0-9]+(-[a-z0-9]+)*$`.
   const lowerAndDigits = {'a'..'z', '0'..'9'}
-  let L = s.len
-
-  if L > 0 and s[0] in lowerAndDigits and s[^1] in lowerAndDigits:
-    var i = 0
-    while true:
-      i += s.skipWhile(lowerAndDigits, start = i)
-      if i == L:
-        return true
-      elif s[i] == '-':
-        if s[i-1] == '-':
-          return false
-        else:
-          inc i
-      else:
-        return false
+  let sLen = s.len
+  var i = 0
+  while i < sLen:
+    if s[i] == '-':
+      return false
+    i += s.skipWhile(lowerAndDigits, start = i)
+    if i == sLen:
+      return true
+    elif s[i] == '-':
+      inc i
+    else:
+      return false
 
 proc isString*(data: JsonNode; key: string; path: Path; context: string;
                isRequired = true; allowed = emptySetOfStrings;
--- a/src/lint/validators.nim
+++ b/src/lint/validators.nim
@@ -211,7 +211,8 @@ proc hasArrayOfStrings*(data: JsonNode;
   if data.hasKey(key, path, context, isRequired):
     let contextAndKey = joinWithDot(context, key)
     result = isArrayOfStrings(data[key], contextAndKey, path, isRequired,
-                              allowed, allowedArrayLen, checkIsKebab = checkIsKebab)
+                              allowed, allowedArrayLen,
+                              checkIsKebab = checkIsKebab)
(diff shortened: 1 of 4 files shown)


src/lint/validators.nim
```diff
@@ -75,9 +75,16 @@ proc isUrlLike(s: string): bool =
 const
   emptySetOfStrings = initHashSet[string](0)
 
+func isLowerKebab(s: string): bool =
+  ## Returns true if `s` is a lowercase and kebab-case string.
+  result = true
+  for c in s:
+    if c notin {'a'..'z', '0'..'9', '-'}:
```
ee7, line 82:
```suggestion
    if c in {'A'..'Z', '_', ' '}:
```
Do we want something like this instead? This allows special characters like `&`.

ErikSchierboom:
Option 1 is what we want. I'll update the docs to clarify.

src/lint/validators.nim
ErikSchierboom: No, the original code is what we want.

src/lint/validators.nim
```diff
@@ -101,6 +108,16 @@ proc isString*(data: JsonNode; key: string; path: Path; context: string;
         result.setFalseAndPrint(&"Not a valid URL: {q s}", path)
     elif s.len > 0:
       if not isEmptyOrWhitespace(s):
+        if checkIsKebab:
+          if not isLowerKebab(s):
```
ErikSchierboom, line 112: The parameter is named `checkIsKebab` but the implementation is named `isLowerKebab`. I don't mind this much, but it is a tiny bit inconsistent.

ErikSchierboom (approved):

src/lint/validators.nim
ee7: Renamed to `isKebabCase`.

ErikSchierboom (approved):
Love the extensive tests!

PR #276, opened 2021-04-09
lint(validators): fix array length check

ee7:
Previously, we didn't actually print the error (or set `result` to `false`) in `isArrayOfStrings` if the array was a non-allowed length. Oops.

Similarly to #275, this resolves these compiler hints:
> ```
> ./src/lint/validators.nim(165, 13) Hint: 'msgEnd' is declared but not used [XDeclaredButNotUsed]
> ./src/lint/validators.nim(163, 13) Hint: 'msgStart' is declared but not used [XDeclaredButNotUsed]
> ```

This PR does not affect the output of `configlet lint`. That's because we currently don't really use the `allowedArrayLen` feature of `isArrayOfStrings` / `hasArrayOfStrings`.


ErikSchierboom (approved):
Whoops! :D

PR #284, opened 2021-04-27
fetch-configlet: improve behavior when run from non-root

IsaacG:


glennj:
Perhaps [Is there a way to get the git root directory in one command?](https://stackoverflow.com/questions/957928/is-there-a-way-to-get-the-git-root-directory-in-one-command)

IsaacG:
> Perhaps [Is there a way to get the git root directory in one command?](https://stackoverflow.com/questions/957928/is-there-a-way-to-get-the-git-root-directory-in-one-command) -- then the script won't get fooled by the presence of a "bin" _subdirectory_

That wouldn't tell us if that git repo/PWD we're in has a `./bin` or not. If we're fetching into a `./bin` then we should check that a `./bin` exists.

glennj:
I was trying to ensure that if we have
```none
root
├── bin
└── foo
    └── bin
```
that we're in "root" not "foo"

IsaacG:
> I was trying to ensure that if we have
> 
> ```
> root
> ├── bin
> └── foo
>     └── bin
> ```
> 
> that we're in "root" not "foo"

Done

ee7 (approved):
Thanks for the PR - I think it's a strict improvement on the current situation, and so we can merge in the current state.

But while we're here, and in case you wanted to work on it more: can we just remove the restriction that the script must be run from the repo root? It was always a bit silly that somebody can't run it while inside the `bin` directory (or, say, `exercises/practice/hello-world`).

Some options below. The script could:
1. alter the `output_dir` (and thus `output_path`) so that we download to the `bin` directory in the repo root
1. change directory to the repo root
1. alter `output_dir` so that we download to the same directory as that of the script
1. change directory to that of the script

I think one of the last two is probably best, since that allows someone to move the script elsewhere on their machine if they wanted to. This would side-step the "the script should create the `bin` directory if it doesn't exist" issue.

But I'm sure you know better than me what is most idiomatic/tasteful here.

IsaacG:
It's easy enough to use `"$(git rev-parse --show-toplevel)/bin"` as the output path so that the script works regardless of the CWD so long as it is somewhere in the repo. But this script is still relying heavily on the CWD to work so you're somewhat hampered in your ability to "move the script elsewhere on their machine" and run the script from any CWD.

If the goal is to fetch and run the code, you could use a `mktemp -d` and not care about the CWD. If you want the script to work from anywhere but update a specific path ... without relying on CWD, you'd need the user to pass in a target dir. But if you want to update some arbitrary `./bin` that is tied to the CWD, you're going to be depending on the user's CWD.

I could tweak it to work anywhere in the repo and target `"$(git rev-parse --show-toplevel)/bin"` I suppose.

scripts/fetch-configlet
```diff
@@ -45,14 +45,31 @@ get_download_url() {
...
-esac
+# Check if we are in the root of repo that has a $REPO_ROOT/bin
+# This is needed as `curl` will output to `$REPO_ROOT/bin/`
+in_repo () {
+  local root
+  root=$(git rev-parse --show-toplevel 2>/dev/null) || return
+  [[ "${PWD}" = "${root}"/* ]] && [[ -d "${root}/bin" ]]
+}
+
+main () {
```
ErikSchierboom, line 56: What was the reason to introduce a `main` function instead of the old approach?

ErikSchierboom:
I like the idea of also being able to run the script from outside the root directory, for example when being in the `bin` directory. Instead of the current git root detection code, maybe we could just download configlet to the same directory in which the `fetch-configlet` script is located? Logically, that seems to make the most sense to me. This approach to me is straightforward and makes intuitive sense. We also don't need any of the git root commands, which I think is a bonus.

scripts/fetch-configlet
ee7: @ErikSchierboom So that we can have the `return 1` line.

If we remove the `main` function then we get the error
```
return: can only `return' from a function or sourced script
```


But I think the script is long enough that `main` improves readability anyway. 

See also https://google.github.io/styleguide/shellguide.html#s7.8-main

ErikSchierboom: > So that we can have the return 1 line.

Right. `exit 1` would also have worked, right?

> But I think the script is long enough that main improves readability anyway.

🤷 I'm okay with it.

IsaacG: What ee7 said. The main function, like in many languages, doesn't alter behavior as much as serve to group a logical unit of code. The top level was getting crowded. A main function isolates the "main" logic.

IsaacG:
Fetching the configlet into the CWD certainly does simplify things a bunch but also alters the behavior of the script, and the workflow around it, pretty significantly. With the old behavior, you would be able to update the contents of bin/ from the repo root and running from inside bin would not work. If we fetch to CWD then you would only be able to update bin from inside bin.

This PR was aiming to address the issue where running the script from non-repo-root would result in cryptic errors. Originally it just added a helpful error, then it got a bit smarter about doing what it used to do in a more flexible manner. Completely alerting the workflow wasn't the intent here.

ErikSchierboom:
> Fetching the configlet into the CWD certainly does simplify things a bunch but also alters the behavior of the script, and the workflow around it, pretty significantly. With the old behavior, you would be able to update the contents of bin/ from the repo root and running from inside bin would not work. If we fetch to CWD then you would only be able to update bin from inside bin.

That was not exactly what I was proposing. I was proposing to download the file to the same directory where the `fetch-configlet` script is located. So running `./bin/fetch-configlet`, `cd bin && ./fetch-configlet` or `cd exercises/practice/hamming && ../../../bin/fetch-configlet` would all have the same result: configlet being downloaded to `./bin/configlet`. I don't think this influences any existing workflow.

ee7:
> This PR was aiming to address the issue where running the script from non-repo-root would result in cryptic errors.

@IsaacG Yeah - I'm sorry that this was a bit of a rabbit hole. If you prefer then we can just merge this PR and handle further changes later. 

It's just that the old behavior was bad anyway, so I'd like to fix the root cause.

But yeah, can we can simplify via something like this?
```shell
cd -- "$(dirname -- "$0")" # Change directory to that of this script.
```

(Though I don't know if that's the most idiomatic command).

IsaacG:
The fetch-configlet is in ./scripts and not in ./bin. If we fetch to the directory that it is in, we'd be altering behavior. 

It's generally frowned upon to rely on the path of $0 for behavior; $0 can be altered via symlinks or the exec built-in. As much as I'm not a huge fan of being reliant on CWD, that's still more idiomatic that relying on $0. When you run ls or curl or chrome, does it matter if they are in /bin or /use/bin or /opt/bin or $HOME/bin? The location of the fetcher likewise shouldn't be influencing the behavior.

See too https://mywiki.wooledge.org/BashFAQ/028

ErikSchierboom:
> The fetch-configlet is in ./scripts and not in ./bin. If we fetch to the directory that it is in, we'd be altering behavior.

It's only in scripts in _this_ repo, all the track repos have it in https://github.com/exercism/csharp/tree/main/bin

I find it hard to describe _why_, but I don't like using git to determine the root directory.

IsaacG:
If you don't want to use `git`, I think the other clean option is to not try to determine if PWD is a git repo or not and stop worrying about the location. Check that a ./bin exists and complain if not. Optionally, if CWD is `*/bin` then just use `./`

ErikSchierboom:
>  I think the other clean option is to not try to determine if PWD is a git repo or not and stop worrying about the location. Check that a ./bin exists and complain if not. Optionally, if CWD is */bin then just use ./

I would prefer this simpler option.

IsaacG:
Done :) This is closer to the first commit.

ErikSchierboom (approved):
Tested and working as expected! Very nice 👍

PR #293, opened 2021-04-29
lint(concept_exercises): validate `icon` property

ErikSchierboom:
Check the `icon` property of the `.meta/config.json` file of Concept Exercises.

See https://github.com/exercism/docs/pull/123

### DIFF
--- a/src/lint/concept_exercises.nim
+++ b/src/lint/concept_exercises.nim
@@ -22,6 +22,7 @@ proc isValidConceptExerciseConfig(data: JsonNode, path: Path): bool =
       hasValidFiles(data, path),
       hasArrayOfStrings(data, "forked_from", path, isRequired = false),
       hasString(data, "language_versions", path, isRequired = false),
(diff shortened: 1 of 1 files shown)


ee7 (approved):
1. Should this eventually check that the `icon` value is an allowed value?
2. If yes, should we do it as we do for `tags` (by defining a list of them in this repo) or as we'll do for `forked_from` (by making it an online check)?
3. Similarly, should `key_features[].icon` be an offline check, or an online check?

ErikSchierboom:
> Should this eventually check that the icon value is an allowed value?

Probably. I didn't worry about that right now as we don't have the full list of icons.

> If yes, should we do it as we do for tags (by defining a list of them in this repo) or as we'll do for forked_from (by making it an online check)?
> Similarly, should key_features[].icon be an offline check, or an online check?

The icons are quite static, so I think we can do it like with tags. One nice thing that I'd like to do is to setup a CI check on the website-icons repo to automatically create an issue (or PR!) in this repo whenever a new icon is added.

PR #294, opened 2021-04-29
lint(track_config): validate editor language properties

ErikSchierboom:
This PR implements the rules for two new fields in the `online_editor` key:

- The `"ace_editor_language"` key is required
- The `"ace_editor_language"` value must be a non-blank string
- The `"highlightjs_language"` key is required
- The `"highlightjs_language"` value must be a non-blank string

See https://github.com/exercism/docs/pull/124/files

### DIFF
--- a/src/lint/track_config.nim
+++ b/src/lint/track_config.nim
@@ -65,6 +65,8 @@ proc hasValidOnlineEditor(data: JsonNode; path: Path): bool =
     let checks = [
       hasString(d, "indent_style", path, k, allowed = indentStyles),
       hasInteger(d, "indent_size", path, k, allowed = 0..8),
+      hasString(d, "ace_editor_language", path, k),
+      hasString(d, "highlightjs_language", path, k),
     ]
     result = allTrue(checks)
 


src/lint/track_config.nim
```diff
@@ -65,6 +65,8 @@ proc hasValidOnlineEditor(data: JsonNode; path: Path): bool =
     let checks = [
       hasString(d, "indent_style", path, k, allowed = indentStyles),
       hasInteger(d, "indent_size", path, k, allowed = 0..8),
+      hasString(d, "ace_editor_language", path, k),
+      hasString(d, "highlightjs_language", path, k),
```
ee7, line 69:
```suggestion
      hasString(d, "highlightjs_language", path, k, checkIsKebab = true),
```
Do we want this?

ee7 (commented):
See questions in https://github.com/exercism/docs/pull/125

src/lint/track_config.nim
ErikSchierboom: I don't think so, as `c++` e.g. is an allowed value but not a kebab-case string. We could do lowercase, but as we don't yet have a validator for that, I'd prefer to merge it _as is_ right now.

ee7 (approved):

PR #295, opened 2021-04-30
lint(validators): support checking arrays for duplicates

ErikSchierboom:
Some string arrays have a constraint that their values must be unique, i.e. they must not be repeated.

The `isArrayOfStrings` and `hasArrayOfStrings` functions now take a `uniqueValues` argument with which uniqueness of the array values can be enforced. The default behavior is to allow duplicate values in an array.

### DIFF
--- a/src/lint/validators.nim
+++ b/src/lint/validators.nim
@@ -283,7 +283,9 @@ proc isArrayOfStrings*(data: JsonNode;
           elif uniqueValues:
             let itemStr = item.getStr()
             if processedItems.containsOrIncl(itemStr):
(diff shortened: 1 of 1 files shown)


ee7 (approved):
LGTM.

As discussed, it might be harder to extend this approach to checking things like

> - The `"authors"` values are treated case-insensitively
> - Users can only be listed in either the `"authors"` or `"contributors"` array (no overlap)
> - The `"exercises.concept[].slug"` value must be unique in `"exercises.concept[].slug"` and may not exist in `"exercises.practice[].slug"`

But we can handle that later.

Issue #64, opened 2020-10-21
Refactor some of the Nim code


ee7:
I intend to get around to refactoring some of the files that were out-of-scope in my first review. In particular:
- [ ] `tracks.nim`
- [ ] `probspecs.nim`
- [ ] `exercises.nim`

Maybe in this order. I'll do it after adding some tests so that we know it doesn't break everything.

PR #302, opened 2021-05-03
sync: prefer the object construction syntax

ee7:
Reasons:
- It works for both `object` and `ref object`. For the latter, we can
  remove the call to `new` - it is invoked implicitly for `ref object`.
- It's easier for the compiler to prove that an object has been fully
  initialized.
- It's arguably more readable.

---

See https://github.com/exercism/configlet/issues/64


ErikSchierboom (approved):
I like it!

PR #307, opened 2021-05-03
sync(tracks): refactor

ee7:
Some of this is a bit opinionated, but overall I think this PR improves readability and maintainability.

The commits are atomic for your reviewing pleasure, but I'll squash this PR.

I've deliberately limited the scope of this PR mostly to the non-exported things. But I have some further changes in the pipeline.

See also: #64

### DIFF
--- a/src/sync/tracks.nim
+++ b/src/sync/tracks.nim
@@ -3,7 +3,7 @@ import pkg/parsetoml
 import ".."/cli
 
 type
-  ConfigJsonExercise = distinct string
+  ConfigJsonSlug = distinct string
 
   TrackDir = distinct string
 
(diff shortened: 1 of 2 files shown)


src/sync/tracks.nim
```diff
@@ -22,73 +14,70 @@ type
...
-  for uuid, val in tests.getTable():
-    if val.hasKey("include"):
-      if val["include"].kind == Bool:
-        let isIncluded = val["include"].getBool()
-        if isIncluded:
-          result.included.incl(uuid)
+  PracticeExercisePath("").testsFile()
+
+proc getPracticeExercisePaths(trackDir: TrackDir): seq[PracticeExercisePath] =
+  let config = json.parseFile(trackDir / "config.json")["exercises"]
```
ErikSchierboom, line 32: We could consider exiting with an error if the `exercises` key does not exist, although that is extremely unlikely. It will probably only be the case for new tracks, but something to consider for later nevertheless.

src/sync/tracks.nim
```diff
@@ -3,17 +3,9 @@ import pkg/parsetoml
 import ".."/cli
 
 type
-  ConfigJsonExercise = object
-    slug: string
+  TrackDir {.requiresInit.} = distinct string
```
ErikSchierboom, line 6: I'm a big favor of these kinds of wrapper types 👍

ErikSchierboom (approved):
All looking good to me!

src/sync/tracks.nim
ee7: I thought the same. I'll propose improving the error for a missing `config.json` file too.

Coming soon.

PR #308, opened 2021-05-03
generate: add `generate` command

ErikSchierboom:
This PR adds basic support for the `generate` command: https://github.com/exercism/docs/blob/main/building/configlet/generating-documents.md

I'd really appreciate feedback on the code to see if there is anything wrong with it.

For now, I've decided to remove the level one heading from the Concept's `introduction.md` file when including it in the generated `introduction.md` file, as tracks might not want multiple level one headings. Also, a lot of the Concept `introduction.md` files have `# Introduction` as their starting (totally valid) level one heading, which would look really weird.

This PR currently does not support only generate the `introduction.md` for a single exercise, due to the inability to share properties within a `case`. This will be supported in a future version of Nim, but for now we should try an alternative approach.

### DIFF
--- a/src/generate/generate.nim
+++ b/src/generate/generate.nim
@@ -10,7 +10,8 @@ proc writeError(description: string, path: Path) =
   stderr.writeLine(path)
   stderr.write "\n"
 
-proc conceptIntroduction(trackDir: Path, slug: string, templatePath: Path): string =
+proc conceptIntroduction(trackDir: Path, slug: string,
+                         templatePath: Path): string =
   ## Returns the contents of the `introduction.md` file for a `slug`, but
   ## without any top-level heading, and without any leading/trailing whitespace.
   let conceptDir = trackDir / "concepts" / slug
@@ -20,7 +21,8 @@ proc conceptIntroduction(trackDir: Path, slug: string, templatePath: Path): stri
       let content = readFile(path)
       var idx = 0
       # Strip the top-level heading (if any)
-      if scanp(content, idx, *{' ', '\t', '\v', '\c', '\n', '\f'}, "#", +' ', +(~'\n')):
+      if scanp(content, idx, *{' ', '\t', '\v', '\c', '\n', '\f'}, "#", +' ',
+               +(~'\n')):
         result = content.substr(idx).strip
       else:
         result = content.strip
@@ -28,7 +30,8 @@ proc conceptIntroduction(trackDir: Path, slug: string, templatePath: Path): stri
       writeError(&"File {path} not found for concept '{slug}'", templatePath)
       quit(1)
   else:
-    writeError(&"Directory {conceptDir} not found for concept '{slug}'", templatePath)
+    writeError(&"Directory {conceptDir} not found for concept '{slug}'",
+               templatePath)
     quit(1)
 
 proc generateIntroduction(trackDir: Path, templatePath: Path): string =
--- a/src/generate/generate.nim
+++ b/src/generate/generate.nim
@@ -43,11 +43,11 @@ proc generateIntroduction(trackDir: Path, templatePath: Path): string =
   while idx < content.len:
     var conceptSlug = ""
     # Here, we implement the syntax for a placeholder as %{concept:some-slug}
-    # where we allow spaces/tabs after the opening brace, around the
-    # colon, and before the closing brace. The slug must be in kebab-case.
+    # where we allow spaces after the opening brace, around the colon,
+    # and before the closing brace. The slug must be in kebab-case.
     if scanp(content, idx,
-            "%{", *{' ', '\t'}, "concept", *{' ', '\t'}, ':', *{' ', '\t'},
-            +{'a'..'z', '-'} -> conceptSlug.add($_), *{' ', '\t'}, '}'):
+             "%{", *{' '}, "concept", *{' '}, ':', *{' '},
+             +{'a'..'z', '-'} -> conceptSlug.add($_), *{' '}, '}'):
       result.add conceptIntroduction(trackDir, conceptSlug, templatePath)
     else:
       result.add content[idx]
--- a/src/generate/generate.nim
+++ b/src/generate/generate.nim
@@ -3,12 +3,12 @@ import ".."/[cli, helpers]
 
 proc writeError(description, path: string) =
   let descriptionPrefix = description & ":"
-  if colorStdout:
-    stdout.styledWriteLine(fgRed, descriptionPrefix)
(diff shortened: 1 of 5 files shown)


src/generate/generate.nim
```diff
@@ -0,0 +1,47 @@
+import std/[os, strformat, strscans, strutils, terminal]
+import ".."/[cli, helpers]
+
+proc writeError(description, path: string) =
```
ErikSchierboom, line 4: This feels like something that should go in the `helpers` module possibly?

src/generate/generate.nim
```diff
@@ -0,0 +1,47 @@
...
+  else:
+    writeError(&"Referenced '{slug}' concept does not have an 'introduction.md' file", path)
+
+proc generateIntroduction(conf: Conf, templateFilePath: Path): string =
+  let content = readFile(templateFilePath)
+  
+  var idx = 0
+  while idx < content.len:
+    var conceptSlug = ""
+    if scanp(content, idx, "%{concept:", +{'a'..'z', '-'} -> conceptSlug.add($_), '}'):
```
ErikSchierboom, line 31: I just found out about `scanp` today, but I _really_ like it. Such an elegant way to handle parsing.

src/helpers.nim
```diff
@@ -39,3 +39,8 @@ proc setFalseAndPrint*(b: var bool; description: string; path: Path) =
 
 proc `$`*(path: Path): string {.borrow.}
 proc `/`*(head: Path; tail: string): Path {.borrow.}
+
+proc dirExists*(path: Path): bool {.borrow.}
+proc fileExists*(path: Path): bool {.borrow.}
+proc readFile*(path: Path): string {.borrow.}
+proc writeFile*(path: Path, content: string) {.borrow.}
```
ErikSchierboom, line 46: I had to move these as they are now used in more than one file.

ee7 (commented):
1. When the concept referred to does not exist, the implementation currently just writes the file without the `%{concept:slug}` contents. Is that what we want? Maybe we shouldn't write the file at all, and/or exit non-zero?
2. Would we prefer to have a single space before the slug? For example,
```
%{concept: user-defined-exceptions}
```
rather than
```
%{concept:user-defined-exceptions}
```
3. If not, should we at least support a single space before the slug?

ErikSchierboom:
When the concept referred to does not exist, the implementation currently just writes the file without the %{concept:slug} contents. Is that what we want? Maybe we shouldn't write the file at all, and/or exit non-zero?

We should indeed exit non-zero and not write the file.

> Would we prefer to have a single space before the slug? For example,

I would prefer _not_ to have a single space before the slug. @iHiD ?

iHiD:
Definitely no space thanks.

ErikSchierboom:
I've had a think, and maybe we should support spaces to be inserted within the enclosing `%{`, and `}` strings. My reason to allow this is that if someone accidentally makes a tiny mistake where a space is added, the user would not know how to fix it.

- `%{concept:user-defined-exceptions}`
- `%{concept:user-defined-exceptions }`
- `%{concept: user-defined-exceptions}`
- `%{ concept: user-defined-exceptions}`
- `%{ concept: user-defined-exceptions }`
- `%{  concept:  user-defined-exceptions  }`

ee7:
> a tiny mistake where a space is added, the user would not know how to fix it.

We could alter the parsing so we produce a good error message for whitespace usage that we want to be invalid. 

We should probably do that anyway, unless we wanted to silently accept/ignore a placeholder containing a newline:
```
here is some filler text so that this line is long enough that a text editor might hard line wrap %{concept:
user-defined-exceptions}
```
or a space **before** the colon:
```
%{concept : user-defined-exceptions}
```
or multiple consecutive spaces:
```
%{concept:  user-defined-exceptions}
```
In an extreme case:
```
%{                        concept:                       user-defined-exceptions                        }
```

But yes, we could also just silently accept the examples you give, or produce a warning.

ErikSchierboom:
@iHiD If you disagree with my above reasoning on allowing spaces, let me know and I'll adjust. We can have all the docs still only list the version without the spaces, but graciously handle them in configlet anyway.

iHiD:
I can't remember where this parsing actually **happens**. Is it in configlet (that syncs the doc) or is it in the website when it renders it?

ErikSchierboom:
@iHiD It is just in configlet.

iHiD:
Ah. Then my opinion weighs less. I'd say no space in the examples, but support spaces if they're put in maybe.

ErikSchierboom:
> Ah. Then my opinion weighs less. I'd say no space in the examples, but support spaces if they're put in maybe.

Cool, that's how it's implemented at the moment. The PR is ready for review @ee7

src/generate/generate.nim
ee7: Indeed, it's been missing in some other places too. I've opened https://github.com/exercism/configlet/issues/325 to track it.

ee7 (approved):
Looks great.

I've done some basic testing, but as future work it would be nice to have a test.

I'd probably use `scanf` or `parseutils` instead of the `scanp` in `conceptIntroduction`, but it's fine for now.

src/generate/generate.nim
```diff
@@ -0,0 +1,68 @@
...
+
+  var idx = 0
+  while idx < content.len:
+    var conceptSlug = ""
+    # Here, we implement the syntax for a placeholder as %{concept:some-slug}
+    # where we allow spaces/tabs after the opening brace, around the
+    # colon, and before the closing brace. The slug must be in kebab-case.
+    if scanp(content, idx,
+            "%{", *{' ', '\t'}, "concept", *{' ', '\t'}, ':', *{' ', '\t'},
+            +{'a'..'z', '-'} -> conceptSlug.add($_), *{' ', '\t'}, '}'):
```
ee7, line 50:
```suggestion
    # where we allow spaces after the opening brace, around the colon,
    # and before the closing brace. The slug must be in kebab-case.
    if scanp(content, idx,
             "%{", *{' '}, "concept", *{' '}, ':', *{' '},
             +{'a'..'z', '-'} -> conceptSlug.add($_), *{' '}, '}'):
```

Maybe better to disallow tabs?

ErikSchierboom:
> We're stripping the first top-level header of a concept introduction.md because we want to allow a track to write whatever header they want (at the cost of writing the same header twice when the desired header is identical), correct?

That is correct.

> I'd probably use scanf or parseutils instead of the scanp in conceptIntroduction, but it's fine for now.

Makes sense. I was just in awesome of `scanp` :)

PR #314, opened 2021-05-07
lint(validators): support checking exercise file paths

ErikSchierboom:
The `.meta/config.json` file of Concept and Practice Exercises contains a
`files` property, which contains the (relative) path to the various files
of the exercise. This PR adds a `hasArrayOfFiles` function that:

1. Check if the property is a string array using `hasArrayOfStrings`
2. For each string value, prepend that value with the exercise dir
   and check if a file exists at that path


ee7 (approved):
Looks great. I've tested that it works with #315.

Issue #316, opened 2021-05-07
sync: Add `reimplements` property to `tests.toml` file


ErikSchierboom:
Having a `reimplements` property added to the `tests.toml` file helps  identify re-implemented test cases:

```toml
[b9228bb1-465f-4141-b40f-1f99812de5a8]
description = "disallow first strand longer"
reimplements = "919f8ef0-b767-4d1b-8516-6379d07fcb28"
```

PR #317, opened 2021-05-07
sync: output `reimplements` property

ErikSchierboom:
This commits adds a `reimplements` property to the `tests.toml` file to help identify re-implemented test cases:

```toml
[b9228bb1-465f-4141-b40f-1f99812de5a8]
description = "disallow first strand longer"
reimplements = "919f8ef0-b767-4d1b-8516-6379d07fcb28"
```

Closes https://github.com/exercism/configlet/issues/316

@ee7 This was suggested during one of the weekly meetups, and I thought it made sense. I've create a small PoC. Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,9 +44,16 @@ If a track implements an exercise for which test data exists in the [problem-spe
 A `tests.toml` file has this format:
 
 ```toml
-# This is an auto-generated file. Regular comments will be removed when this
-# file is regenerated. Regenerating will not touch any manually added keys,
-# so comments can be added in a "comment" key.
+# This is an auto-generated file.
+#
+# Regenerating this file will:
+# - Update the `description` property
+# - Update the `reimplements` property
+# - Remove `include = true` properties
+# - Preserve any other properties
+#
+# As regular comments will be removed when this file is regenerated, comments
+# can be added in a "comment" key
 
 [1e22cceb-c5e4-4562-9afe-aef07ad1eaf4]
 description = "basic"
--- a/src/sync/exercises.nim
+++ b/src/sync/exercises.nim
@@ -145,8 +145,8 @@ proc toToml(exercise: Exercise, testsPath: string): string =
         result.add "include = false\n"
 
       # Always output the `reimplements` value, if present
-      if testCase.reimplements.isSome:
-        result.add &"reimplements = \"{testCase.reimplements.get.uuid}\"\n"
+      if testCase.reimplements.isSome():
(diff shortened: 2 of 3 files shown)


ee7 (commented):
I like it. The only downside I see is that a user might think that we actually use the `reimplements` value in some way. But actually, it's just there to improve readability/understanding, right? Nothing bad happens if the user accidentally edits the `reimplements` UUID. 

I think it's fine, but do you think it's worth trying to explain `reimplements` in the comment at the top of the file? Now that I look at it again, should we explain the implicit `include = true` too?

Some nitpicks in #318.

ErikSchierboom:
> I like it. The only downside I see is that a user might think that we actually use the reimplements value in some way. But actually, it's just there to improve readability/understanding, right? Nothing bad happens if the user accidentally edits the reimplements UUID.

That is correct.

> I think it's fine, but do you think it's worth trying to explain reimplements in the comment at the top of the file? Now that I look at it again, should we explain the implicit include = true too?

We probably should.

README.md
```diff
@@ -44,9 +44,16 @@ If a track implements an exercise for which test data exists in the [problem-spe
...
+# This is an auto-generated file.
+#
+# Regenerating this file will:
+# - Update the `description` property
+# - Update the `reimplements` property
+# - Remove `include = true` properties
+# - Preserve any other properties
+#
+# As regular comments will be removed when this file is regenerated, comments
+# can be added in a "comment" key
```
ErikSchierboom, line 56: I've gone and updated the comment at the top of the file to be a bit more in-depth, to give the maintainer a better idea of what will happen when resyncing. Let me know what you think.

ee7 (approved):
I'd like to tweak the comments at the top of the `tests.toml` file, but I didn't find a wording I was happy with. So let's just handle that later - I'll make the PR.

Some potential changes:
- It looks like "property" isn't actually the preferred terminology for TOML (oops) - I think it's just "key/value pair".
- We might like to mention `configlet sync` somewhere, rather than just "regenerating" (which might be confused with `configlet generate`).

I have another PR that should use this PR as a base (several PRs require changing the integration tests for `sync`) so I'll merge this.

ErikSchierboom:
Thanks!

>It looks like "property" isn't actually the preferred terminology for TOML (oops) - I think it's just "key/value pair".

Whoops.

> We might like to mention configlet sync somewhere, rather than just "regenerating" (which might be confused with configlet generate).

Yeah I considered this. I struggled a bit with the wording, so I omitted it for now.

Issue #61, opened 2020-10-20
If a test case was added that reimplements another case, print the reimplemented UUID as well


cmccandless:
Related to #44

For this example, assume that the `grains` case `10f45584-2fc3-4875-8ec6-666065d1163b` reimplements another case `b4bacc5a-3c12-43b6-ab17-d5ec78d70a15`

### As of v0.10.0:

```
$ ./canonical_data_syncer -c --verbosity=detailed
[...]
[warn] grains: missing 2 test cases
       - grains on square 3 (10f45584-2fc3-4875-8ec6-666065d1163b)
```

### Proposed

Put reimplemented case description and UUID on following line (for human readability)

```
$ ./canonical_data_syncer -c --verbosity=detailed
[...]
[warn] grains: missing 2 test cases
       - grains on square 3 (10f45584-2fc3-4875-8ec6-666065d1163b)
         - reimplements: square 3 grains (b4bacc5a-3c12-43b6-ab17-d5ec78d70a15)
```

### Alternative notation

Put reimplemented case UUID on same line (simpler to parse for scripts)

```
$ ./canonical_data_syncer -c --verbosity=detailed
[...]
[warn] grains: missing 2 test cases
       - grains on square 3 (10f45584-2fc3-4875-8ec6-666065d1163b -> b4bacc5a-3c12-43b6-ab17-d5ec78d70a15)
```

ErikSchierboom:
> Put reimplemented case UUID on same line (simpler to parse for scripts)

IMHO we should optimize this for readability, not for the parsing by scripts. If it turns out that is something people want to do, we should perhaps look at a JSON output option.

cmccandless:
My personal use-case for scripting: I want to create an automated workflow that does the following nightly:
1. Scans for missing test cases
2. For each missing case, create a PR that includes that case and runs the track's test generator

ee7:
> If a test case was added that reimplements another case, print the reimplemented UUID as well

Good idea.

> we should perhaps look at a JSON output option.

I agree - I don't think people should consider the `--check` output as stable enough to parse, at least right now.

> I want to create an automated workflow that does the following nightly

It'd be great if we can support this kind of workflow.

ErikSchierboom:
@cmccandless Could you open an issue for that?

ErikSchierboom:
I've tried output several notations

Option 1:

```
[warn] grains: missing 2 test cases
[...]

PR #318, opened 2021-05-07
review: "sync: output reimplements property (#317)"

ee7:
Some suggestions for #317.

Changes:
- Make the tests pass.
- Add some parens. This is an opinionated change, and isn't a strict improvement in readability. But I think it helps make the codebase more reader-friendly when done consistently. It also currently improves the syntax highlighting on GitHub and in many editors.
- Bikeshed comment - I didn't love "output" and "value" because it could sound like it addresses https://github.com/exercism/configlet/issues/61, but we'll address that elsewhere.

### DIFF
--- a/src/sync/exercises.nim
+++ b/src/sync/exercises.nim
@@ -124,7 +124,7 @@ proc toToml(exercise: Exercise, testsPath: string): string =
       if testCase.uuid notin exercise.tests.included:
         result.add "include = false\n"
 
-      # Always output the `reimplements` value, if present
+      # Always add the `reimplements` property, if present
       if testCase.reimplements.isSome():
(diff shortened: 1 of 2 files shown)


ErikSchierboom (approved):
I like it. I'm always unsure when to use parentheses.

ee7:
Another reputation edge case: this PR awarded 12 rep. Is that what we want?

ErikSchierboom:
> Another reputation edge case: this PR awarded 12 rep. Is that what we want?
> This style of reviewing is less common in Exercism repos, though. But I like it much better than GitHub's "suggested changes" - I can check that the code compiles and tests pass myself. And edit files/lines that weren't touched by the parent PR.

I think this is indeed quite uncommon, but might be worth discussing. We could add the `reputation/contributed_code/minor` label to reduce the amount of rep by 3.

PR #320, opened 2021-05-08
tests(sync/tracks): compare objects, not strings

ee7:
We refactored `sync/tracks.nim` to remove a non-exported field from the
`TrackExercise` object. Therefore we can now simplify the test like
this, without changing field/type visibility just for the purpose of
testing.

### DIFF
--- a/tests/test_tracks.nim
+++ b/tests/test_tracks.nim
@@ -1,22 +1,6 @@
 # This file implements tests for `src/tracks.nim`
-import std/[os, osproc, parseutils, sets, strformat, unittest]
+import std/[os, osproc, sets, strformat, unittest]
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
Much cleaner

ee7:
> Much cleaner

Indeed. This was always the plan, but here I preferred the pattern of

0. Have existing code that lacks a test
1. Add a test
2. Refactor the code
3. Refactor the test

over

0. Have existing code that lacks a test
1. Refactor the code
2. Add a test

because it also checks that the refactoring worked as expected.

ErikSchierboom:
Absolutely agree with the plan.

PR #324, opened 2021-05-08
git: ignore all binaries

ee7:
This helps to reduce the noise in `git status` for a `configlet`
contributor who, for example:
- wants to compile some already committed Nim files individually, and
  sometimes uses `nim c -r` rather than `nim r`.
- has created some files that should be available on any branch, and so
  are kept uncommitted.

---

Let's just say that even **with** this PR, on my machine:
```
$ git status -s | wc -l
78
```

So this helps a bit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,25 @@
-# IDE files
+# The first 3 rules below cause every file without an extension to be ignored
+# Ignore everything
+*
+
+# Unignore all directories, including subdirectories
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
I feel like without the comments, this would be the most unreadable `.gitignore` file I've ever seen :) But I'm okay with it due to the comments.

PR #330, opened 2021-05-12
lint: require more arrays to have unique values

ee7:
This PR adds the new linting rules from https://github.com/exercism/docs/pull/127

With this commit, `configlet lint` now checks the following rules for a
track-level `config.json` file:
- The `exercises.foregone` values must not have duplicates
- The `tags` values must not have duplicates

And these rules for a Concept Exercise `config.json` file:
- The `files.solution` values must not have duplicates
- The `files.test` values must not have duplicates
- The `files.exemplar` values must not have duplicates

And these rules for a Practice Exercise `config.json` file:
- The `files.solution` values must not have duplicates
- The `files.test` values must not have duplicates
- The `files.example` values must not have duplicates

---

This PR currently keeps the output of `configlet lint` the same on every track.

### DIFF
--- a/src/lint/track_config.nim
+++ b/src/lint/track_config.nim
@@ -43,7 +43,8 @@ const tags = [
 ].toHashSet()
 
 proc hasValidTags(data: JsonNode; path: Path): bool =
-  result = hasArrayOfStrings(data, "tags", path, allowed = tags)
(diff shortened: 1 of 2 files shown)


ErikSchierboom (approved):
Thanks!

PR #332, opened 2021-05-14
sync: another refactoring pass

ee7:
Not everything here is strictly better, but it's hopefully in the right direction overall.

I've got lots more tidying up in the pipeline - the uglier changes shouldn't be here in the long-term.

This PR is part of a series that addresses #64.

### DIFF
--- a/src/sync/exercises.nim
+++ b/src/sync/exercises.nim
@@ -31,11 +31,11 @@ func initExerciseTests*(included, excluded, missing: HashSet[string]): ExerciseT
     missing: missing,
   )
 
-proc initExerciseTests(trackExercise: TrackExercise, probSpecsExercise: ProbSpecsExercise): ExerciseTests =
+proc initExerciseTests(practiceExercise: PracticeExercise, probSpecsExercise: ProbSpecsExercise): ExerciseTests =
   for testCase in probSpecsExercise.testCases:
-    if trackExercise.tests.included.contains(testCase.uuid):
+    if practiceExercise.tests.included.contains(testCase.uuid):
       result.included.incl(testCase.uuid)
-    elif trackExercise.tests.excluded.contains(testCase.uuid):
+    elif practiceExercise.tests.excluded.contains(testCase.uuid):
       result.excluded.incl(testCase.uuid)
     else:
       result.missing.incl(testCase.uuid)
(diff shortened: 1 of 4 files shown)


src/sync/exercises.nim
```diff
@@ -1,8 +1,8 @@
-import std/[algorithm, json, options, os, sequtils, sets, strformat, strutils,
-            tables]
+import std/[json, options, os, sequtils, sets, strformat, strutils, tables]
 import pkg/parsetoml
 import ".."/cli
 import "."/[probspecs, tracks]
+export tracks.`$`
```
ee7, line 5: This will probably be refactored away later.

tests/test_tracks.nim
```diff
@@ -52,7 +52,7 @@ proc main =
         )
 
       check:
-        trackExercises[1] == expectedTwoFer
+        trackExercises[65] == expectedTwoFer
```
ee7, line 55: I don't love this, but I guess it's fine.

src/sync/exercises.nim
```diff
@@ -68,7 +69,7 @@ proc findExercises*(conf: Conf): seq[Exercise] =
   let probSpecsExercises = findProbSpecsExercises(conf).mapIt((it.slug, it)).toTable
 
   for practiceExercise in findPracticeExercises(conf):
-    result.add(initExercise(practiceExercise, probSpecsExercises.getOrDefault(practiceExercise.slug)))
+    result.add(initExercise(practiceExercise, probSpecsExercises.getOrDefault(practiceExercise.slug.string)))
```
ee7: Most of the things that use `.string` should be refactored away entirely, or become a `{.borrow.}`

src/sync/exercises.nim
ErikSchierboom: Good. It looks slightly odd :)

src/sync/exercises.nim
```diff
@@ -47,29 +48,48 @@ proc newExerciseTestCase(testCase: ProbSpecsTestCase): ExerciseTestCase =
...
+
+proc initExerciseTestCases(testCases: seq[ProbSpecsTestCase]): seq[ExerciseTestCase] =
+  result = newSeq[ExerciseTestCase](testCases.len)
+  for i, testCase in testCases:
+    result[i] = newExerciseTestCase(testCase)
 
-  let reimplementations = testCases.filterIt(it.isReimplementation).mapIt((it.uuid, it.reimplements)).toTable()
-  let testCasesByUuids = result.newTableFrom(proc (testCase: ExerciseTestCase): string = testCase.uuid)
+  let reimplementations = getReimplementations(testCases)
+  let testCasesByUuids =  uuidToTestCase(result)
```
ErikSchierboom, line 66: Was the space intentional?

```suggestion
  let testCasesByUuids = uuidToTestCase(result)
```

ErikSchierboom (approved):

src/sync/exercises.nim
ee7: Wasn't intentional - thanks.

PR #339, opened 2021-05-28
lint: remove ace_editor_language check

ErikSchierboom:
With the switch to CodeMirror as the online editor library, the
`online_editor.ace_editor_language` key is now obsolete and its presence
should not be checked anymore. This PR removes that check.


ee7 (approved):
Looks good.

I've made a PR to also remove it from the spec: https://github.com/exercism/docs/pull/137

Issue #312, opened 2021-05-06
lint: consider the JSON parsing/deserialization design


ee7:
Main options:
1. [`std/json`](https://nim-lang.github.io/Nim/json.html)
   1a. The approach so far: parse into a `JsonNode` and work only with that.
   1b. Parse into a `JsonNode`, then unmarshall into some object using [`to`](https://nim-lang.github.io/Nim/json.html#to%2CJsonNode%2Ctypedesc%5BT%5D).
1. [`std/jsonutils`](https://nim-lang.github.io/Nim/jsonutils.html) - still requires going via the stdlib `JsonNode`.
1. [`Araq/packedjson`](https://github.com/Araq/packedjson) - keeps everything as a string.
1. [`status-im/nim-json-serialization`](https://github.com/status-im/nim-json-serialization) - deserializes using [`nim-faststreams`](https://github.com/status-im/nim-faststreams) directly to an `object`. Probably the most mature third-party option.
1. [`treeform/jsony`](https://github.com/treeform/jsony) - deserializes from `string` directly to an `object`. 
1. [`planetis-m/eminim`](https://github.com/planetis-m/eminim) - deserializes using `std/streams` directly to an `object`. Doesn't fully support object variants, but maybe that isn't a problem for us.

Some of the above are possibly too lenient or require special handling in some edge cases.

For example:
- There is no correct behaviour for a duplicate key - one library may produce an error, another may silently use the value of the first key, and another may silently use the value of the last key.
- `std/json` permits a trailing comma and comments with `//`. This is the main reason that I never checked the boxes for "the file must be valid JSON" boxes in https://github.com/exercism/configlet/issues/249. But it's probably simple to use a modified `std/json` with stricter parsing.
- Some libraries may use a default value when the key is missing, which we might want to distinguish from e.g. a value that is the empty string.
- Edge cases around a literal `null`.

See also:
- https://einarwh.wordpress.com/2020/05/08/on-the-complexity-of-json-serialization/
- https://labs.bishopfox.com/tech-blog/an-exploration-of-json-interoperability-vulnerabilities

I'd suggest that `jsony` or `nim-json-serialization` might be best in the long-term. But maybe it's better to stick with the current approach until we've implemented all the linting rules, and refactor it later.

One advantage of the current approach is that it's more low-level, which helps to ensure that we're "checking the JSON file itself" rather than "checking that each value is valid when parsed with library X".

ErikSchierboom:
[...]

PR #346, opened 2021-06-02
lint, sync: make JSON parsing stricter

ee7:
This PR is the "option 2" approach I mentioned in https://github.com/exercism/configlet/issues/312#issuecomment-835800075. I started trying the "run `jq` on everything" way, but this implementation is better because:
- It doesn't require/prefer the user to have `jq` installed, or special-case running in CI
- The current linting code doesn't build a list of all the JSON files - it just iterates over each one. But naively calling `jq` once for each file is too much of a performance hit - e.g. about 10 seconds to lint the `csharp` track, rather than about 20 ms. So we'd want to pass every file as an argument to and call `jq` once, but this makes our code longer and more complex, makes it hard for the errors for the same JSON file to be grouped together, and has no benefit over the current pure Nim approach anyway (as far as I know).

To-do:
- [x] Add tests
- [x] Check the output of `configlet lint` on every track.
- [x] Do some quick profiling and see if the "minimal diff" parsing change should be refactored slightly.
- [x] ~Consider renaming the files. Perhaps `strict_json.nim` and `strict_parsejson.nim`~ Bad idea. Requires non-intuitive changes to the existing codebase.
- [x] Consider putting them in a separate folder.
- [x] Consider moving the `patchFile`s to the top of `config.nims`, and ordering them alphabetically.

Closes: #345


ErikSchierboom (approved):
So we're basically patching the stdlib's implementation? Fancy.

I think as an approach this is fine, as long as we remember to update this file once the upstream file is changed.

ee7:
Added some tests, and tidied up a few things. Could you take another quick look?

> So we're basically patching the stdlib's implementation?

Yes. The `patchFile` mechanism is somewhat obscure, but works. One caveat is that the files should keep the same name, otherwise explicitly prefixing a call with the module name (e.g. `json.parseFile` to disambiguate from `parsetoml.parseFile`) doesn't work. We'd have to change to e.g. `strict_json.parseFile`, which looks really weird when the import itself is still `import std/json`.

We'll just maintain our slightly different implementation for now, but it might become unnecessary in the future - for example, if we ever move to some other JSON library, or if the upstream `parseJson` gains e.g. a `allowTrailingCommas = true` argument (I intend to PR the latter at some point).

> as long as we remember to update this file

Indeed. 

>  once the upstream file is changed.

Let's do it when we bump the Nim version - we should keep the file in sync with the stable release branch (currently `version-1-4`).

Thankfully, the upstream files don't change that often, or that much. For the history of these files on the release branch, see:
- https://github.com/nim-lang/Nim/commits/version-1-4/lib/pure/json.nim
- https://github.com/nim-lang/Nim/commits/version-1-4/lib/pure/parsejson.nim (`parsejson` has barely changed since 2011)

PR #349, opened 2021-06-04
sync: improve `tests.toml` comment

ee7:
Follow-up from https://github.com/exercism/configlet/pull/317#pullrequestreview-671985491

I think it's pretty hard to find a great wording here. Any suggestions?

Summary:
- Don't say "update" - we recreate a `description`, even if the `tests.toml` lacked it.
- Mention `configlet sync`, so that people don't think of `configlet generate`
- Say "key/value pair", rather than "property" (which I think is incorrect terminology for TOML)
- Explicitly say that `include = true` is implied

I'll update this PR to fix the tests later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,14 +46,14 @@ A `tests.toml` file has this format:
 ```toml
 # This is an auto-generated file.
 #
-# Regenerating this file will:
-# - Update the `description` property
-# - Update the `reimplements` property
-# - Remove `include = true` properties
-# - Preserve any other properties
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
 #
-# As regular comments will be removed when this file is regenerated, comments
-# can be added in a "comment" key
+# As other TOML comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
 
 [1e22cceb-c5e4-4562-9afe-aef07ad1eaf4]
 description = "basic"
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ A `tests.toml` file has this format:
 # - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
 # - Preserve any other key/value pair
 #
-# As other TOML comments (using the # character) will be removed when this file
+# As user-added comments (using the # character) will be removed when this file
 # is regenerated, comments can be added via a `comment` key.
 
 [1e22cceb-c5e4-4562-9afe-aef07ad1eaf4]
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -46,14 +46,14 @@ A `tests.toml` file has this format:
...
-# - Preserve any other properties
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
 #
-# As regular comments will be removed when this file is regenerated, comments
-# can be added in a "comment" key
+# As other TOML comments (using the # character) will be removed when this file
```
ErikSchierboom, line 55: I found the "other" a bit weird. I know what you're trying to describe here, but I'm not sure it is very clear with this wording. The question then becomes: how _do_ we want to word it? Maybe `user-added` comments? Or `Any comments (except these introductory comments)`?

README.md
```diff
@@ -46,14 +46,14 @@ A `tests.toml` file has this format:
...
 #
-# Regenerating this file will:
-# - Update the `description` property
-# - Update the `reimplements` property
-# - Remove `include = true` properties
-# - Preserve any other properties
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` (an omitted `include` key implies inclusion)
```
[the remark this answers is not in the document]
ee7: I think so. Added.

README.md
ee7: Yeah, "other" was bad. I've added a commit for "user-added".

README.md
```diff
@@ -46,14 +46,14 @@ A `tests.toml` file has this format:
...
 # This is an auto-generated file.
 #
-# Regenerating this file will:
-# - Update the `description` property
-# - Update the `reimplements` property
-# - Remove `include = true` properties
-# - Preserve any other properties
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
```
ee7, line 51: I don't love "recreate", since it doesn't communicate very well that the latest value of the description is used.

The problem with just "update" was that the `description` and `reimplements` properties are added even if they didn't previously exist in the `tests.toml` for some reason.

ee7:
@ErikSchierboom Any other ideas for improving the wording?

I think this PR is currently an improvement, but I'm not completely happy with it. Do you think it's good enough for now? And did you want to do a mass PR to tracks with the recent `tests.toml` changes (nested `description`, `reimplements`, updated introductory comment)? If so, I'd like us to be relatively happy with the wording before doing that.

I'm not saying that we should do such a mass PR, or that it needs to be done immediately. But maintainers might not like the noisy diff when updating practice exercises with canonical data, so maybe it's worth doing sometime?

ErikSchierboom (approved):
👍

ErikSchierboom:
> Do you think it's good enough for now?

Absolutely.

> I'm not saying that we should do such a mass PR, or that it needs to be done immediately. But maintainers might not like the noisy diff when updating practice exercises with canonical data, so maybe it's worth doing sometime?

Yeah, would be nice, but low priority.

PR #359, opened 2021-06-11
sync(exercises): remove duplicate `tests.toml` parsing

ee7:
Under some circumstances, we previously parsed `tests.toml` once per test case (!).


ErikSchierboom (approved):
😮

PR #380, opened 2021-06-29
tests: refactor to share code for exec and git

ee7:
This PR is a follow-up of #379, where I wrote:

> I was going to push these changes (and much more upcoming) to #366, but it makes that PR (and eventual squashed commit) too large - it'll already be pretty big. So I want to merge a few PRs first - this won't slow things down more. I consider adding decent tests one of the biggest parts of #366, so in this case I want to not hate the initial state of the new tests.

> The next PR will improve the cloning process further, so we'll just have `setupExercismRepo` at the call site that does "clone if necessary, and checkout".

> I think I've finally figured out an ergonomic design for some exec/git helper commands, so they'll soon infect the rest of our codebase too.

See the individual commits.

The hard question: can we think of a better name than `myExecCmdEx`? I might use just `exec` instead - it's hard for a reader to confuse with [`nimscript.exec`](https://nim-lang.github.io/Nim/nimscript.html#exec%2Cstring) because that proc doesn't return anything.

A later PR will make `src/probspecs.nim` also use this shared code. And other things in `tests/test_binary.nim` too.

### DIFF
--- a/src/exec.nim
+++ b/src/exec.nim
@@ -42,7 +42,8 @@ proc git*(args: openArray[string]): ProcessResult =
   result = myExecCmdEx("git", args = args)
 
 proc cloneExercismRepo*(repoName, dest: string; isShallow = false) =
-  ## Clones the Exercism repo named `repoName` to the location `dest`.
+  ## Clones the Exercism repo named `repoName` to the location `dest`. Performs
+  ## a shallow clone if `isShallow` is `true`.
   ##
   ## Quits if unsuccessful.
   let url = &"https://github.com/exercism/{repoName}/"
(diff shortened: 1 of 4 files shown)


src/exec.nim
```diff
@@ -0,0 +1,83 @@
...
+  var line = newStringOfCap(120)
+  while true:
+    if outp.readLine(line):
+      result[0].add line
+      result[0].add "\n"
+    else:
+      result[1] = peekExitCode(p)
+      if result[1] != -1:
+        break
+  close(p)
```
ee7, line 38: See the current upstream implementation of [`osproc.execCmdEx`](https://github.com/nim-lang/Nim/blob/6387e289631c22b8fa81e8e478c97cc500b90e26/lib/pure/osproc.nim#L1571-L1630).

src/exec.nim
```diff
@@ -0,0 +1,83 @@
+import std/[os, osproc, streams, strformat, strtabs]
+
+type
+  ProcessResult* = tuple
+    output: string
+    exitCode: int
+
+proc myExecCmdEx*(command: string; args: openArray[string] = [];
```
ErikSchierboom, line 8: As suggested, I also like `exec` here better.

ErikSchierboom (approved):

PR #406, opened 2021-09-01
tests(binary): add more tests for sync

ee7:
Let's add our first tests for `--exercise` and the interactive mode, with a user typing `y`/`n`.

I'll refactor this later to be less terrible.

### DIFF
--- a/tests/test_binary.nim
+++ b/tests/test_binary.nim
@@ -161,6 +161,47 @@ proc testsForSync(binaryPath: string) =
     let anagramTestsTomlPath = joinPath("exercises", "practice", "anagram", ".meta", "tests.toml")
     check execCmdEx(&"git -C {trackDir} restore {anagramTestsTomlPath}")[1] == 0
 
+    test "single-exercise `sync --update --mode=exclude` exits with 0 and excludes the expected test cases":
+      execAndCheck(0):
+        execCmdEx(&"{binaryPath} -t {trackDir} sync --update -me -o -p {psDir} -e anagram")
+
+      check outp == """
+        Syncing exercises...
+        [info] anagram: excluded 1 missing test case
+        All exercises are synced!
+      """.unindent()
+
+    test "`git diff` shows the expected diff":
+      execAndCheck(0):
+        execCmdEx(diffCmd)
+
+      const expectedAnagramDiffExclude = """
+        --- exercises/practice/anagram/.meta/tests.toml
+        +++ exercises/practice/anagram/.meta/tests.toml
+        -# This is an auto-generated file. Regular comments will be removed when this
+        -# file is regenerated. Regenerating will not touch any manually added keys,
+        -# so comments can be added in a "comment" key.
+        +# This is an auto-generated file.
+        +#
+        +# Regenerating this file via `configlet sync` will:
+        +# - Recreate every `description` key/value pair
+        +# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+        +# - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
+        +# - Preserve any other key/value pair
+        +#
+        +# As user-added comments (using the # character) will be removed when this file
+        +# is regenerated, comments can be added via a `comment` key.
+        +[03eb9bbe-8906-4ea0-84fa-ffe711b52c8b]
+        +description = "detects two anagrams"
+        +include = false
+        +reimplements = "b3cca662-f50a-489e-ae10-ab8290a09bdc"
+        +
+      """.unindent()
+
+      check outp.conciseDiff() == expectedAnagramDiffExclude
+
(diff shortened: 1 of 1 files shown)


tests/test_binary.nim
```diff
@@ -94,6 +94,141 @@ proc testsForSync(binaryPath: string) =
         [warn] some exercises are missing test cases
       """.dedent(8) # Not `unindent`. We want to preserve the indentation of the list items.
 
+    test "a single-exercise `sync` without `--update` exits with 1 and prints the expected output":
```
ErikSchierboom, line 97: This might be slightly much, but maybe be slightly more specific about the test case?

```suggestion
    test "a single-exercise `sync` without `--update` and missing test case exits with 1 and prints the expected output":
```

tests/test_binary.nim
```diff
@@ -94,6 +94,141 @@ proc testsForSync(binaryPath: string) =
...
+
+      check outp == """
+        Checking exercises...
+        [warn] anagram: missing 1 test case
+               - detects two anagrams (03eb9bbe-8906-4ea0-84fa-ffe711b52c8b)
+        [warn] some exercises are missing test cases
+      """.dedent(8)
+
+    test "when passing multiple exercises, only the final exercise is acted upon":
+      # TODO: We should instead support multiple exercises being passed.
```
ErikSchierboom, line 109: Actually, I don't think we necessarily need to support this. It seems like rather an edge case.

ErikSchierboom (approved):

tests/test_binary.nim
ee7: Good catch, thanks. The previous description implied that `sync` without `--update` always exits with 1.

I've pushed https://github.com/exercism/configlet/pull/406/commits/735138dc8b4 to this PR, which tries to clarify the other test names too.

#### Before that commit:
```
[Suite] sync
  [OK] a `sync` without `--update` exits with 1 and prints the expected output
  [OK] a single-exercise `sync` without `--update` exits with 1 and prints the expected output
  [OK] when passing multiple exercises, only the final exercise is acted upon
  [OK] single-exercise `sync --update --mode=include` exits with 0 and includes the expected test cases
  [OK] `git diff` shows the expected diff
  [OK] single-exercise `sync --update --mode=exclude` exits with 0 and excludes the expected test cases
  [OK] `git diff` shows the expected diff
  [OK] single-exercise `sync --update` with input of 'y' includes as expected
  [OK] `git diff` shows the expected diff
  [OK] single-exercise `sync --update` with input of 'n' excludes as expected
  [OK] `git diff` shows the expected diff
  [OK] single-exercise `sync --update` with input of 's' exits with 1
  [OK] `git diff` shows the expected diff
  [OK] `sync --update --mode=include` exits with 0 and includes the expected test cases
  [OK] `git diff` shows the expected diff
  [OK] after syncing, another `sync --update --mode=include` performs no changes
  [OK] after syncing, a `sync` without `--update` shows that exercises are up to date
  [OK] the `git diff` output is still the same
```

#### With that commit:
```
[Suite] sync, without --update
  [OK] multiple exercises with missing test cases: prints the expected output, and exits with 1
  [OK] a given exercise with a missing test case: prints the expected output, and exits with 1
  [OK] when passing multiple exercises, only the final exercise is acted upon

[Suite] sync, with --update
  [OK] -mi: includes a missing test case for a given exercise, and exits with 0
  [OK] the diff is as expected
  [OK] -me: excludes a missing test case for a given exercise, and exits with 0
  [OK] the diff is as expected
  [OK] -mc: includes a missing test case for a given exercise when the input is 'y', and exits with 0
  [OK] the diff is as expected
  [OK] -mc: excludes a missing test case for a given exercise when the input is 'n', and exits with 0
  [OK] the diff is as expected
  [OK] -mc: neither includes nor excludes a missing test case for a given exercise when the input is 's', and exits with 1
  [OK] the diff is as expected
  [OK] -mi: includes every missing test case when not specifying an exercise, and exits with 0
  [OK] the diff is as expected
  [OK] after updating, another update using -mi performs no changes, and exits with 0
  [OK] after updating, a `sync` without `--update` shows that exercises are up to date, and exits with 0
  [OK] the diff is still the same
```

tests/test_binary.nim
ee7: Agreed. It's not a priority for me - I'm mainly adding a test case while I'm here to remind myself of the current behavior.

That said, if the user **does** pass multiple `--exercise` arguments, the possible behaviors seem to be:
1. Act on the first one (and probably print a warning)
1. Act on the final one (and probably print a warning)
1. Fail with an error, saying the only one `--exercise` is allowed
1. Act on all of the given exercises

In which case I probably prefer option 4. I think it can sometimes be useful to the maintainer, and is the least surprising and least friction for somebody who does try multiple `--exercise` arguments. And maybe it's useful for scripting.

I don't like the current behavior of acting on just one exercise without a warning. But we can discuss later.

PR #414, opened 2021-09-06
lint(track_config): fix handling of missing `status`

ee7:
A Concept Exercise in the track-level `config.json` file may have a
`status` property. If that property is missing, the implied value is
`active`, and so the Concept Exercise appears on the website.

With this commit, `configlet lint` no longer skips some checks of a
Concept Exercise when its `status` property is missing.


ErikSchierboom:
> A missing status property is the same thing as a status of active - correct?

That is correct.

ErikSchierboom (approved):

PR #417, opened 2021-09-07
lint(track_config): add more checks of exercise slugs

ee7:
With this PR, `configlet lint` now checks that a track-level
`config.json` file follows the below rules:

- The `exercises.concept[].slug` value must be unique in
  `exercises.concept[].slug`  and may not exist in
  `exercises.practice[].slug`
- The `exercises.practice[].slug` value must be unique in
  `exercises.practice[].slug` and may not exist in
  `exercises.concept[].slug`
- There must be exactly one `exercises.practice[].slug` value that is
  the string `hello-world`
- The `exercises.foregone` values must not match any of the concept or
  practice exercise slugs

---

At the time of writing (`2021-09-07T13:37:00Z`), this PR produces the following diff to the output of `configlet lint`, per track:

#### julia
https://github.com/exercism/julia/blob/0a134640aeed/config.json#L102-L296
```diff
+The slug `leap` is used for both a Concept Exercise and a Practice Exercise, but must only appear once on the track:
+./config.json
+

```

Looks like Sascha has a PR for it: https://github.com/exercism/julia/pull/365

### DIFF
--- a/src/lint/track_config.nim
+++ b/src/lint/track_config.nim
@@ -389,22 +389,22 @@ proc checkExercisesPCP(exercises: seq[ConceptExercise] | seq[PracticeExercise];
               "Exercise is allowed to have that"
       b.setFalseAndPrint(msg, path)
 
-proc checkExerciseSlugsAndForegone(trackConfig: TrackConfig; b: var bool;
+proc checkExerciseSlugsAndForegone(exercises: Exercises; b: var bool;
                                    path: Path) =
   ## Sets `b` to `false` if the below conditions are not satisfied:
-  ## - Each slug of a Concept Exercise or a Practice Exercise in `trackConfig`
+  ## - Each slug of a Concept Exercise or a Practice Exercise in `exercises`
   ##   only exists once on the track.
   ## - There is exactly one Practice Exercise with the slug `hello-world`.
   ## - The `foregone` array does not contain a slug of an implemented exercise.
-  var conceptExerciseSlugs = initHashSet[string](trackConfig.exercises.`concept`.len)
-  for conceptExercise in trackConfig.exercises.`concept`:
+  var conceptExerciseSlugs = initHashSet[string](exercises.`concept`.len)
+  for conceptExercise in exercises.`concept`:
     let slug = conceptExercise.slug
     if conceptExerciseSlugs.containsOrIncl slug:
       let msg = &"There is more than one Concept Exercise with the slug {q slug}"
       b.setFalseAndPrint(msg, path)
 
-  var practiceExerciseSlugs = initHashSet[string](trackConfig.exercises.practice.len)
-  for practiceExercise in trackConfig.exercises.practice:
+  var practiceExerciseSlugs = initHashSet[string](exercises.practice.len)
+  for practiceExercise in exercises.practice:
     let slug = practiceExercise.slug
     if practiceExerciseSlugs.containsOrIncl slug:
       let msg = &"There is more than one Practice Exercise with the slug {q slug}"
@@ -420,7 +420,7 @@ proc checkExerciseSlugsAndForegone(trackConfig: TrackConfig; b: var bool;
     let msg = &"There must be a Practice Exercise with the slug `hello-world`"
     b.setFalseAndPrint(msg, path)
 
-  for slug in trackConfig.exercises.foregone:
+  for slug in exercises.foregone:
     if slug in conceptExerciseSlugs or slug in practiceExerciseSlugs:
       let msg = &"The `exercises.foregone` array contains the slug {q slug}, " &
                  "but there is an implemented exercise with that slug"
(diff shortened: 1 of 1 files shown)


src/lint/track_config.nim
```diff
@@ -388,6 +389,43 @@ proc checkExercisesPCP(exercises: seq[ConceptExercise] | seq[PracticeExercise];
...
       b.setFalseAndPrint(msg, path)
 
+proc checkExerciseSlugsAndForegone(trackConfig: TrackConfig; b: var bool;
+                                   path: Path) =
+  ## Sets `b` to `false` if the below conditions are not satisfied:
+  ## - Each slug of a Concept Exercise or a Practice Exercise in `trackConfig`
+  ##   only exists once on the track.
+  ## - There is exactly one Practice Exercise with the slug `hello-world`.
+  ## - The `foregone` array does not contain a slug of an implemented exercise.
+  var conceptExerciseSlugs = initHashSet[string](200)
```
ErikSchierboom, line 399: Is the `200` just an initial size, or is it a maximum?

ErikSchierboom (approved):
Excellent

src/lint/track_config.nim
ee7: It's an initial size - how many items the `HashSet` should support. But, like Nim sequences, a `HashSet` "expands" automatically if you keep adding items.

---

In case you're interested, the implementation details for `HashSet` are here:
- [`lib/pure/collections/sets.nim`](https://github.com/nim-lang/Nim/blob/ee2eb5cae258/lib/pure/collections/sets.nim#L58-L67)
- [`lib/pure/collections/setimpl.nim`](https://github.com/nim-lang/Nim/blob/ee2eb5cae258/lib/pure/collections/setimpl.nim#L18-L24)
- [`lib/pure/collections/hashcommon.nim`](https://github.com/nim-lang/Nim/blob/ee2eb5cae258/lib/pure/collections/hashcommon.nim#L10-L34)

We allocate an initial number of "slots" using the formula `theIntegerThatYouPass * 3 div 2 + 4`, and in general the HashSet expands when it's more than two-thirds full.

And as a quick demonstration, let's use `initHashSet(3)` (8 slots, by the above calculation) and then insert 5 items,  print the representation, then add another item, and print the representation agin:

```Nim
import std/sets

var s = initHashSet[string](3)
s.incl "aaa"
s.incl "bbb"
s.incl "ccc"
s.incl "ddd"
s.incl "eee"
echo s.repr
echo "--------------"
# We rehash when `(t.dataLen * 2 < t.counter * 3) or (t.dataLen - t.counter < 4)`
# An `initialSize` of 3 produces a `dataLen` of 8 , so the above condition becomes true when `t.counter` is 5
# Basically, we rehash when the HashSet is more than two thirds full.
s.incl "fff"
echo s.repr
```

The output looks something like the below. We can see that:
- the `HashSet` grows as we insert the 6th item
- the resulting HashSet occupies double the memory (because `growthFactor == 2`)
- (if you stare hard enough) the pointer to the `HashSet` changes - we allocated a new block of memory

```
[data = 0x7fac8a41c050@[
  [Field0 = 735026264, Field1 = 0x7fac8a41d0c0"ccc"],
  [Field0 = 2446311049, Field1 = 0x7fac8a41d090"bbb"],
  [Field0 = 4095589471, Field1 = 0x7fac8a41d0f0"ddd"],
  [Field0 = 1505603105, Field1 = 0x7fac8a41d120"eee"],
  [Field0 = 0, Field1 = ""],
  [Field0 = 0, Field1 = ""],
  [Field0 = 0, Field1 = ""],
  [Field0 = 3033554871, Field1 = 0x7fac8a41d060"aaa"]],
counter = 5]

--------------

[data = 0x7fac8a41e050@[
  [Field0 = 966543936, Field1 = 0x7fac8a41dab0"fff"],
  [Field0 = 1505603105, Field1 = 0x7fac8a41da50"eee"],
  [Field0 = 0, Field1 = ""],
  [Field0 = 0, Field1 = ""],
  [Field0 = 0, Field1 = ""],
  [Field0 = 0, Field1 = ""],
  [Field0 = 0, Field1 = ""],
  [Field0 = 3033554871, Field1 = 0x7fac8a41da80"aaa"],
  [Field0 = 735026264, Field1 = 0x7fac8a41d9c0"ccc"],
  [Field0 = 2446311049, Field1 = 0x7fac8a41d9f0"bbb"],
  [Field0 = 0, Field1 = ""],
  [Field0 = 0, Field1 = ""],
  [Field0 = 0, Field1 = ""],
  [Field0 = 0, Field1 = ""],
  [Field0 = 0, Field1 = ""],
  [Field0 = 4095589471, Field1 = 0x7fac8a41da20"ddd"]],
counter = 6]
```

Needless allocations are an important source of performance problems, so it's a good optimization to avoid growing strings/seqs/HashSets etc when they're large and you're doing a lot of work. But of course, it doesn't matter at all in our case.

ErikSchierboom: Cool, thanks.

PR #418, opened 2021-09-07
.gitignore, config, nimble: use lock file

### LINKED ISSUE #467: build: use lockfiles
See https://github.com/exercism/configlet/commit/d6d728372477702e70b0c5c338a0b054a1ed134e.

ee7:
Before this commit, configlet pinned the version of each Nimble dependency by using the .nimble file. This was the best approach until recently, and was good enough for creating configlet releases, but it [wasn't robust][1] - Nimble wasn't designed to produce reproducible builds with that mechanism. For example, Nimble didn't check the hash of a package at build time to ensure that it was unmodified.

Now that we've [updated to Nim 2.0][2] and [vendored the parseopt3 dependency][3], let's use Nimble's new lock file mechanism. This should be robust.

Run `nimble lock` and `nimble setup`, and remove versions from the nimble file.

From the [docs for `nimble lock`][4]:

>The `nimble lock` command will generate or update a package lock file named `nimble.lock`. This file is used for pinning the exact versions of the dependencies of the package. The file is intended to be committed and used by other developers to ensure that exactly the same version of the dependencies is used by all developers.
>
>[...]
>
>If a lock file `nimble.lock`` exists, then on performing all Nimble commands which require searching for dependencies and downloading them in the case they are missing (like `build`, `install`, `develop`), it is read and its content is used to download the same version of the project dependencies by using the URL, download method and VCS revision written in it. The checksum of the downloaded package is compared against the one written in the lock file. In the case the two checksums are not equal then it will be printed error message and the operation will be aborted. Reverse dependencies are added for installed locked dependencies just like for any other package being locally installed.

and the [docs for `nimble setup`][5]:

>The `nimble setup` command creates a `nimble.paths` file containing file system paths to the dependencies. It also includes the paths file in the `config.nims` file (by creating it if it does not already exist) to make them available for the compiler. `nimble.paths` file is user-specific and MUST NOT be committed. The command also adds `nimble.develop` and `nimble.paths` files to the `.gitignore` file.`

Closes: #467

[1]: d6d728372477 ("build: pin versions of Nimble packages", 2021-11-25)
[2]: fa7d0bba04aa (".github, config, json, nimble: bump Nim from 1.6.12 to 2.0.0", 2023-08-08)
[3]: 7471af3add00 ("nimble, patches, cli: vendor parseopt3 dependency", 2023-08-08)
[4]: https://github.com/nim-lang/nimble/blob/412af022a441/readme.markdown#nimble-lock
[5]: https://github.com/nim-lang/nimble/blob/412af022a441/readme.markdown#nimble-setup

---

To-do:

- [x] Wait for a Nim release that adds support for the Nimble that supports lock files.
- [x] Either wait for Nim to **ship** with a Nimble that supports lock files, or manually install such a Nimble version in our CI workflows. Edit: Nim 2.0.0 should ship with a Nimble that supports lock files.
- [x] Bump package versions
- [x] Merge #723
- [x] Resolve problem with importing nimble packages when lock file present, when not using `nimble build`.
- [x] Consider if there's a better place to run `nimble setup`

Refs: d6d728372477702e70b0c5c338a0b054a1ed134e
Closes: #467

### DIFF
--- a/configlet.nimble
+++ b/configlet.nimble
@@ -17,9 +17,9 @@ bin           = @["configlet"]
 
 # Dependencies
 requires "nim >= 2.0.0"
-requires "jsony#ea811bec7fa50f5abd3088ba94cda74285e93f18"       # 1.1.5  (2023-02-09)
-requires "parsetoml#6e5e16179fa2db60f2f37d8b1af4128aaa9c8aaf"   # 0.7.1  (2023-08-06)
-requires "supersnappy#e4df8cb5468dd96fc5a4764028e20c8a3942f16a" # 2.1.3  (2022-06-12)
+requires "jsony"
+requires "parsetoml"
+requires "supersnappy"
 
 task test, "Runs the test suite":
   exec "nim r ./tests/all_tests.nim"
--- a/configlet.nimble
+++ b/configlet.nimble
@@ -22,4 +22,6 @@ requires "parsetoml"
 requires "supersnappy"
 
 task test, "Runs the test suite":
+  if not fileExists("nimble.paths"):
+    exec "nimble setup"
   exec "nim r ./tests/all_tests.nim"
--- a/.gitignore
+++ b/.gitignore
@@ -22,5 +22,7 @@
 # Ignore temporary repos
 tests/.test_elixir_track_repo/
 tests/.test_nim_track_repo/
+
+# Ignore user-specific Nimble files
 nimble.develop
 nimble.paths
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@
 # Ignore temporary repos
 tests/.test_elixir_track_repo/
 tests/.test_nim_track_repo/
+nimble.develop
+nimble.paths
--- a/config.nims
+++ b/config.nims
@@ -28,8 +28,9 @@ if defined(release):
     elif defined(clang):
       switch("clang.exe", "musl-clang")
       switch("clang.linkerexe", "musl-clang")
-# begin Nimble config (version 2)
+
+# Tell Nim the paths to Nimble packages. We need this because we ran `nimble lock`.
+# The below lines are added by `nimble setup`.
 --noNimblePath
 when withDir(thisDir(), system.fileExists("nimble.paths")):
   include "nimble.paths"
-# end Nimble config
--- a/config.nims
+++ b/config.nims
@@ -28,3 +28,8 @@ if defined(release):
     elif defined(clang):
       switch("clang.exe", "musl-clang")
       switch("clang.linkerexe", "musl-clang")
+# begin Nimble config (version 2)
+--noNimblePath
+when withDir(thisDir(), system.fileExists("nimble.paths")):
+  include "nimble.paths"
+# end Nimble config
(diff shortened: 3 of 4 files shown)


ErikSchierboom (approved):

ee7:
This looks promising: https://github.com/nim-lang/Nim/commit/908fc2a22e7336670173bcd06ab30f440ca6d321

ee7:
> This looks promising: [nim-lang/Nim@908fc2a](https://github.com/nim-lang/Nim/commit/908fc2a22e7336670173bcd06ab30f440ca6d321)

Nim 1.6.2 RC1 is now out, which is intended to support the Nimble that can use lockfiles. But it doesn't ship with it. See:

- https://forum.nim-lang.org/t/8710
- https://forum.nim-lang.org/t/8712

ee7:
Nim 2.0 should ship with a nimble version that supports lock files.

ee7:
With f3cfca9ae05f1eefc3b8c6bac52895deb2868dff, `nimble build` works, but `nimble test` doesn't. I feel like it should.

ee7:
The error here is unexpected to me. I'm not sure what I'm supposed to do.

To ask, I've opened [nim-lang/Nim#22438](https://www.github.com/nim-lang/Nim/issues/22438).

ee7:
Only 23 months for this one.

Please see:

- https://github.com/nim-lang/nimble#nimble-lock
- https://github.com/nim-lang/nimble#nimble-setup
- [nim-lang/nimble#1004](https://www.github.com/nim-lang/nimble/issues/1004)

configlet.nimble
```diff
@@ -17,9 +17,11 @@ bin           = @["configlet"]
...
-requires "jsony#ea811bec7fa50f5abd3088ba94cda74285e93f18"       # 1.1.5  (2023-02-09)
-requires "parsetoml#6e5e16179fa2db60f2f37d8b1af4128aaa9c8aaf"   # 0.7.1  (2023-08-06)
-requires "supersnappy#e4df8cb5468dd96fc5a4764028e20c8a3942f16a" # 2.1.3  (2022-06-12)
+requires "jsony"
+requires "parsetoml"
+requires "supersnappy"
 
 task test, "Runs the test suite":
+  if not fileExists("nimble.paths"):
+    exec "nimble setup"
```
ee7, line 26: This is enough to make the tests workflow (and therefore CI for this PR) pass.

But it might not be enough to make the build workflow pass, if the build workflow doesn't call `nimble test` before executing a `nim` command. I'll make sure that works before we merge this PR.

ee7: The `build.yml` workflow [succeeded](https://github.com/exercism/configlet/actions/runs/5848264602) for commit https://github.com/exercism/configlet/commit/adb465d513efd960d00539cc298fc1712e83f687.

So this change is OK for CI. Just be aware that you'll get an error locally if you run some `nim` command on a file in the project without either:

- running `nimble test` after this PR is merged
- or running `nimble setup` manually

ErikSchierboom: Okay I can live with that 🙂

PR #420, opened 2021-09-08
lint(track_config): tidy up second pass

ee7:
This PR:
- adds some comments that try to explain the approach for the second linting pass
- tries to improve an error message that a user shouldn't encounter.

Reproducing the latter commit message below.

---

Before this commit, an error during jsony's deserialization would look
something like:

```
jsony.nim(46)            parseHook
Error: unhandled exception: Expected " but got 1 instead. At offset: 73830 [JsonError]
```

With this commit, it instead looks like:

```
  Error: JSON parsing error during the second linting pass:
  config.json(3156, 10): Expected " but got 1 instead.
```

That is, we:
- Handle the exception
- Better explain what happened
- Mention the JSON file being parsed, rather than `jsony.nim`
- Provide the line and column number, rather than the offset, where the
  parsing error occurred

However, we hope that a user shouldn't encounter these errors, which
should only occur when a value's type in the `config.json` file doesn't
match that in the `TrackConfig` object. The current approach tries to
validate the types in the first pass, and only perform the second pass
when the types are correct.

Currently `showError` also prints the help message. We'll add a better
error proc in the future (maybe in `helpers.nim`, not `cli.nim`).

### DIFF
--- a/src/lint/track_config.nim
+++ b/src/lint/track_config.nim
@@ -426,8 +426,8 @@ proc checkExerciseSlugsAndForegone(exercises: Exercises; b: var bool;
                  "but there is an implemented exercise with that slug"
       b.setFalseAndPrint(msg, path)
 
-proc satisfiesSecondPass(s: string; path: Path): bool =
-  let trackConfig = fromJson(s, TrackConfig)
+proc satisfiesSecondPass(trackConfigContents: string; path: Path): bool =
+  let trackConfig = fromJson(trackConfigContents, TrackConfig)
   result = true
 
   let conceptSlugs = getConceptSlugs(trackConfig)
--- a/src/lint/track_config.nim
+++ b/src/lint/track_config.nim
@@ -205,6 +205,8 @@ type
     sActive = "active"
     sDeprecated = "deprecated"
 
+  # We can use a `HashSet` for `concepts`, `prerequisites` and `practices`
+  # because the first pass has already checked that each has unique values.
   ConceptExercise = object
     slug: string
     # name: string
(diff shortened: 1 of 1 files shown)


src/lint/track_config.nim
```diff
@@ -238,6 +240,36 @@ type
...
+      result.col = 0
+    inc result.col
+
+proc toTrackConfig(trackConfigContents: string): TrackConfig =
+  ## Deserializes `trackConfigContents` using `jsony` to a `TrackConfig` object.
+  try:
+    result = fromJson(trackConfigContents, TrackConfig)
+  except jsony.JsonError:
+    let jsonyMsg = getCurrentExceptionMsg()
+    var msg = "JSON parsing error during the second linting pass:\nconfig.json"
```
ErikSchierboom, line 260: I would argue that configlet users don't have to know about first/second passes, but just about errors:

```suggestion
    var msg = "JSON parsing error:\nconfig.json"
```

ErikSchierboom (approved):
I like the approach, just one tiny nit on the message.

ErikSchierboom:
👍

src/lint/track_config.nim
ee7: I wrote it like this to make it more distinct from the other JSON parsing errors, on the basis that:
1. This error indicates a bug in `configlet`, rather than a problem with `config.json`, because type errors in `config.json` should be caught by the first pass.
2. We build the release binary with `-d:release`, so there's no stacktrace for errors

But if a user opens an issue for the error message, I think we'll be able to detect that it happens during the second pass anyway. So I've pushed a commit with your suggestion (on top of some refactoring).

But now that I think about it, given point 1 above, I'm leaning towards adding something like:

> `If you see this error message, please open an issue in https://github.com/exercism/configlet, including the contents of the track config.json file  at the time configlet lint ran`

What do you think?

ErikSchierboom: I like that suggestion. It should indeed not happen

ee7: Pushed.

Visible enough? :)

PR #423, opened 2021-09-10
lint(track_config): refactor slightly

ee7:
Improve proc names and type signatures.

### DIFF
--- a/src/lint/track_config.nim
+++ b/src/lint/track_config.nim
@@ -580,17 +580,22 @@ proc satisfiesSecondPass(trackConfigContents: string; path: Path): bool =
   let trackConfig = toTrackConfig(trackConfigContents)
   result = true
 
-  let conceptSlugs = getConceptSlugs(trackConfig.concepts)
-  checkPractices(trackConfig.exercises.practice, conceptSlugs, result, path)
-  let conceptsTaught = checkExerciseConcepts(trackConfig.exercises.`concept`,
-                                             conceptSlugs, result, path)
-  checkPrerequisites(trackConfig.exercises.`concept`, conceptSlugs,
-                     conceptsTaught, result, path)
-  checkPrerequisites(trackConfig.exercises.practice, conceptSlugs,
-                     conceptsTaught, result, path)
-  checkExercisesPCP(trackConfig.exercises.`concept`, result, path)
-  checkExercisesPCP(trackConfig.exercises.practice, result, path)
-  checkExerciseSlugsAndForegone(trackConfig.exercises, result, path)
+  let exercises = trackConfig.exercises
+  let conceptExercises = exercises.`concept`
+  let practiceExercises = exercises.practice
+  let concepts = trackConfig.concepts
+
+  let conceptSlugs = getConceptSlugs(concepts)
+  checkPractices(practiceExercises, conceptSlugs, result, path)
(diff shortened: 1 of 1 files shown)


src/lint/track_config.nim
```diff
@@ -579,19 +580,24 @@ proc satisfiesSecondPass(trackConfigContents: string; path: Path): bool =
...
+                                             result, path)
+  checkPrerequisites(conceptExercises, conceptSlugs, conceptsTaught, result,
+                     path)
+  checkPrerequisites(practiceExercises, conceptSlugs, conceptsTaught, result,
+                     path)
+  checkExercisesPCP(conceptExercises, result, path)
+  checkExercisesPCP(practiceExercises, result, path)
+  checkExerciseSlugsAndForegone(exercises, result, path)
+
+proc satisfiesFirstPass(data: JsonNode; path: Path): bool =
```
ErikSchierboom, line 600: I'm not sure about this name. It explains how something is happening, but not why. I think we want that indicates that the type of the JSON object is checked.

ErikSchierboom (approved):
Looking good! One tiny naming nit.

src/lint/track_config.nim
```diff
@@ -280,19 +302,19 @@ proc tidyJsonyErrorMsg(trackConfigContents: string): string =
...
     let msg = tidyJsonyErrorMsg(trackConfigContents)
     showError(msg)
 
-func getConceptSlugs(trackConfig: TrackConfig): HashSet[string] =
+func getConceptSlugs(concepts: Concepts): HashSet[string] =
   ## Returns a set of every `slug` in the top-level `concepts` array of a track
   ## `config.json` file.
   result = initHashSet[string]()
-  for con in trackConfig.concepts:
+  for con in concepts:
```
ee7, line 317: @ErikSchierboom: FYI, in Nim 1.6 we can import `std/sugar` and instead write (with zero overhead!):

```Nim
collect(for con in concepts: {con.slug})
```

or

```Nim
collect:
  for con in concepts:
    {con.slug}
```


So after the Nim 1.6 release I'll probably write this kind of proc as a `collect` one-liner.

---

In Nim 1.4 we have to be more verbose:

```Nim
collect(initHashSet, for con in concepts: {con.slug})
```

or

```Nim
collect(initHashSet):
  for con in concepts:
    {con.slug}
```

---

See also:
- [Nim `collect` compared with Python list comprehensions](https://github.com/nim-lang/Nim/wiki/Nim-for-Python-Programmers#List-Comprehensions)
- `std/sugar` [docs](https://nim-lang.github.io/Nim/sugar.html#collect.m%2Cuntyped)
- [Nim 1.2.0 release notes](https://nim-lang.org/blog/2020/04/03/version-120-released.html)

src/lint/track_config.nim
ee7: I don't like the name either. But the problem with something like `checkTypes` is that our first pass does a lot more than check the types - it's where most of our current linting logic is.

I didn't find a better name, so let's see if we can find a better name later. `satisfiesFirstPassThatIncludesATypeCheck` is a bit verbose :)

In the meantime, I've added a comment to help a little.

ee7:
Force-pushed a no-op to restart the macOS job. I still have no idea why it takes so long, especially in the "Install Nim" phase with no log output.

I clicked the "Cancel workflow" button, aiming to restart it, but it did nothing

PR #425, opened 2021-09-13
lint(track_config): check for `concept`-`prerequisite` cycles in `exercises.concept[]`

jmrunkle:
Full disclosure: I have never written anything in nim before, I just used documentation to figure out the syntax.

It also looks like most of this is lacking tests which is unfortunate. I am happy to take feedback since I am basically just hacking away over here.


jmrunkle:
I was finally able to actually test this against the go track to verify that there are no cycles (which I have already done manually). It works as expected. Also tested by introducing a real cycle just to make sure it works.

Here is an example error:

> The Concept Exercise `blackjack` has a cycle in its `prerequisites`: @["conditionals-switch", "numbers", "nil", "maps", "numbers"]:

For that case, conditionals-switch is the concept taught by blackjack, it depends on numbers which depends on nil which depends on maps which it turn depends on numbers creating a cycle. I could try to make that error message more readable I suppose, but this works for a first-pass at catching these cyclic issues.

jmrunkle:
Improved the error message:

> The Concept Exercise `logs-logs-logs` has a cycle in its `prerequisites`: `strings-package` depends on `slices`, which depends on `range-iteration`, which depends on `slices` forming a cycle

ee7:
(I'll fix the merge conflict)

ee7:
> (I'll fix the merge conflict)

See below for the "diff between the diffs" (that is, the diff between `git diff 39fd168 c618c0a` and `git diff 8121d24 7ba1e1a`).

```diff
5c5
< @@ -368,13 +368,40 @@ proc checkExerciseConcepts(trackConfig: TrackConfig;
---
> @@ -393,13 +393,40 @@ proc checkExerciseConcepts(conceptExercises: seq[ConceptExercise];
25c25
< +  
---
> +
30,32c30,32
<  proc checkConceptExercisePrerequisites(trackConfig: TrackConfig;
<                                         conceptSlugs, conceptsTaught: HashSet[string];
<                                         b: var bool; path: Path) =
---
>  proc checkPrerequisites(conceptExercises: seq[ConceptExercise];
>                          conceptSlugs, conceptsTaught: HashSet[string];
>                          b: var bool; path: Path) =
35,36c35,36
<    ## `trackConfig`, and sets `b` to `false` if a check fails.
<    for conceptExercise in visibleConceptExercises(trackConfig):
---
>    ## `conceptExercises`, and sets `b` to `false` if a check fails.
>    for conceptExercise in visible(conceptExercises):
45,47c45,47
<                    &"{q preReq} in both its `prerequisites` and its `concepts`"
< @@ -390,6 +417,17 @@ proc checkConceptExercisePrerequisites(trackConfig: TrackConfig;
<                    &"{q preReq} in its `prerequisites`, which is not a " &
---
>                    &"{q prereq} in both its `prerequisites` and its `concepts`"
> @@ -415,6 +442,17 @@ proc checkPrerequisites(conceptExercises: seq[ConceptExercise];
>                    &"{q prereq} in its `prerequisites`, which is not a " &
52c52
< +  for conceptExercise in visibleConceptExercises(trackConfig):
---
> +  for conceptExercise in visible(conceptExercises):
```

jmrunkle:
> See below for the "diff between the diffs" (that is, the diff between `git diff 39fd168 c618c0a` and `git diff 8121d24 7ba1e1a`).

Just double-checking: is there an action you would like me to take on this?

ee7:
> @ee7 - Just double-checking: is there an action you would like me to take on this?

Sorry - I should've written something. No, it's okay - just documenting what I did, since I rudely updated your branch.

ee7:
Many thanks for working on this. I think it's one of the more useful to-be-implemented linting rules.

I made a nitpick-only PR to your branch here: https://github.com/jmrunkle/configlet/pull/1#pullrequestreview-753305287 (I don't like the GitHub "Suggested changes" UI - sorry).

I didn't look at this PR properly yet, but it looks promising. 

So far I can say:
1. This PR doesn't currently change the output of `configlet lint` on any track
1. The `configlet lint` output becomes very long if an "earlier" exercise has a cycle. Could we limit it? 

> It also looks like most of this is lacking tests which is unfortunate. 

Indeed. I'm acutely aware of the lack of `configlet lint` tests.

jmrunkle:
> The `configlet lint` output becomes very long if an "earlier" exercise has a cycle (for the extreme case, see here). Could we limit it?

It is probably possible to limit it, but probably not in a way that will be clean in the code. Since this is very likely to be exactly one cycle now that we have this detection in place, I think the duplicates are fine. You will most likely just fix the first one rather than trying to do all of them and it will almost certainly be the only cycle.

> What was the difficulty? 

Figuring out how to get nim installed in the Gitpod setup and have all of the various tools it needs setup in the right places (nim, nimble, and nimpretty in particular). I got the basic nim command setup and it was figuring out that I needed nimble as well that proved tricky.

jmrunkle:
Actually, I just thought of a clean-ish way that would at least limit it to one per exercise rather than one per exercise/prereq combo. That seems like the most appropriate level to me. It still means that an early cycle (eg. basics) would cause every exercise to get one print out, but having this be one for each exercise seems like a dramatic improvement. Let me give that a shot.

jmrunkle:
> Actually, I just thought of a clean-ish way that would at least limit it to one per exercise rather than one per exercise/prereq combo. That seems like the most appropriate level to me. It still means that an early cycle (eg. basics) would cause every exercise to get one print out, but having this be one for each exercise seems like a dramatic improvement. Let me give that a shot.

Done.

ErikSchierboom:
This is looking great! Thanks for working on this. I'll let @ee7 review the latest code.

jmrunkle:
@ee7 - anything else you would like me to address?

ee7:
> @ee7 - anything else you would like me to address?

Nope, thanks. I'll push a few more trivial commits directly to this PR's branch, then I'll merge it and create a release.

I have a few ideas for some minor improvements, but they aren't blocking, and I think this PR gets us nearly all of the way there in terms of real-world usefulness.

src/lint/track_config.nim
```diff
@@ -393,13 +393,54 @@ proc checkExerciseConcepts(conceptExercises: seq[ConceptExercise];
...
+  ## - exercise slug A teaches concept A', and requires concept B'
+  ## - exercise slug B teaches concept B', and requires concept C'
+  ## - exercise slug C teaches concept C', and requires concept A'
+  ##
+  ## This is important to forbid because any exercise that is involved in such a
+  ## cycle would not be unlockable.
+  if hadCycle:
+    return
+
+  let updatedPrereqPath = prereqPath & @[currentConcept]
```
ee7, line 414: Relatively costly allocations here - I think we can refactor later by making `prereqPath` a `var seq`.

jmrunkle: I think you need a copy for the recursion to work properly otherwise you'll keep appending to it (improperly). The other option is to remove things at various spots in the recursion. Either way, I think this can be deferred for later optimization.

src/lint/track_config.nim
```diff
@@ -393,13 +393,54 @@ proc checkExerciseConcepts(conceptExercises: seq[ConceptExercise];
...
+  if currentConcept in prereqPath:
+    var formattedCycle = &"{q updatedPrereqPath[0]} depends on {q updatedPrereqPath[1]}"
```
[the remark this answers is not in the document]
jmrunkle: Correct - it certainly has at least two.

ee7 (approved):

PR #426, opened 2021-09-13
lint(track_config): check exercise directories are slugs

ee7:
This PR currently causes the below diff to the output of `configlet lint`, per track:

#### clojure
```diff
+`./exercises/concept` contains a directory named `international-calling-connoisseur`, which is not a `slug` in the array of concept exercises:
+./config.json
+
+`./exercises/concept` contains a directory named `squeaky-clean`, which is not a `slug` in the array of concept exercises:
+./config.json
+
+Configlet detected at least one problem.
+For more information on resolving the problems, please see the documentation:
+https://github.com/exercism/docs/blob/main/building/configlet/lint.md

```

#### go
```diff
+`./exercises/concept` contains a directory named `the-farm`, which is not a `slug` in the array of concept exercises:
+./config.json
+
+Configlet detected at least one problem.
+For more information on resolving the problems, please see the documentation:
+https://github.com/exercism/docs/blob/main/building/configlet/lint.md

```

#### javascript
```diff
+`./exercises/concept` contains a directory named `elyses-transformative-enchantments`, which is not a `slug` in the array of concept exercises:
+./config.json
+
+Configlet detected at least one problem.
+For more information on resolving the problems, please see the documentation:
+https://github.com/exercism/docs/blob/main/building/configlet/lint.md

```

#### php
```diff
+`./exercises/practice` contains a directory named `darts`, which is not a `slug` in the array of practice exercises:
+./config.json
+
+`./exercises/practice` contains a directory named `mask-credit-card`, which is not a `slug` in the array of practice exercises:
+./config.json
+
+`./exercises/practice` contains a directory named `ordinal-number`, which is not a `slug` in the array of practice exercises:
+./config.json
+
+`./exercises/practice` contains a directory named `scale-generator`, which is not a `slug` in the array of practice exercises:
+./config.json
+
+Configlet detected at least one problem.
+For more information on resolving the problems, please see the documentation:
+https://github.com/exercism/docs/blob/main/building/configlet/lint.md

```

#### vbnet
```diff
+`./exercises/practice` contains a directory named `matching-brackets`, which is not a `slug` in the array of practice exercises:
+./config.json
+
+`./exercises/practice` contains a directory named `perfect-numbers`, which is not a `slug` in the array of practice exercises:
+./config.json
+
+Configlet detected at least one problem.
+For more information on resolving the problems, please see the documentation:
+https://github.com/exercism/docs/blob/main/building/configlet/lint.md

```

#### wren
```diff
+`./exercises/practice` contains a directory named `sieve`, which is not a `slug` in the array of practice exercises:
+./config.json
+
+Configlet detected at least one problem.
+For more information on resolving the problems, please see the documentation:
+https://github.com/exercism/docs/blob/main/building/configlet/lint.md

```

#### z3
```diff
+`./exercises/practice` contains a directory named `propositional-logic`, which is not a `slug` in the array of practice exercises:
+./config.json
+
+Configlet detected at least one problem.
+For more information on resolving the problems, please see the documentation:
+https://github.com/exercism/docs/blob/main/building/configlet/lint.md

```

### DIFF
--- a/src/lint/track_config.nim
+++ b/src/lint/track_config.nim
@@ -654,8 +654,8 @@ proc checkExerciseDirsAreInTrackConfig(trackDir: Path; data: JsonNode;
           if dirSlug notin exerciseSlugs:
             let msg = &"{q $exercisesDir} contains a directory named {q dirSlug}, " &
                       &"which is not a `slug` in the array of {exerciseKind} " &
-                       "exercises. Please add an entry for this exercise. If " &
-                       "the exercise is not ready to be shown on the " &
+                       "exercises. Please add the exercise to that array. " &
+                       "If the exercise is not ready to be shown on the " &
                        "website, please set its `status` value to \"wip\""
             b.setFalseAndPrint(msg, path)
 
--- a/src/lint/track_config.nim
+++ b/src/lint/track_config.nim
@@ -623,20 +623,22 @@ proc satisfiesSecondPass(trackConfigContents: string; path: Path): bool =
 
 proc getExerciseSlugs(data: JsonNode; k: string): HashSet[string] =
   result = initHashSet[string]()
-  if data.kind == JObject:
-    if data.hasKey("exercises"):
-      let exercises = data["exercises"]
-      if exercises.kind == JObject:
-        if exercises.hasKey(k):
-          if exercises[k].kind == JArray:
-            for exercise in exercises[k]:
-              if exercise.kind == JObject:
-                if exercise.hasKey("slug"):
-                  let slug = exercise["slug"]
-                  if slug.kind == JString:
-                    let slugStr = slug.getStr()
-                    if slugStr.len > 0:
-                      result.incl slugStr
+  if data.kind != JObject or "exercises" notin data:
+    return
+
+  let exercises = data["exercises"]
+
+  if exercises.kind != JObject or k notin exercises or exercises[k].kind != JArray:
+    return
+
+  for exercise in exercises[k]:
+    if exercise.kind == JObject:
+      if exercise.hasKey("slug"):
+        let slug = exercise["slug"]
+        if slug.kind == JString:
+          let slugStr = slug.getStr()
+          if slugStr.len > 0:
+            result.incl slugStr
 
 proc checkExerciseDirsAreInTrackConfig(trackDir: Path; data: JsonNode;
                                        b: var bool; path: Path) =
--- a/src/lint/track_config.nim
+++ b/src/lint/track_config.nim
@@ -1,4 +1,4 @@
-import std/[json, sets, strformat, strscans, strutils, tables]
+import std/[json, os, sets, strformat, strscans, strutils, tables]
 import pkg/jsony
 import ".."/[cli, helpers]
 import "."/validators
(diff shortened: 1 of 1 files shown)


src/lint/track_config.nim
```diff
@@ -621,6 +621,39 @@ proc satisfiesSecondPass(trackConfigContents: string; path: Path): bool =
...
+        if exercises.hasKey(k):
+          if exercises[k].kind == JArray:
+            for exercise in exercises[k]:
+              if exercise.kind == JObject:
+                if exercise.hasKey("slug"):
+                  let slug = exercise["slug"]
+                  if slug.kind == JString:
+                    let slugStr = slug.getStr()
+                    if slugStr.len > 0:
+                      result.incl slugStr
```
ee7, line 639: Ugly.

src/lint/track_config.nim
```diff
@@ -621,6 +621,39 @@ proc satisfiesSecondPass(trackConfigContents: string; path: Path): bool =
...
```
ErikSchierboom, line 639: So.... I take you're not a fan of early returns? :)

src/lint/track_config.nim
```diff
@@ -621,6 +621,39 @@ proc satisfiesSecondPass(trackConfigContents: string; path: Path): bool =
...
+  for exerciseKind in ["concept", "practice"]:
+    let exerciseSlugs = getExerciseSlugs(data, exerciseKind)
+    if exerciseSlugs.len > 0:
+      let exercisesDir = trackDir / "exercises" / exerciseKind
+      if dirExists(exercisesDir):
+        for exerciseDir in getSortedSubdirs(exercisesDir):
+          let dirSlug = lastPathPart(exerciseDir.string)
+          if dirSlug notin exerciseSlugs:
+            let msg = &"{q $exercisesDir} contains a directory named {q dirSlug}, " &
+                      &"which is not a `slug` in the array of {exerciseKind} exercises"
```
ErikSchierboom, line 654: I'm wondering if perhaps we should let people know that if the exercise is not yet done, they can add an entry to the `config.json` with the status set to `wip`, to prevent the exercise from showing up on the website.

ErikSchierboom (approved):

src/lint/track_config.nim
ee7: Good idea. Pushed a commit - any complains about the wording?

src/lint/track_config.nim
```diff
@@ -621,6 +621,42 @@ proc satisfiesSecondPass(trackConfigContents: string; path: Path): bool =
...
+    let exerciseSlugs = getExerciseSlugs(data, exerciseKind)
+    if exerciseSlugs.len > 0:
+      let exercisesDir = trackDir / "exercises" / exerciseKind
+      if dirExists(exercisesDir):
+        for exerciseDir in getSortedSubdirs(exercisesDir):
+          let dirSlug = lastPathPart(exerciseDir.string)
+          if dirSlug notin exerciseSlugs:
+            let msg = &"{q $exercisesDir} contains a directory named {q dirSlug}, " &
+                      &"which is not a `slug` in the array of {exerciseKind} " &
+                       "exercises. Please add an entry for this exercise. If " &
```
ErikSchierboom, line 655: Maybe mention where the entry should be added (the config.json file)?

```suggestion
                       "exercises. Please add an entry for this exercise. If " &
```

ErikSchierboom (commented):
One tiny nit.

src/lint/track_config.nim
ee7: Mmm, it depends. This was a silly block of code, but here I think it's ugly regardless of how it's written.

I've changed to the below:
```Nim
proc getExerciseSlugs(data: JsonNode; k: string): HashSet[string] =
  result = initHashSet[string]()
  if data.kind != JObject or "exercises" notin data:
    return

  let exercises = data["exercises"]

  if exercises.kind != JObject or k notin exercises or exercises[k].kind != JArray:
    return

  for exercise in exercises[k]:
    if exercise.kind == JObject:
      if exercise.hasKey("slug"):
        let slug = exercise["slug"]
        if slug.kind == JString:
          let slugStr = slug.getStr()
          if slugStr.len > 0:
            result.incl slugStr
```

But I claim that it isn't a strict improvement. The indentation reflected the underlying design issue, and early `return` conceals it, making it less likely to be refactored in the future :)

Background: I wanted to check the directories even if the exercise part of the `config.json` is invalid - so with the current deserialization design, and without doing some refactoring, or a big `try`, we're stuck with validating the types a second time.


The conditions would be much more readable using some helper procs named e.g. `hasString`, but those procs we've implemented also write an error message if the condition is false. And I didn't want to print some duplicate error messages, or refactor those procs to take an extra parameter, so...

src/lint/track_config.nim
ee7: Tweaked it a bit. I didn't mention "the track-level `config.json` file" because our current error messages always print it at the end, like:

> \`./exercises/concept\` contains a directory named \`the-farm\`, which is not a \`slug\` in the array of concept exercises. Please add the exercise to that array. If the exercise is not ready to be shown on the website, please set its \`status\` value to "wip":
./config.json

src/lint/track_config.nim
```diff
@@ -621,6 +621,44 @@ proc satisfiesSecondPass(trackConfigContents: string; path: Path): bool =
...
+      let exercisesDir = trackDir / "exercises" / exerciseKind
+      if dirExists(exercisesDir):
+        for exerciseDir in getSortedSubdirs(exercisesDir):
+          let dirSlug = lastPathPart(exerciseDir.string)
+          if dirSlug notin exerciseSlugs:
+            let msg = &"{q $exercisesDir} contains a directory named {q dirSlug}, " &
+                      &"which is not a `slug` in the array of {exerciseKind} " &
+                       "exercises. Please add the exercise to that array. " &
+                       "If the exercise is not ready to be shown on the " &
+                       "website, please set its `status` value to \"wip\""
```
ee7, line 659: Comment: there's an ongoing question about how to refer to JSON key/value pairs. Here it feels painful to write `wip` with backticks like:

> please set its \`status\` value to \`wip\`

rather than with double quotes:

> please set its \`status\` value to \"wip\"

But that raises the question about whether we should write \"status\" with double quotes too (which sometimes feels worse in other error messages).

Issue #249, opened 2021-03-28
lint: implement remaining linting rules


ee7:
This issue will try to track which linting rules are implemented, and to what extent.

The list below is from [the spec](https://github.com/exercism/docs/blob/main/building/configlet/lint.md). If we change the spec then we should remember to make the same changes in the below.

---

### Rule: required files are present

The linter should check if all the required files are present. The non-exercise specific files that must be present are:

- `config.json`

The Concept Exercise specific files that must be present are:

- `exercises/concept/<slug>/.docs/hints.md`
- `exercises/concept/<slug>/.docs/instructions.md`
- `exercises/concept/<slug>/.docs/introduction.md`
- `exercises/concept/<slug>/.meta/config.json`

There will be a similar list for Practice Exercises, but we've not yet defined the spec for that.

Each concept listed in the `config.json` should have the following files:

- `concepts/<slug>/about.md`
- `concepts/<slug>/introduction.md`
- `concepts/<slug>/links.json`

### Rule: config.json file is valid

The `config.json` file should have the following checks:

- The file must be valid JSON
- The `"language"` key is required
- The `"language"` value must be a non-empty, non-blank string
- The `"slug"` key is required
- The `"slug"` value must be a non-empty, non-blank, lowercased string using kebab-case
- The `"active"` key is required
- The `"active"` value must be a boolean
- The `"blurb"` key is required
- The `"blurb"` value must be a non-empty, non-blank string with length <= 400
- The `"version"` key is required
- The `"version"` value must be the integer `3`
- The `"status.concept_exercises"` key is required
- The `"status.concept_exercises"` value must be a boolean
- The `"status.test_runner"` key is required
- The `"status.test_runner"` value must be a boolean
- The `"status.representer"` key is required
- The `"status.representer"` value must be a boolean
- The `"status.analyzer"` key is required
- The `"status.analyzer"` value must be a boolean
- The `"online_editor.indent_style"` key is required
- The `"online_editor.indent_style"` value must be the string `space` or `tab`
- The `"online_editor.indent_size"` key is required
- The `"online_editor.indent_size"` value must be an integer >= 0 and <= 8
- The `"exercises"` key is required
- The `"exercises.concept"` key is required
- The `"exercises.concept"` value must be an array
- The `"exercises.concept[].slug"` key is required
- The `"exercises.concept[].slug"` value must be a non-empty, non-blank, lowercased string using kebab-case
[...]

PR #435, opened 2021-10-05
Implement the config.json files section linting.

bobtfish:
Addresses some of #249 

Implements the following rules:

-  The "files" key is optional
-  The "files" value must be an object
-  The "files.solution" key is optional
-  The "files.solution" value must be an array
-  The "files.solution" values must be valid patterns⁴
-  The "files.solution" values must not have duplicates
-  The "files.test" key is optional
-  The "files.test" value must be an array
-  The "files.test" values must be valid patterns⁴
-  The "files.test" values must not have duplicates
-  The "files.example" key is optional
-  The "files.example" value must be an array
-  The "files.example" values must be valid patterns⁴
-  The "files.example" values must not have duplicates
-  The "files.exemplar" key is optional
-  The "files.exemplar" value must be an array
-  The "files.exemplar" values must be valid patterns⁴
-  The "files.exemplar" values must not have duplicates
-  The "files.editor" key is optional
-  The "files.editor" value must be an array
-  The "files.editor" values must be valid patterns⁴
-  The "files.editor" values must not have duplicates
-  Patterns can only be listed in either the "files.solution", "files.test", "files.example, "files.exemplar or "files.editor array (no overlap)

### DIFF
--- a/src/lint/validators.nim
+++ b/src/lint/validators.nim
@@ -176,10 +176,38 @@ func isUuidV4*(s: string): bool =
 
 var seenUuids = initHashSet[string](250)
 
+proc extractPlaceholder(value: string): string =
+  var i = 0
+  var phStart = -1
+  var ph = ""
+  while i < value.len:
+    var c = value[i]
+    if phStart == -1:
+      if c == '%' and value[i+1] == '{':
+        phStart = i+2
+        inc i
+    else:
+      if c == '}':
+        return ph
+      else:
+        ph = ph & c
+    inc i
+  return ""
+
+func isFilesPattern*(s: string): bool =
+  if not isEmptyOrWhitespace(s):
+    let ph = extractPlaceholder(s)
+    if ph == "" or ph == "kebab_slug" or ph == "snake_slug" or ph ==
+        "camel_slug" or ph == "pascal_slug":
+      result = true
+
+var seenFilePatterns = initHashSet[string](250)
+
 proc isString*(data: JsonNode; key: string; path: Path; context: string;
                isRequired = true; allowed = emptySetOfStrings;
                checkIsUrlLike = false; maxLen = int.high; checkIsKebab = false;
-               checkIsUuid = false; isInArray = false): bool =
+               checkIsUuid = false; isInArray = false;
+               checkIsFilesPattern = false): bool =
   result = true
   case data.kind
   of JString:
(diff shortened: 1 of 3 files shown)


src/lint/validators.nim
```diff
@@ -176,10 +176,38 @@ func isUuidV4*(s: string): bool =
...
+      if c == '}':
+        return ph
+      else:
+        ph = ph & c
+    inc i
+  return ""
+
+func isFilesPattern*(s: string): bool =
+  if not isEmptyOrWhitespace(s):
+    let ph = extractPlaceholder(s)
```
ErikSchierboom, line 199: Theoretically speaking, people could be using multiple placeholders in the files pattern. It is not likely, but it is something to consider supporting.

src/lint/validators.nim
```diff
@@ -176,10 +176,38 @@ func isUuidV4*(s: string): bool =
...
+    if ph == "" or ph == "kebab_slug" or ph == "snake_slug" or ph ==
+        "camel_slug" or ph == "pascal_slug":
```
ErikSchierboom, line 201: This looks like a good candidate for a `HashSet` to me.

tests/test_lint.nim
```diff
@@ -1,6 +1,26 @@
 import std/unittest
 import "."/lint/validators
 
+proc testIsFilesPattern =
+  suite "isFilesPattern":
```
ErikSchierboom, line 5: Could you also add test cases for file pattters with slashes? As in: sub-directories.

ErikSchierboom (commented):
Great! I'll leave the Nim code review to @ee7 .

src/lint/validators.nim
ErikSchierboom: Well, there could possibly be a path like: `%{pascal_slug}/%{snake_slug}.vb` or whatever. I've just checked all the track's `.files` properties, and there _is_ one track (Swift) that uses: `"Tests/%{pascal_slug}Tests/%{pascal_slug}Tests.swift"`

So we do have to support multiple placeholders.

> Also what about error cases? Presumably it's an error to add a placeholder which doesn't exist, so %{foobar} is an error

That is correct. That would be an error.

> [...] but how about incomplete placeholders? Is %{foobar an error, or valid?

Maybe it can be a warning?

ErikSchierboom:
@bobtfish Great!

bobtfish:
hey @ee7 thanks for brushing up my code.

Are you going to take this over the line, or would you like me to add the warning? (I think you added the extra tests already, right?)

src/lint/validators.nim
ee7: Added a low-tech warning in https://github.com/exercism/configlet/pull/435/commits/60e26ada55557742ea5d275221e7bef191b11813.

I might refactor later - e.g. `isFilesPattern` could be renamed and return an enum like `fpOk`, `fpIsBlank`, `fpHasInvalidPlaceholder`, `fpContainsForbiddenCharacter`.

PR #436, opened 2021-10-05
lint(track_config): check hello-world `status`

bobtfish:
The "exercises.practice[].status" value must, if "exercises.practice[].slug" is equal to hello-world, be either omitted or the string active

Refs: #249


ErikSchierboom:
Nice.

PR #437, opened 2021-10-12
lint(track_config): check concept directory names are in config

ErikSchierboom:
With this PR, `configlet lint` now produces an error if there is a subdirectory in the `concepts` directory of a track that does not have a corresponding entry in the track-level `config.json` file.

The implementation is similar to the exercise-based check.

See https://github.com/exercism/rust/pull/1363#issue-1022516101 CC @coriolinus


ErikSchierboom:
Done. Thanks for the review.

PR #462, opened 2021-10-28
Add configlet action

wolf99:
Per https://github.com/exercism/github-actions/issues/29

In draft until there is a decision on that issue about how to propagate the change to tracks.


README.md
```diff
@@ -99,20 +99,57 @@ Each concept exercise and concept have an `introduction.md` file. If you want th
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Fetch configlet
+      uses: exercism/github-actions/configlet-ci@main
+
+    - name: Configlet linter
+      uses: configlet lint
```
SaschaMann, line 148: This workflow is untested.

```suggestion
name: Configlet

permissions:
  contents: read

on:
  workflow_dispatch:
  push:
    branches:
      - main
  pull_request:

jobs:
  configlet:
    runs-on: ubuntu-latest

    steps:
    - uses: actions/checkout@v2

    - name: Fetch configlet
      uses: exercism/github-actions/configlet-ci@main

    - name: Configlet linter
      uses: configlet lint
```

@SleeplessByte will yell at me for suggesting to only run `push` builds on `main` but for configlet I think it's sufficient. Having `workflow_dispatch` as manual trigger is always useful. For those constructed examples of edge cases where one actually needs `push` in a PR branch (that imo will never occur for configlet), `workflow_dispatch` can be used. 

[Specifying the permissions will set all other permissions to `none`](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#permissions). Therefore it doesn't really matter if we pin the actions or not. If they're compromised, the worst that can happen would be abusing the runner resources. If that happens to `actions/checkout`, that's GH's fault.

action.yml
```diff
@@ -0,0 +1,20 @@
...
+  - run: mkdir -p bin/
+    shell: bash
+  - run: ${{ github.action_path }}/scripts/fetch-configlet
+    shell: bash
+  - run: chmod +x bin/configlet
+    shell: bash
+  - run: echo "$PWD/bin" >> $GITHUB_PATH
+    shell: bash
+  - run: echo "configlet version $(bin/configlet --version)"
+    shell: bash
```
SaschaMann, line 20:
```suggestion
    shell: bash

```

README.md
SleeplessByte: I won't yell but yes, why would we do this 👁️

SaschaMann: Because the config needs to be valid after being merged, the state of the config on the branch on its own is largely irrelevant. It's wasteful to have builds that are useless 99.9999% of the time they run.

wolf99:
Tests failing, but I dont think that is related to this change ?

ee7:
The CI failure will be resolved after I merge https://github.com/exercism/configlet/pull/466. Sorry about that. It's due to an unfortunate combination of issues that will hopefully be resolved upstream in the near future.

I haven't thought about this PR yet, though - and I'd like to. Can we avoid merging this until I've finished with #366 (shouldn't be too much longer now) and I've reviewed? Apologies.

wolf99:
Noticed the action path wasn't updated! 😄

README.md
```diff
@@ -6,7 +6,7 @@ This is a development version of Configlet for use with v3 of Exercism.
 
 The application is a single binary and can be used as follows:
 
-```
+```console
```
SleeplessByte, line 9:
````suggestion
```shell
````

I think this is the language, but might be shellscript or bash.

SleeplessByte (approved):

README.md
wolf99: `console` is an alias for a shell *session* (https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L5689) - i.e. a block that includes the prompt and output.

`shell` or `bash` is used for actual scripts (https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L5602) - i.e. that would go into a script file

SleeplessByte: Oh today I learned. Very nice.

wolf99: To be fair, I don't think either of them provide very good syntax highlighting for this purpose on GitHub (at least in my browser).

kotp: `shell` or `console` preferred over `bash` unless that is the specific shell being used, the other two are "generic".

ee7:
Commits 1fd2a46fddfb and 4a88c133fe98 made a bunch of changes to the repo `README.md`, and introduced some markdown issues - sorry.

Can we fix them in a separate PR, then rebase this PR on top? I've created #488.

ErikSchierboom:
I'd like to get this moving. Do people still feel that moving the github action workflow here is desirable? AFAIC, these are the advantages:

1. Everything related to configlet is in the same repo
2. The script to create a new release of the workflow is easier

The main disadvantages are:

1. There is no central place to easily find all exercism-wide workflows
2. There is possibly some duplication of code, due to workflows being spread of different repos

I'm not sure the advantages outweigh the disadvantages, so I'm curious to everyone's opinion.

SaschaMann:
The central place for that would be `org-wide-files`, wouldn't it? The workflow could still live there, this PR moves the action that's called by the workflow, not the workflow. Actions live all over the place already.

ErikSchierboom:
Okay. So no objections then?

README.md
ee7: Handled markdown linting separately in #488.

For now, we went with `text` instead of `console` here (because we don't currently show the command that produces the output).

ee7:
> The script to create a new release of the workflow is easier

@ErikSchierboom I didn't understand what this means (or refers to) - could you expand?

ee7:
Sorry - it took me a while to compose some thoughts on this.

As I understand it, part of the original motivation for moving the action was to get releases for **the action itself** (and because a shared repo of actions makes it messier for any action to have individual releases). But we think that we really do want `@main` for the configlet action because:
- The security and reproducibility benefits of version pinning are decreased for code that's entirely within our control.
- We really do want the latest configlet version to run in track CI immediately after a configlet release.
- Bumping the configlet version in every track repo is extremely noisy.

As I see it, there are at least 5 partially independent design choices. If we assume that the action should always download the latest configlet version (which I think it should, as it does now) rather than a fixed configlet version, then 4 remaining decisions are below.

Options for where the action can live:
- 1a. Keep the configlet action in the `exercism/github-actions` repo.
- 1b. Move the configlet action to the `exercism/configlet` repo.
- 1c. Move the configlet action to its own `exercism/configlet-action` repo.

Options for action versioning:
- 2a. The action has its own tags and releases, and tracks use a pinned action version. This seems a bit silly if the action always downloads the latest configlet version and does nothing else - then the action itself is probably not meaningfully updated.
- 2b. Unique option for the action being in `exercism/configlet`: the action is untagged, but configlet workflows on tracks pin the **configlet** tag, which pins the `action.yml` that downloads the latest configlet. This is confusing, and provides little benefit.
- 2c. The action is untagged, and tracks use `@main`.

Options for how the action downloads configlet:
- 3a. Use a Linux-only script, like [the `fetch-configlet` script that we use before this PR](https://github.com/exercism/github-actions/blob/22d087772dac/configlet-ci/fetch-configlet#L15).
- 3b. Use the same `fetch-configlet` script as the tracks.

Options for what the action does:
- 4a. It just downloads configlet (as it does currently)
- 4b. It runs `configlet lint` automatically, and maybe `configlet fmt` too
- 4c. It runs `configlet lint` automatically, with configuration options for other things.

I think my preferences are:
- Mild preference for **1c** over **1b**. Is that the more common design - a separate repo for the action? It also makes it easier to version the configlet action, in case that becomes more useful.
- Strong preference for **2c**.
- Mild preference for **3a**, especially if we do **1c**. At some point, there might be a reason for configlet to perform extra checks only during CI. For example, these checks could be slower than what we want locally, or use software/data that we don't want to add as a dependency for local use. If the configlet action no longer uses a Linux-only fetch script, then either those new checks have to be written in a platform-independent way, or they run only on one platform. If we do the latter, then we must ensure that tracks run configlet on that platform. This is OK if the configlet workflow lives in `org-wide-files`.
- Mild preference for **4a** for now, and maybe for **4c** eventually. Actions usually do more than just download something - for example, to run shellcheck we write just `uses: ludeeus/action-shellcheck@1.1.0` - we don't need to write a `shellcheck` command too. But if we want configlet to run the same on every track, then making the action configurable doesn't help much.

My biggest objection is to 2a (extra tags and releases in the configlet repo) due to these reasons that weren't already mentioned:
- It's better for tags in the configlet repo to be for configlet only, not the configlet action. That is, I don't want to add tags like `configlet-action-0.1.0` and consider changing the current format from e.g. `4.0.0-alpha.35` to `configlet-4.0.0-alpha.35`.
- It's better for releases in the configlet repo to be for configlet only, not the configlet action. Otherwise, the `fetch-configlet` scripts become more complicated: they currently find the latest (non-draft, non-prerelease) release in the repo, and try to download the configlet asset from that release. If the latest release was a configlet action release, then the `fetch-configlet` script as currently implemented would fail. The [releases page](https://github.com/exercism/configlet/releases) would be noisier too.

Thoughts?

wolf99:
Your preferences sound good to me @ee7

If I was being extremely picky I would say that 1b could be a good first step and we can move to 1c if and when future functionality is forthcoming. The same could be said of 3a vs 3b.

Similar to @ee7, I have a mild preference for 4a.

ErikSchierboom:
> Options for where the action can live:

Preference for 1b over 1a. I feel like 1c is likely to really benefit us anytime soon, as I don't see the action changing much anyway.

> Options for action versioning:

Strong preference for 2c.

> Options for how the action downloads configlet:

I don't really care much, but the argument mentioned about being able to do additional things makes me believe that 3a is a  slightly more futureproof approach.

> Options for what the action does:

Mild preference for 4a over 4c. I could imagine having linting run by default being nice, but we don't have to build that into the action and can instead build it into the org-wide-files workflow.

ee7:
@ErikSchierboom with the reusable workflow stuff (e.g. https://github.com/exercism/github-actions/commit/a02b33b33098a501cef966575014d31f9f6f6f68) are we now leaning towards closing this PR?

ErikSchierboom:
@ee7 No, this is just about the action that downloads configlet, not the actual workflow. But it will be a lot easier to update, as we only need to change this line: https://github.com/exercism/org-wide-files/blob/main/tracks-files/.github/workflows/configlet.yml#L15

PR #468, opened 2021-11-30
fmt: implement

ee7:
@ErikSchierboom what do you think about the interactivity here? Options:
1. `configlet fmt` prompts the user when the terminal is interactive (like `configlet sync`)
2. `configlet fmt` non-interactively formats.

For consistency with `configlet sync`, ~I did the latter~ (edit: I meant "I did option 1")

Ideally I'd have some "end-to-end" tests for this. But if you're happy, we can add them later.

To-do:
- [x] Support the `-e, --exercise` option that `configlet sync` supports.
- [x] Support the `-y, --yes` option that `configlet sync` supports. Note that
providing 'y' or 'yes' on stdin also works, like `echo 'y' | configlet fmt`
- [x] Merge the big sync PR first, and rebase this PR on top.

---

This commit adds the `fmt` command to configlet.

Running `configlet fmt` will print a list of paths for which there is
not already a formatted exercise `.meta/config.json` file, and write the
formatted files if the user confirms.

"Formatting" means:

- Write the key/value pairs in the canonical order.

- Use two spaces for indentation, and use a separate line for each item
  in a JSON array (and each key in a JSON object).

- Strip key/value pairs for keys that are optional and have empty
  values. For example, `contributors: []` is removed.

- Strip `test_runner: true` for Practice Exercise config files. This is
  an optional key - the spec says that an omitted `test_runner` key
  implies the value `true`.

The canonical order is:

    - authors
    - [contributors]
    - files
      - solution
      - test
      - exemplar           (Concept Exercises only)
      - example            (Practice Exercises only)
      - [editor]
    - [language_versions]
    - [forked_from]        (Concept Exercises only)
    - [icon]               (Concept Exercises only)
    - [test_runner]        (Practice Exercises only)
    - blurb
    - [source]
    - [source_url]
    - [custom]

where [key] indicates that the key is optional. This order tries to:

- maximize visual consistency between Concept and Practice Exercise
  configs by writing their unique keys in the same place.

- maximize visual consistency between files that contain and omit
  optional keys, by writing those keys later (e.g. `custom`), except
  that we:

  - write `contributors` under `authors`, since these keys are similar.

  - write `blurb` near the bottom, and above `source`, and `source_url`,
    since these keys are all synced with `problem-specifications` for a
    Practice Exercise.

Some possible future work:

- Support formatting the track `config.json` file.

### DIFF
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -337,8 +337,8 @@ proc parseActionKind(key: string): ActionKind =
   try:
     result = parseEnum[ActionKind](keyNormalized)
   except ValueError:
-    if keyNormalized == "format": # Silently accept `configlet format`.
-      result = actFmt
+    if keyNormalized == "format":
+      showError(&"invalid command: '{key}'\nDid you mean 'fmt'?")
     else:
       showError(&"invalid command: '{key}'")
 
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -427,11 +427,9 @@ proc handleOption(conf: var Conf; kind: CmdLineKind; key, val: string) =
     of actFmt:
       case opt
       of optFmtSyncExercise:
-        conf.action.exerciseFmt = val
-        isActionOpt = true
+        setActionOpt(exerciseFmt, val)
       of optFmtSyncYes:
-        conf.action.yesFmt = true
-        isActionOpt = true
+        setActionOpt(yesFmt, true)
       else:
         discard
     of actLint:
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -32,10 +32,11 @@ type
     of actNil:
       discard
     of actFmt:
-      # We can't name these fields `exercise` and `yes` because we use those
-      # names in `actSync`, and Nim doesn't yet support duplicate field names in
-      # object variants.
+      # We can't name these fields `exercise`, `update`, and `yes` because we
+      # use those names in `actSync`, and Nim doesn't yet support duplicate
+      # field names in object variants.
       exerciseFmt*: string
+      updateFmt*: bool
       yesFmt*: bool
     of actLint:
       discard
@@ -68,12 +69,12 @@ type
 
     # Options for both `fmt` and `sync`
     optFmtSyncExercise = "exercise"
+    optFmtSyncUpdate = "update"
     optFmtSyncYes = "yes"
 
     # Options for `sync`
     optSyncProbSpecsDir = "probSpecsDir"
     optSyncOffline = "offline"
-    optSyncUpdate = "update"
     # Scope to sync
     optSyncDocs = "docs"
     optSyncFilepaths = "filepaths"
@@ -96,8 +97,8 @@ const
   repoRootDir = currentSourcePath().parentDir().parentDir()
   configletVersion = staticRead(repoRootDir / "configlet.version").strip()
   short = genShortKeys()
-  optsNoVal = {optHelp, optVersion, optFmtSyncYes, optSyncOffline,
-               optSyncUpdate, optSyncDocs, optSyncFilepaths, optSyncMetadata}
+  optsNoVal = {optHelp, optVersion, optFmtSyncUpdate, optFmtSyncYes,
+               optSyncOffline, optSyncDocs, optSyncFilepaths, optSyncMetadata}
 
 func generateNoVals: tuple[shortNoVal: set[char], longNoVal: seq[string]] =
   ## Returns the short and long keys for the options in `optsNoVal`.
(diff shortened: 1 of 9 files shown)


ErikSchierboom:
> @ErikSchierboom what do you think about the interactivity here? Options:
>
> configlet fmt prompts the user when the terminal is interactive (like configlet sync)
> configlet fmt non-interactively formats.
> For consistency with configlet sync, I did the latter.

The latter also has my preference.

> Ideally I'd have some "end-to-end" tests for this. But if you're happy, we can add them later.

I'm fine with adding them later.

> Running configlet fmt will print a list of paths for which there is
not already a formatted exercise .meta/config.json file, and write the
formatted files if the user confirms.

Excellent.

> "Formatting" means:
> 
> [...]

Those formatting decisions all sound good.

> Support the -e, --exercise option that configlet sync supports.

Yes please!

> Support the -y, --yes option that configlet sync supports. 

That would be consistent _and_ useful.

> Support formatting the track config.json file.

👍

ee7:
@ErikSchierboom I've added some tests (just unit tests only for now) and added support for `-e` and `-y`. This PR will be just for the final two commits. 

Could you please test this PR?

src/cli.nim
```diff
@@ -32,7 +32,11 @@ type
     of actNil:
       discard
     of actFmt:
-      discard
+      # We can't name these fields `exercise` and `yes` because we use those
+      # names in `actSync`, and Nim doesn't yet support duplicate field names in
+      # object variants.
```
ErikSchierboom, line 37: Sad panda :(

src/cli.nim
```diff
@@ -171,13 +177,13 @@ func genHelpText: string =
     optTrackDir: "Specify a track directory to use instead of the current directory",
     optVerbosity: &"The verbosity of output.\n" &
                   &"{padding}{allowedValues(Verbosity)} (default: normal)",
-    optSyncExercise: "Only operate on this exercise",
+    optFmtSyncExercise: "Only operate on this exercise",
+    optFmtSyncYes: "Auto-confirm every prompt for updating docs, filepaths, and metadata",
```
ErikSchierboom, line 181: Maybe do:

> optFmtSyncYes: "Auto-confirm every prompt for formatting docs, filepaths, and metadata",

ErikSchierboom:
For an unformatted file, the output is currently:

> Not formatted: /home/erik/exercism/csharp/exercises/practice/zipper/.meta/config.json

However, if I run `configlet fmt` from the `/home/erik/exercism/csharp` dir, I think it might make sense to only output the relative path:

> Not formatted: ./exercises/practice/zipper/.meta/config.json

Minor detail: if I use `-e` to only format a single exercise, the output is still in plural:

> Format the above exercise configs

The behavior of `configlet fmt` is now slightly different than `configlet sync`, as the latter requires `--update` to be passed for it to do actual updating. That might be good to apply to `configlet fmt` too, for two reasons:

1. Consistency with `configlet sync`
2. It allows `configlet fmt` to return an exit code and be used in CI to verify formatting

What do you think?

Apart from the above, all seems to be working well!

src/cli.nim
ee7: Indeed.

src/cli.nim
ee7: It's confusing, but this line is actually defining the description of the `--yes` option as used for `configlet sync`.

We're using `parseEnum` to get the option, so we can't have separate `optFmtYes` and `optSyncYes` both with the string value of `"yes"`. So right now, there's some mess to set a different description for `--yes` as an option for `fmt`, and I think we really do want different descriptions, because it's complicated that `--yes` doesn't auto-confirm a prompt for `--tests`

It's one of the things I was thinking of when I wrote in https://github.com/exercism/configlet/pull/468/commits/4e11cabd803e27d2956f01eec7320eb725b04e47:
> This could use some refactoring later. In particular, the generation
of the help message is becoming convoluted.

But "formatting docs, filepaths, and metadata" doesn't make sense, right? Because only `configlet sync` will operate on docs, and not `configlet fmt`, right? And when writing a `.meta/config.json` file with `configlet sync`, we try to avoid "formatting-only" changes, and only update when there was unsynced data.

With "updating" I was trying to imply that `--yes` is only meaningful as a `sync` option when `--update` is passed. Do you prefer e.g. "writing"?

ErikSchierboom: Thanks for explaining. Let's hold off on this for now.

ee7:
> if I run `configlet fmt` from the `/home/erik/exercism/csharp` dir, I think it might make sense to only output the relative path

I think using `./` is okay when `configlet` is using a track that's in the current directory. But what about when the `-t, --track-dir` option is used? I think we should avoid `./` in that case - a user might have a track in the current directory too. Thoughts?

Maybe we could add this to the `configlet fmt` output?
```
the above paths are relative to this directory:
/foo/bar"
```

> Minor detail: if I use `-e` to only format a single exercise, the output is still in plural:

Yeah - will fix.

> The behavior of `configlet fmt` is now slightly different than `configlet sync`, as the latter requires `--update` to be passed for it to do actual updating. That might be good to apply to `configlet fmt` too

I agree that `sync` and `fmt` should be consistent. And I guess `generate` too.

> It allows `configlet fmt` to return an exit code and be used in CI to verify formatting

I want `configlet fmt` to be usable in CI too. In the v2 era, there were a handful of tracks (including `exercism/nim`) that used `configlet fmt` (and the old `configlet generate`) in CI,. But it required running `git diff --exit-code` or `git diff --quiet` afterwards.

I guess the lack of `--update` is blocking, right? I guess I can add it, and we can discuss later if necessary.

ErikSchierboom:
> I think using ./ is okay when configlet is using a track that's in the current directory. But what about when the -t, --track-dir option is used? I think we should avoid ./ in that case - a user might have a track in the current directory too. Thoughts?

I would definitely use an absolute path if `--track-dir` is specified, yes.

> Maybe we could add this to the configlet fmt output?

Ehm, maybe rephrase slightly to something like:

< Formatting files in the <..> directory:

> I guess the lack of --update is blocking, right? I guess I can add it, and we can discuss later if necessary.

Yeah I think so.

src/cli.nim
ee7: OK. Added a comment in https://github.com/exercism/configlet/pull/468/commits/94ff798d4967f67193d01187574f9b9730256374

ee7:
@ErikSchierboom please try again with the latest changes. I've tried to address all the points raised so far. Good enough?

ErikSchierboom:
The `--update` functionality works great! I like that a lot.

As for the relative file path, we currently output:

> The below paths are relative to '/home/erik/exercism/csharp/exercises'
> Not formatted: concept/annalyns-infiltration/.meta/config.json

I think I'd opt for having the base path be `/home/erik/exercism/csharp' and then including the `exercises/` prefix.
The reason for this being that we'll also be updating files in the `concept` directory and in the root.

ee7:
OK - please take another look.

We could also consider omitting the trailing `.meta/config.json` in the paths, with some rephrasing.

Something like:
```
Found 36 Concept Exercises and 94 Practice Exercises in /path/to/track-dir/config.json
Looking for concepts and exercises that lack a formatted '.meta/config.json' file...

No formatted '.meta/config.json' file in these concept directories, relative to '/path/to/track-dir':
concept/foo
concept/bar

No formatted '.meta/config.json' file in these exercise directories, relative to '/path/to/track-dir':
exercises/concept/aaa
exercises/practice/bob

Format the above config files ([y]es/[n]o)?
```

But I'd prefer to leave that for later.

ErikSchierboom (approved):

ErikSchierboom:
I think it's great now!

ee7:
Great. Looks like we've (finally) resolved the last blocker for the release, then. I'll release as soon as I can.

---

Some thoughts for later: when `fmt` learns to format concept `.meta/config.json` files, I think we'll need to decide between:
- Adding a `--concept` option
- Renaming `--exercise` to `--slug`, which can then be used for formatting an exercise or a concept (although this is probably bad if a concept can have the same name as an exercise)

I guess `configlet fmt` could eventually grow some scoping options like `sync`, e.g.:
- `--track`
- `--exercises`
- `--concepts`

With the default of operating on all of them.

PR #470, opened 2021-12-08
cli: refactor

ee7:
This is a small refactoring-only PR.


ErikSchierboom (approved):
I agree with the changes in all four of those commits. All improvements 👍

ee7:
I dropped the change from 88c405d0f8bd - it's hard to see in the squashed diff, and I have a better fix anyway. We can generate the list of commands that take an option directly from the type.

Issue #484, opened 2021-12-18
build: create only one draft release


ee7:
https://github.com/exercism/configlet/commit/a103ac4607d9eb6e4217cf9d324705fc43ad012c didn't fix the issue, and I believe I know why.

In the below, `latest_release_tag` is not updated after the first draft release, because `gh release view` returns the latest non-draft release.

https://github.com/exercism/configlet/blob/d290a303715662664064166658ab25db76cfa8c1/.github/bin/publish-release#L1-L9

PR #486, opened 2021-12-18
github: create only one draft release

ee7:
Even after commit a103ac4607d9, three draft releases were created during
the latest release build.

It looks like the flaw was that `gh release view` returns the latest
non-draft release, and so `gh release create --draft` always ran.

Fixes: #484

---

This time, I tested it better. It seems to work here: https://github.com/ee7/exercism-configlet/releases/tag/untagged-820ff06a337a5d420c03

It seems that `gh release view foo` returns an exit code of 0 when there is a draft release or release with the tag `foo`, and an exit code of 1 otherwise.

### DIFF
--- a/.github/bin/publish-release
+++ b/.github/bin/publish-release
@@ -1,9 +1,8 @@
 #!/usr/bin/env sh
 
-latest_release_tag="$(gh release view --json tagName --jq '.tagName')"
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
The new version is also easier to read!

Issue #482, opened 2021-12-18
info: use a non-custom serialization for prob-specs slugs


ee7:
As mentioned in https://github.com/exercism/configlet/pull/474#issuecomment-989727131

This would make the code a little cleaner until we do https://github.com/exercism/configlet/issues/483.

PR #487, opened 2021-12-19
info: refactor to use JSON prob-specs slugs list

ee7:
Originally, we baked the list of `problem-specifications` exercise slugs
into the configlet executable simply by writing the slugs
newline-separated in a file, and parsing that into a `HashSet` of slugs
at compile-time. This is arguably a little cleaner than writing all the
slugs as a `const` in a nim file if we wanted to generate and update the
list automatically.

Then, we wanted to make `configlet info` omit upstream-deprecated
exercises when listing the unimplemented exercises on a track. So it
made sense to group the slugs by kind (rather than just omitting or
commenting-out the deprecated slugs in the list) and
commit f798115c2250 did that grouping via a simple custom file format
and parser.

With this commit, we refactor that grouping to use a JSON-formatted list
of prob-specs slugs, and thereby remove the custom parsing code. This
also makes it easier to bake-in more information that we could display
to the end user, like a timestamp and commit ref for the prob-specs
state.

We still parse the file and populate the `HashSet`s at compile-time, so
the main cost is that the file is now more verbose (we need quotes and
commas now).

Eventually, `configlet info` will probably use a cached clone of
`problem-specifications`, and respect the `--offline` and
`--prob-specs-dir` options that work for `configlet sync`. Then we won't
need to include the slugs in the configlet executable, and we won't need
to update that list whenever an exercise in prob-specs is added,
deprecated, or gains canonical data.

This is a refactoring-only commit. It doesn't change the list of
exercises in `problem-specifications`.

Closes: #482


ErikSchierboom (approved):
Very nice!

PR #488, opened 2021-12-19
README: lint

ee7:
#462 included some markdown linting, but commits 1fd2a46fddfb and 4a88c133fe98 added a lot to the repo `README.md`, and introduced some markdown issues.

Let's fix these markdown issues in this PR, then rebase that PR on top.

Resolved `markdownlint` rule violations:

- [MD014](https://github.com/DavidAnson/markdownlint/blob/v0.24.0/doc/Rules.md#md014)
- [MD040](https://github.com/DavidAnson/markdownlint/blob/v0.24.0/doc/Rules.md#md040)
- [MD032](https://github.com/DavidAnson/markdownlint/blob/v0.24.0/doc/Rules.md#md032)
- [MD034](https://github.com/DavidAnson/markdownlint/blob/v0.24.0/doc/Rules.md#md034)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ This is a development version of Configlet for use with v3 of Exercism.
 
 The application is a single binary and can be used as follows:
 
-```
+```text
 Usage:
   configlet [global-options] <command> [command-options]
 
@@ -84,26 +84,26 @@ A Practice Exercise that is derived from the `problem-specifications` repo must 
 
 To check every Practice Exercise on the track for available documentation updates (exiting with a non-zero exit code if at least one update is available):
 
-```
+```shell
 configlet sync --docs
 ```
 
 To interactively update the docs for every Practice Exercise, add the `--update` option (or `-u` for short):
 
-```
+```shell
 configlet sync --docs --update
 ```
 
 To non-interactively update the docs for every Practice Exercise, add the `--yes` option (or `-y` for short):
 
-```
+```shell
 configlet sync --docs --update --yes
 ```
 
 To operate on a single Practice Exercise, use the `--exercise` option (or `-e` for short).
 For example, to non-interactively update the docs for the `prime-factors` exercise:
 
-```
+```shell
 configlet sync --docs -uy -e prime-factors
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,7 +6,7 @@ This is a development version of Configlet for use with v3 of Exercism.
 
 The application is a single binary and can be used as follows:
 
-```
+```text
```
ee7, line 9: @wolf99 I notice that #462 made this `console` instead - is that better? 

>  - `console` is an alias for a shell *session*
> (https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L5689) - i.e. a block that includes the prompt and output.

seems not apply to this block - for that we'd need to add `$ configlet --help` at the top, right?

README.md
```diff
@@ -346,13 +348,13 @@ Each concept exercise and concept have an `introduction.md` file. If you want th
...
 %{concept:<slug>}
 ```
 
 For example, if the track has a concept named `floating-point-numbers` then an `introduction.md.tpl` file can contain:
 
-```
+```text
```
ee7, line 357: @wolf99 I notice that #462 used `markdown` instead as the language for these. Is that better?

README.md
wolf99: Could reasonably be either I suppose.
I changed to `console` based on this being the output we see in a session when `configlet --help` is used, despite the command itself not being included.
If you think that does't make sense I think it's OK to set it to `text`, or could include the command and leave at `console` ?

README.md
wolf99: This is based on the converstation on Slack here: https://exercism-team.slack.com/archives/CAQP7JL3T/p1635445165124700

I dont think the discussion was conclusive though - `text` could be good either.

ErikSchierboom (approved):

README.md
ee7: Thanks.

Let's go with `text` for now, and make it `console` in the future if we add `$ configlet --help` at the start.

README.md
ee7: Thanks.

Let's go with `text` here. `configlet generate` uses custom templating syntax - even if it might have the same syntax as some other templating language, it's confusing to suggest that configlet actually uses that language internally. And we'd barely benefit from syntax highlighting in this code block anyway.

Issue #500, opened 2022-01-22
sync throws error when config.json has `"contributors": null`


verdammelt:
Not sure if this is an issue with `configlet` _per se_ or a general problem of my track's practice execise config.json files. 

Out of 43 exercises my track has 15 which has `"contributors": null` in their config.json files. (I don't think I did that myself so perhaps it is a left-over of some earlier tooling?)

At least one of these needs their metadata updated (for example 'binary') and I get the following when i try to update it:

```
./bin/configlet sync --update --yes --metadata -e binary
Checking exercises...
Cloning https://github.com/exercism/problem-specifications/... success
[warn] metadata: unsynced: binary
options.nim(194)         get
Error: unhandled exception: Can't obtain a value from a `none` [UnpackDefect]
```

Replacing the contributors value with `[]` or removing it altogether corrects the issue.

BethanyG:
@verdammelt -- I _think_ there was a change in the  `config.json` spec a while ago that required either **no key at all** -- or a `[]` (array) for the contributors key.   But configlet wasn't enforcing it.  Now, it's enforced.

verdammelt:
If by 'enforced' you mean: "sync will throw an error" then yep - seeing that behavior. If you mean "lint will signal a problem"  then nope...

BethanyG:
_oof!_  Yep, that looks like a problem...

verdammelt:
@BethanyG thanks for the info. Sounds like this might actually be a lint issue...

I'm going to push a commit that fixes all my configs... I'll leave this issue so correct fix can be discussed

ee7:
@verdammelt Please run `configlet fmt -uy` as a workaround - this will format the exercise config files, removing the `"contributors": null` in the process.

I can confirm that the common-lisp track is the only track right now with this problem:

```console
$ rg --sort path --files-with-matches -uu '"contributors": null'
common-lisp/exercises/practice/all-your-base/.meta/config.json
common-lisp/exercises/practice/atbash-cipher/.meta/config.json
common-lisp/exercises/practice/binary/.meta/config.json
common-lisp/exercises/practice/binary-search/.meta/config.json
common-lisp/exercises/practice/collatz-conjecture/.meta/config.json
common-lisp/exercises/practice/crypto-square/.meta/config.json
common-lisp/exercises/practice/grains/.meta/config.json
common-lisp/exercises/practice/hamming/.meta/config.json
common-lisp/exercises/practice/isogram/.meta/config.json
[...]

PR #501, opened 2022-01-25
sync: fix null handling when updating filepaths/metadata

ee7:
Before this commit, when updating a `.meta/config.json` file that
contained an optional key with the value of `null`, configlet would
produce an `UnpackDefect`. For example:

    git clone https://github.com/exercism/common-lisp
    cd common-lisp
    git checkout 4fc524e24a69
    configlet sync --metadata -e binary -u
    Checking exercises...
    Cloning https://github.com/exercism/problem-specifications/... success
    [warn] metadata: unsynced: binary
    sync the above metadata ([y]es/[n]o)? y
    options.nim(194)         get
    Error: unhandled exception: Can't obtain a value from a `none` [UnpackDefect]

For now, let's resolve the problem by stripping any optional key with
the value of `null` when updating filepaths or metadata.

Fixes: #500

### DIFF
--- a/src/sync/sync_common.nim
+++ b/src/sync/sync_common.nim
@@ -490,20 +490,23 @@ proc pretty*(e: ConceptExerciseConfig | PracticeExerciseConfig,
     of eckAuthors:
       result.addArray("authors", e.authors)
     of eckContributors:
-      result.addArray("contributors", e.contributors.get())
+      if e.contributors.isSome():
+        result.addArray("contributors", e.contributors.get())
     of eckFiles:
       result.addFiles(e.files, prettyMode)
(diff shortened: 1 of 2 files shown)


ErikSchierboom (approved):

ee7:
I don't know how outrageous it is to transform the `untyped` field name into a string via `strformat` (see [this line](https://github.com/exercism/configlet/pull/501/commits/aa8cc637c4cccc03ade9b1bfc30815bcd9d96ce8#diff-3f470d6fedf3c42b163870c75eee51dfc210c98fb818a4ea7809353c4c2d79f8R471)).

Issue #424, opened 2021-09-13
uuid: use best practices for UUID generation


ee7:
Low priority.

Currently to generate a UUID we use:
- `pragmagic/uuids`: [uuids.nim](https://github.com/pragmagic/uuids/blob/8cb8720b567c6bcb261bd1c0f7491bdb5209ad06/uuids.nim#L76-L92)
- `pragmagic/isaac`: [isaac.nim](https://github.com/pragmagic/isaac/blob/45a5cbbd54ff59ba3ed94242620c818b9aad1b5b/src/isaac.nim) (a port of https://burtleburtle.net/bob/c/readable.c)

This approach:
>  Uses a thread-local cryptographically secure PRNG (ISAAC) seeded with true random values obtained from OS.

But ISAAC is outdated, and it "remains to be seen" whether ISAAC is actually "cryptographically secure"  (at least according to Thomas Pornin, a notable cryptographer).

See also:
- https://burtleburtle.net/bob/rand/isaacafa.html
- https://burtleburtle.net/bob/rand/isaac.html
- https://en.wikipedia.org/wiki/ISAAC_(cipher)
- https://crypto.stackexchange.com/a/1744
- https://security.stackexchange.com/a/33670
- https://news.ycombinator.com/item?id=11626088
- https://news.ycombinator.com/item?id=10631806

It's not strictly necessary to use a cryptographically secure PRNG if `configlet uuid` is **only** used for populating values in a track `config.json` file. But I don't want to write/explain warnings everywhere saying:

> DANGER: `configlet uuid` is only suitable for ...

---

It's better to use the operating system's built-in UUID generation, as seen in:
- `status-im/nim-eth`: [eth/keyfile/uuid.nim](https://github.com/status-im/nim-eth/blob/e219547d640723797dfb84b9dedc6b5e4809e1e5/eth/keyfile/uuid.nim)
- `cheatfate/nimcrypto`: [sysrand.nim](https://github.com/cheatfate/nimcrypto/blob/a5742a9a214ac33f91615f3862c7b099aec43b00/nimcrypto/sysrand.nim)

which generates the UUIDs as follows:
| Operating system | UUID generation method |
| - | - |
| Windows | `UuidCreate()` from `rpcrt4.dll` |
| Linux, Android | `/proc/sys/kernel/random/uuid` |
| MacOS, iOS | `uuid_generate_random()` |
| FreeBSD, OpenBSD, NetBSD, DragonflyBSD | `uuid_create()` |

[...]

PR #508, opened 2022-01-26
cli: avoid the word "generate" in uuid help text

ee7:
This removes potential for confusion with `configlet generate`.

---

A bit nitpick-y, but worthwhile?

Another reason: if `configlet uuid` is ever updated to just "read" from a system UUID provider (see #424), we might say that it no longer "generates" the UUIDs.


ErikSchierboom (approved):
🤷

PR #510, opened 2022-01-26
cli: add configlet description to help text

ee7:
The help message was missing a description of what configlet does.

The help message in the configlet for Exercism v2 began like:

```console
$ ./configlet --help
./configlet version 3.9.2

A tool for managing Exercism language track repositories.

Usage:
  ./configlet [command]

[...]
```

### DIFF
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -217,6 +217,9 @@ func genHelpText: string =
   const binaryExt = when defined(windows): ".exe" else: ""
 
   result = fmt"""
+    configlet {configletVersion}
+    The official tool for managing Exercism language track repositories.
+
     Usage:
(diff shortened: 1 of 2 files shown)


ErikSchierboom (approved):
Good idea. I wonder whether we might also want to point people to the docs: https://exercism.org/docs/building/configlet

ee7:
> Good idea. I wonder whether we might also want to point people to the docs: https://exercism.org/docs/building/configlet

Sounds good. Especially until we have per-command help descriptions and examples in the executable itself.

But let's handle it in a follow-up - it'll require a further tweak to the tests. And we should update those docs first - the docs are currently only up to date at `exercism/configlet`.

PR #511, opened 2022-01-26
cli: refactor help message construction

ee7:
Refactoring only.

The main improvement is 67927847db08 - we no longer need to specify the commands that take options.

### DIFF
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -125,7 +125,8 @@ func list*(opt: Opt): string =
     &"-{short[opt]}, --{camelToKebab($opt)}"
 
 func genHelpText: string =
-  ## Returns a string that lists all the CLI options.
+  ## Generates a string that describes every configlet command and option, to be
+  ## shown in the `configlet --help` message.
 
   func allowedValues(T: typedesc[enum]): string =
     ## Returns a string that describes the allowed values for an enum `T`.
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):

ee7:
> Isn't result `0` by default?

Yes, but it looks like Nim will require that every variable is explicitly initialized before use in the future. As far as I understand, the rationale is something like:
- It stops special-casing of types that lack a default value.
- It removes a class of bugs.
- It removes the need for the compiler to prove that uninitialized variables aren't accessed.

But the old style will keep working for a long time:
> Code like the following `var x: T; use x` is not allowed anymore. As a long transition
period the compiler will warn about the "usage before definition" and translate it as
if `var x = default(T)`; use x was written.

See:
- [nim-lang/Nim/blob/4b723c0f5350/lib/system.nim#L1843-L1850](https://github.com/nim-lang/Nim/blob/4b723c0f5350/lib/system.nim#L1843-L1850)
- https://github.com/nim-lang/Nim/commit/da29222f86f7689227ffe12605842d18c9bf0fc1#diff-c7ae564e61082887ea50f0d58a637cb12fa78261f51aa4eface24ababfeee299R1770 (explicit `result = 0` was added in `system.find`)
- [nim-lang/Nim/pull/19413#issuecomment-1016376360](https://github-redirect.dependabot.com/nim-lang/Nim/pull/19413#issuecomment-1016376360) 
- [nim-lang/RFCs/issues/437](https://github-redirect.dependabot.com/nim-lang/RFCs/issues/437)
- [nim-lang/RFCs/issues/378](https://github-redirect.dependabot.com/nim-lang/RFCs/issues/378)

Similarly, it's better to use `initHashSet`:
https://github.com/exercism/configlet/blob/a6c27d34a2ff4a5f01fc976df63a8ad756a6d997/src/info/info.nim#L35-L41

even though right now you can `result.incl` without it.

ErikSchierboom:
Hah, interesting! I always quite liked not having to do the initialization, but I can see the argument against it.

PR #517, opened 2022-02-03
lint: support error annotations

ErikSchierboom:
This PR adds support for linting errors to also print an annotation,
which can be used to give more detailed information on a certain
linting rule validation error.

For example, the following is now printed when the `tags` key in the
track's `config.json` file does not validate:

```
The `tags` array is empty:
../csharp/config.json
Tracks are annotated with tags to allow searching for tracks with certain tags.
The tags should be chosen based on the general usage of their language.
For more information and the full list of supported tags see:
https://exercism.org/docs/building/tracks/config-json#h-tags
```

Note that the first line ("The `tags` array is empty:") is output in the color red, while the annotation lines are printed in yellow.

### DIFF
--- a/src/helpers.nim
+++ b/src/helpers.nim
@@ -25,7 +25,7 @@ proc getSortedSubdirs*(dir: Path): seq[Path] =
       result.add Path(path)
   sort result
 
-proc setFalseAndPrint*(b: var bool; description: string; path: Path) =
+proc setFalseAndPrint*(b: var bool; description: string; path: Path, annotation = "") =
   ## Sets `b` to `false` and writes a message to stdout containing `description`
   ## and `path`.
   b = false
@@ -35,6 +35,11 @@ proc setFalseAndPrint*(b: var bool; description: string; path: Path) =
   else:
     stdout.writeLine(descriptionPrefix)
   stdout.writeLine(path.string)
+  if annotation.len > 0:
+    if colorStdout:
+      stdout.styledWriteLine(fgYellow, annotation)
+    else:
+      stdout.writeLine(descriptionPrefix, annotation)
   stdout.write "\n"
 
 var printedWarning* = false
--- a/src/lint/validators.nim
+++ b/src/lint/validators.nim
@@ -37,30 +37,32 @@ func format*(context, key: string): string =
     q(key)
 
 proc hasKey(data: JsonNode; key: string; path: Path; context: string;
-            isRequired: bool): bool =
+            isRequired: bool; errorAnnotation = ""): bool =
   ## Returns true if `data` contains the key `key`. Otherwise, returns false
   ## and, if `isRequired` is true, prints an error message.
   if data.hasKey(key):
     result = true
   elif isRequired:
-    result.setFalseAndPrint(&"The {format(context, key)} key is missing", path)
+    result.setFalseAndPrint(&"The {format(context, key)} key is missing", path,
+                             annotation = errorAnnotation)
 
-proc isObject*(data: JsonNode; key: string; path: Path; context = ""): bool =
+proc isObject*(data: JsonNode; key: string; path: Path; context = "";
+               errorAnnotation = ""): bool =
   if data.kind == JObject:
     result = true
   else:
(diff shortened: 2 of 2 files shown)


src/helpers.nim
```diff
@@ -25,7 +25,7 @@ proc getSortedSubdirs*(dir: Path): seq[Path] =
       result.add Path(path)
   sort result
 
-proc setFalseAndPrint*(b: var bool; description: string; path: Path) =
+proc setFalseAndPrint*(b: var bool; description: string; path: Path, annotation = "") =
```
ErikSchierboom, line 28: I'm open to any refactoring/renaming suggestions.

src/lint/track_config.nim
```diff
@@ -228,8 +228,12 @@ const tags = [
 ].toHashSet()
 
 proc hasValidTags(data: JsonNode; path: Path): bool =
+  const errorAnnotation = "Tracks are annotated with tags to allow searching for tracks with certain tags.\n" &
+                          "The tags should be chosen based on the general usage of their language.\n" &
+                          "For more information and the full list of supported tags see:\n" &
+                          "https://exercism.org/docs/building/tracks/config-json#h-tags"
```
ErikSchierboom, line 234: This is the most important part, where we direct people to the documentation page.

ee7, line 234:
```suggestion
  const errorAnnotation = """
    Tracks are annotated with tags to allow searching for tracks with certain tags.
    The tags should be chosen based on the general usage of their language.
    For more information and the full list of supported tags see:
    https://exercism.org/docs/building/tracks/config-json#h-tags""".unindent()
```

Nit: can we use triple-quotes string literals and `unindent`? It's the style I favored elsewhere, with the rationale that it's a little cleaner.

But writing multiline strings in Nim has always felt a bit inelegant to me - I don't love either style.

See [nim-lang/RFCs#161](https://github-redirect.dependabot.com/nim-lang/RFCs/issues/161) and [nim-lang/RFCs#248](https://github-redirect.dependabot.com/nim-lang/RFCs/issues/248) (with redirects via dependabot).

ee7 (commented):
Nice - I'd like to improve the error messages too.

I'm OK with the word "annotation", but it does make me (and maybe others) think of an annotation in the context of GitHub Actions.

Non-blocking question: Was "annotation" chosen with the intent of "this message will eventually appear in the annotations part of a PR"?

One problem with the `std/json` approach is that we don't have line number information at the time of linting, so it's only simple to produce an annotation on GitHub at the correct line for a subset of errors (malformed JSON). `jsony` does have the line number information, but there is another subset of errors that `jsony` does not straightforwardly catch, so we might be stuck with both.

But we could move `lint` towards "first pass with `std/json` to check syntax, second pass with `jsony` for everything else."

---

(I need to test this PR locally before approving - will test later today).

ErikSchierboom:
> Non-blocking question: Was "annotation" chosen with the intent of "this message will eventually appear in the annotations part of a PR"?

It wasn't, but it is a good idea. 

> But we could move lint towards "first pass with std/json to check syntax, second pass with jsony for everything else."

Sounds good.

ErikSchierboom:
> (I need to test this PR locally before approving - will test later today).

Bump @ee7 :)

ee7 (commented):
Sorry for the insane delay.

Output looks like e.g.:

```
The value of `blurb` is a zero-length string:
../tracks/ada/config.json

The `tags` array is empty:
../tracks/ada/config.json
Tracks are annotated with tags to allow searching for tracks with certain tags.
The tags should be chosen based on the general usage of their language.
For more information and the full list of supported tags see:
https://exercism.org/docs/building/tracks/config-json#h-tags

Missing file:
../tracks/ada/docs/SNIPPET.txt
```

(Edit: please ignore the previous edit - had a bug in one of my scripts).

ee7 (approved):

ErikSchierboom:
I've split the original PR into two. This one to add the error annotation support, and https://github.com/exercism/configlet/pull/519 which will adds the actual annotation messages.

PR #520, opened 2022-02-16
lint: link to docs on exercism.org, not GitHub

ee7:
Better to link to exercism.org everywhere, I guess?

Follow-up from https://github.com/exercism/configlet/pull/517#pullrequestreview-884706384


ErikSchierboom (approved):
Good idea

PR #521, opened 2022-02-17
Allow file pattern overlap exceptions

ErikSchierboom:
- lint: allow overlap of `files.example` and `files.exemplar` values
- lint: allow overlap of solution and test files for `d` and `plsql` tracks

See https://github.com/exercism/docs/pull/321/files

### DIFF
--- a/src/lint/track_config.nim
+++ b/src/lint/track_config.nim
@@ -293,7 +293,16 @@ type
 
   Concepts* = seq[Concept]
 
+  FilePatterns* = object
+    solution*: seq[string]
+    test*: seq[string]
+    example*: seq[string]
+    exemplar*: seq[string]
+    editor*: seq[string]
+
   TrackConfig* = object
+    slug*: string
+    files*: FilePatterns
     exercises*: Exercises
     concepts*: Concepts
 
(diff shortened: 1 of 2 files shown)


src/lint/track_config.nim
```diff
@@ -293,7 +293,16 @@ type
...
+  FilePatterns* = object
+    solution*: seq[string]
+    test*: seq[string]
+    example*: seq[string]
+    exemplar*: seq[string]
+    editor*: seq[string]
+
   TrackConfig* = object
+    slug*: string
+    files*: FilePatterns
```
ErikSchierboom, line 305: I've considered making this an `Option[FilePatterns]` value, but the default handling (making the patterns be empty sequences) works well in this case. Let me know what you think.

src/lint/track_config.nim
```diff
@@ -293,7 +293,16 @@ type
...
 
+  FilePatterns* = object
+    solution*: seq[string]
+    test*: seq[string]
+    example*: seq[string]
+    exemplar*: seq[string]
+    editor*: seq[string]
+
   TrackConfig* = object
+    slug*: string
```
ErikSchierboom, line 304: I added the slug to allow for the `d` and `plsql` tracks to be exceptional cases.

PR #535, opened 2022-03-31
ci: add workflow to lint whitespace

### LINKED ISSUE #534: ci: check whitespace (trailing whitespace, missing final newlines)
This can speed up reviews and prevent PRs like #533 in the future.

ee7:
Closes #534.

Failing check that ran before #533 was merged: https://github.com/exercism/configlet/runs/5779411023

### DIFF
--- a/.github/bin/lint-whitespace
+++ b/.github/bin/lint-whitespace
@@ -7,7 +7,7 @@ repo_root="$(git rev-parse --show-toplevel)"
 cd "${repo_root}" || exit
 exit_status=0
 
-# Check that each git-tracked plaintext file has no trailing whitespace.
+# Check that each git-tracked non-binary file has no trailing whitespace.
 trailing_whitespace="$(git grep --cached -I '[[:blank:]]$')"
 if [ -n "${trailing_whitespace}" ]; then
   echo "There is trailing whitespace on the below lines:"
@@ -15,7 +15,7 @@ if [ -n "${trailing_whitespace}" ]; then
   exit_status=1
 fi
 
-# Check that each git-tracked plaintext file has one final newline.
+# Check that each git-tracked non-binary file has one final newline.
 for file in $(git grep --cached -I --files-with-matches ''); do
   if [ "$(tail -c 1 "${file}" | wc -l)" -eq 0 ]; then
     echo "No newline at end of file: ${file}"
--- a/.github/bin/lint-whitespace
+++ b/.github/bin/lint-whitespace
@@ -9,22 +9,15 @@ if git grep --cached -Ir '[[:blank:]]$'; then
   exit_status=1
 fi
 
-# Check for missing final newline.
+# Check for exactly one final newline.
 for file in $(git grep --cached -Il ''); do
   if [ "$(tail -c 1 "${file}" | wc -l)" -eq 0 ]; then
     echo "No newline at end of file: ${file}"
     exit_status=1
-  fi
-done
-
-# Check for multiple final newlines.
-if type rg > /dev/null; then
-  if rg -l --multiline '\n\n\z'; then
+  elif [ "$(tail -c 2 "${file}" | wc -l)" -eq 2 ]; then
+    echo "Multiple newlines at end of file: ${file}"
     exit_status=1
   fi
-else
-  exit_status=1
-  echo 'rg is not installed'
-fi
+done
 
 exit "${exit_status}"
--- a/.github/bin/lint-whitespace
+++ b/.github/bin/lint-whitespace
@@ -1,0 +1,25 @@
+#!/usr/bin/env sh
+
+repo_root=$(git rev-parse --show-toplevel)
+cd "${repo_root}" || exit
+exit_status=0
+
+# Check for trailing whitespace.
+if git grep --cached -Ir '[[:blank:]]$'; then
+  exit_status=1
+fi
+
+# Check for missing final newline.
+for file in $(git grep --cached -Il ''); do
+  if [ "$(tail -c 1 "${file}" | wc -l)" -eq 0 ]; then
+    echo "No newline at end of file: ${file}"
+    exit_status=1
+  fi
+done
+
+# Check for multiple final newlines.
+if rg -l --multiline '\n\n\z'; then
+  exit_status=1
+fi
+
+exit "${exit_status}"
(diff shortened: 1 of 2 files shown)


.github/workflows/lint-whitespace.yml
```diff
@@ -0,0 +1,19 @@
...
+      - name: Checkout
+        uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846
+
+      - name: Check for trailing whitespace
+        shell: bash
+        run: if git grep -Ir '[[:blank:]]$'; then exit 1; fi
+
+      - name: Check for missing final newlines
+        shell: bash
+        run: git grep --cached -Il '' | xargs -L1 bash -c 'if test "$(tail -c 1 "$0")"; then echo "No new line at end of $0"; exit 1; fi'
```
ee7: Taken from https://stackoverflow.com/a/54348931

Ideally, it would error if the file ended in multiple newlines.

.github/workflows/lint-whitespace.yml
```diff
@@ -0,0 +1,19 @@
...
+  lintwhitespace:
+    name: Check for trailing whitespace and missing final newline
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846
+
+      - name: Check for trailing whitespace
+        shell: bash
+        run: if git grep -Ir '[[:blank:]]$'; then exit 1; fi
```
ee7, line 15: Should probably put `--cached` here too.

ee7 (commented):
There are some GitHub Actions for whitespace linting, but I didn't immediately see a great one.

.github/workflows/lint-whitespace.yml
ee7: Probably add `--cached` here too.

.github/workflows/lint-whitespace.yml
```diff
@@ -0,0 +1,19 @@
+name: Lint whitespace
+
+on: [push, pull_request, workflow_dispatch]
```
ErikSchierboom, line 3: You could consider doing:

```suggestion
on:
  pull_request:
  push:
    branches:
      - main
  workflow_dispatch:
```

That way, pushes to PRs are not doubly run.

.github/workflows/lint-whitespace.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  lintwhitespace:
```
ErikSchierboom: Maybe do:

```suggestion
  lint_whitespace:
```

.github/bin/lint-whitespace
```diff
@@ -0,0 +1,25 @@
...
+# Check for missing final newline.
+for file in $(git grep --cached -Il ''); do
+  if [ "$(tail -c 1 "${file}" | wc -l)" -eq 0 ]; then
+    echo "No newline at end of file: ${file}"
+    exit_status=1
+  fi
+done
+
+# Check for multiple final newlines.
+if rg -l --multiline '\n\n\z'; then
```
ee7: Oh, I thought `rg` was installed in the GitHub Actions environment...

.github/workflows/lint-whitespace.yml
ee7: The rationale for the current way is:
- I like to run CI by pushing to my fork (without needing to trigger via `workflow_dispatch`) before creating a PR. If I forgot to run `shellcheck` or `markdownlint` on something, it's nice to get a notification for that before creating the PR. So it can reduce initial noise in a PR, making things easier for a reviewer, and reduce the number of jobs that Exercism runs (limit of 20 concurrent jobs, I think). 
- GitHub claims a runtime of a few seconds for these workflows, so the damage to the planet is a small fraction of the other jobs in this repo.
- It's consistent with:

https://github.com/exercism/configlet/blob/4ab7a531a5968339e69d10990f0188e3c57d2e97/.github/workflows/markdownlint.yml#L3

https://github.com/exercism/configlet/blob/4ab7a531a5968339e69d10990f0188e3c57d2e97/.github/workflows/shellcheck.yml#L3

I've had some conversations about this in the past. I think the condition I want is "run on push, unless there's an open PR for that branch" - but at the time, it wasn't obvious to me how to do that. Have you ever done this? I'd be happy to change the workflows in this repo as such.

.github/bin/lint-whitespace
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env sh
+
+repo_root=$(git rev-parse --show-toplevel)
+cd "${repo_root}" || exit
```
ee7: Rationale: the current `git grep` commands only search in the working directory. CI already runs the script from the repo root dir, but it's nice if a user doesn't have to.

Another approach is `git grep -- :/`, but then the paths are printed relative to the working directory, and I'd prefer them to be relative to the repo root.

If a user's working directory is a different repo (not configlet), this script will currently check that repo instead. If we wanted to avoid that, we can instead `cd` after getting the script's location.

ee7: An alternative that I like less:
```suggestion
if [ ! -d ./.git ]; then
  echo "Error: no .git directory found. This script must be run from a repo root."
  exit 1
fi
```

.github/workflows/lint-whitespace.yml
ErikSchierboom: I've never done that. It might be possible by inspecting the `github.event` object. I'm okay with keeping it as is.

ErikSchierboom (approved):

ee7:
Tested:
```console
$ sed -i 's/Checkout/Checkout /g' .github/workflows/markdownlint.yml
$ git add .github/workflows/markdownlint.yml
$ git commit -n -m 'foo'
$ printf '\n' >> README.md
$ truncate -s -1 src/sync/sync.nim
$ .github/bin/lint-whitespace
There is trailing whitespace on the below lines:
.github/workflows/markdownlint.yml:10:      - name: Checkout 
Multiple newlines at end of file: README.md
No newline at end of file: src/sync/sync.nim
```

.github/workflows/lint-whitespace.yml
ee7: Tracking in #543.

PR #537, opened 2022-04-01
logger: refactor

ee7:
Summary:
- Clarify that `setupLogging` only needs only the verbosity information
  from `conf`.
- Improve proc name.
- Shorten a long line.
- Fix `conf` parameter name.
- Use `template` for logging routines. Before this commit, arguments to
  `logNormal` and `logDetailed` were evaluated even if the logging level
  is such that the message is not printed. Note that the underlying
  `notice` and `info` are both templates.

From the [template section of the Nim Tutorial (Part II)](https://nim-lang.org/docs/tut2.html#templates):

> Templates are especially useful for lazy evaluation purposes.
> Consider a simple proc for logging:
>
> ```nim
> const debug = true
>       
> proc log(msg: string) {.inline.} =
>   if debug: stdout.writeLine(msg)
>       
> var x = 4
> log("x has the value: " & $x)
> ```
>
> This code has a shortcoming: if `debug` is set to `false` someday,
> the quite expensive `$` and `&` operations are still performed!
> (The argument evaluation for procedures is eager).
>
> Turning the `log` proc into a template solves this problem.

### DIFF
--- a/src/configlet.nim
+++ b/src/configlet.nim
@@ -8,7 +8,7 @@ proc main =
 
   let conf = processCmdLine()
 
-  setupLogging(conf)
+  setupLogging(conf.verbosity)
 
   case conf.action.kind
   of actNil:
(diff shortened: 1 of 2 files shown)


ErikSchierboom (approved):
Nice!

PR #540, opened 2022-04-01
sync: cache prob-specs repo, and remove `--prob-specs-dir` option

### LINKED ISSUE #477: sync: cache the `problem-specifications` repo
Currently, a plain `configlet sync` always clones `problem-specifications` if the `--prob-specs-dir` option is not passed. This is simple and robust, but slow and wasteful.

Instead, configlet could re-use the cloned `problem-specifications` repo. For example, on Linux, we could download it to `~/.cache/exercism/configlet/problem-specifications`.

Then, a plain `configlet sync` can:
- Look for the local `problem-specifications`
- If found: run `git pull` in that repo if `--offline` was not passed (possibly not pulling if the last pull was very recent)
- If not found: clone it

Such caching is more useful now that `configlet sync` can update docs and metadata (see 1fd2a46fddfb). Without caching, most track maintainers would probably want to use `--offline --prob-specs-dir` all the time.

Another benefit is that `configlet info` can use the cached `problem-specifications` too. We could consider no longer baking the prob-specs slugs into the binary (see [`src/info/prob_specs_exercises.json`](https://github.com/exercism/configlet/blob/34a1299e8b7c9c0c2563e78348ccdab723ccf95c/src/info/prob_specs_exercises.json)). This is tracked in #483.

ee7:
Before this commit, the default behavior of `configlet sync` was to
clone `exercism/problem-specifications` and delete it afterwards. The
`-p, --prob-specs-dir` and `-o, --offline` options influenced the
behavior as shown in the below table, where an `x` in one of the first
two columns means the corresponding option was given.

| -p | -o | Behavior                                              |
| -- | -- | ----------------------------------------------------- |
|    |    | Clone prob-specs, and delete it afterwards            |
| x  |    | Use the given prob-specs dir; error if outdated       |
| x  | x  | Use the given prob-specs dir; don't check if outdated |
|    | x  | Error: `-p` must be provided when `-o` is given       |

This was simple and robust, but slow in the zero-option case. Most track
maintainers would probably want to use both `--offline` and
`--prob-specs-dir` nearly all the time, especially because we often want
to run several `configlet sync` commands in quick succession.

With this commit, `configlet sync` now looks for a cached
problem-specifications on the user's machine, behaving as follows:

| Cache hit? | -o | Behavior                                           |
| ---------- | -- | -------------------------------------------------- |
|            |    | Clone prob-specs, and don't delete it afterwards   |
|     x      |    | Use cached prob-specs, and `git pull` in it        |
|     x      | x  | Use cached prob-specs, and don't `git pull` in it  |
|            | x  | Error: local prob-specs not found, and won't clone |

This optimizes for the convenience of `configlet sync` and
`configlet sync --offline` - to perform an offline sync, the user no
longer needs to also write a path to a prob-specs directory.

In fact, this commit also removes the `--prob-specs-dir` option
entirely; its remaining usefulness is not worth the added complexity in
the help message (or the code).

Despite the `--prob-specs-dir` removal, note that you can still:
- Sync from an older prob-specs commit, by checking out that commit in
  the cached repo and then using `--offline`. We require the commit
  to exist on the upstream `main` branch.
- Sync using `--offline` when your only copy of problem-specifications
  is outside the cache directory (by copying to the cached location). 

The path to the cached prob-specs is:
    
    $PATHSTART/exercism/configlet/problem-specifications

where the value of `$PATHSTART` is given by this table:

| Platform | $PATHSTART                                                  |
| -------- | ----------------------------------------------------------- |
| Windows  | getEnv("LOCALAPPDATA")                                      |
| macOS    | getEnv("XDG_CACHE_HOME", getEnv("HOME") / "Library/Caches") |
| Other    | getEnv("XDG_CACHE_HOME", getEnv("HOME") / ".cache")         |

and `getEnv` returns the value of the environment variable with the name
of the first parameter (or if that does not exist, the value of the
second parameter).

For example, on Linux, the cached path is typically:

    ~/.cache/exercism/configlet/problem-specifications

Closes: #477

---

To-do:
- [x] Allow `-o, --offline` without `--prob-specs-dir`, since that now makes sense.
- [x] Remove `--prob-specs-dir`.
- [x] Automatically `git pull` in the cached dir when `--offline` is not given.
- [x] Make online sync test runnable outside CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Options for fmt:
 Options for sync:
   -e, --exercise <slug>        Only operate on this exercise
   -o, --offline                Do not update the cached 'problem-specifications' data
-  -u, --update                 Prompt to update the seen data that are unsynced
+  -u, --update                 Prompt to update the seen track data that are unsynced
   -y, --yes                    Auto-confirm prompts from --update for updating docs, filepaths, and metadata
       --docs                   Sync Practice Exercise '.docs/introduction.md' and '.docs/instructions.md' files
       --filepaths              Populate empty 'files' values in Concept/Practice exercise '.meta/config.json' files
--- a/README.md
+++ b/README.md
@@ -45,7 +45,6 @@ Global options:
   -t, --track-dir <dir>        Specify a track directory to use instead of the current directory
   -v, --verbosity <verbosity>  The verbosity of output.
                                Allowed values: q[uiet], n[ormal], d[etailed] (default: normal)
-
 ```
 
 ## `configlet lint`
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Options for fmt:
 Options for sync:
   -e, --exercise <slug>        Only operate on this exercise
   -o, --offline                Do not update the cached 'problem-specifications' data
-  -u, --update                 Prompt to update the seen track data that are unsynced
+  -u, --update                 Prompt to update the unsynced track data
   -y, --yes                    Auto-confirm prompts from --update for updating docs, filepaths, and metadata
       --docs                   Sync Practice Exercise '.docs/introduction.md' and '.docs/instructions.md' files
       --filepaths              Populate empty 'files' values in Concept/Practice exercise '.meta/config.json' files
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Options for fmt:
 
 Options for sync:
   -e, --exercise <slug>        Only operate on this exercise
-  -o, --offline                Do not check that the cached 'problem-specifications' directory is up to date
+  -o, --offline                Do not update the cached 'problem-specifications' data
   -u, --update                 Prompt to update the seen data that are unsynced
   -y, --yes                    Auto-confirm prompts from --update for updating docs, filepaths, and metadata
       --docs                   Sync Practice Exercise '.docs/introduction.md' and '.docs/instructions.md' files
--- a/README.md
+++ b/README.md
@@ -26,8 +26,7 @@ Options for fmt:
 
 Options for sync:
   -e, --exercise <slug>        Only operate on this exercise
-  -p, --prob-specs-dir <dir>   Use this 'problem-specifications' directory, rather than cloning temporarily
-  -o, --offline                Do not check that the directory specified by --prob-specs-dir is up to date
+  -o, --offline                Do not check that the cached 'problem-specifications' directory is up to date
   -u, --update                 Prompt to update the seen data that are unsynced
   -y, --yes                    Auto-confirm prompts from --update for updating docs, filepaths, and metadata
       --docs                   Sync Practice Exercise '.docs/introduction.md' and '.docs/instructions.md' files
@@ -46,6 +45,7 @@ Global options:
   -t, --track-dir <dir>        Specify a track directory to use instead of the current directory
   -v, --verbosity <verbosity>  The verbosity of output.
                                Allowed values: q[uiet], n[ormal], d[etailed] (default: normal)
+
 ```
 
 ## `configlet lint`
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -197,7 +197,7 @@ func genHelpText: string =
     optVerbosity: &"The verbosity of output.\n" &
                   &"{paddingOpt}{allowedValues(Verbosity)} (default: normal)",
     optFmtSyncExercise: "Only operate on this exercise",
-    optFmtSyncUpdate: "Prompt to update the seen data that are unsynced",
+    optFmtSyncUpdate: "Prompt to update the seen track data that are unsynced",
     optFmtSyncYes: &"Auto-confirm prompts from --{$optFmtSyncUpdate} for updating docs, filepaths, and metadata",
     optSyncOffline: "Do not update the cached 'problem-specifications' data",
     optSyncDocs: "Sync Practice Exercise '.docs/introduction.md' and '.docs/instructions.md' files",
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -197,7 +197,7 @@ func genHelpText: string =
     optVerbosity: &"The verbosity of output.\n" &
                   &"{paddingOpt}{allowedValues(Verbosity)} (default: normal)",
     optFmtSyncExercise: "Only operate on this exercise",
-    optFmtSyncUpdate: "Prompt to update the seen track data that are unsynced",
+    optFmtSyncUpdate: "Prompt to update the unsynced track data",
     optFmtSyncYes: &"Auto-confirm prompts from --{$optFmtSyncUpdate} for updating docs, filepaths, and metadata",
     optSyncOffline: "Do not update the cached 'problem-specifications' data",
     optSyncDocs: "Sync Practice Exercise '.docs/introduction.md' and '.docs/instructions.md' files",
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -199,7 +199,7 @@ func genHelpText: string =
     optFmtSyncExercise: "Only operate on this exercise",
     optFmtSyncUpdate: "Prompt to update the seen data that are unsynced",
     optFmtSyncYes: &"Auto-confirm prompts from --{$optFmtSyncUpdate} for updating docs, filepaths, and metadata",
-    optSyncOffline: "Do not check that the cached 'problem-specifications' directory is up to date",
+    optSyncOffline: "Do not update the cached 'problem-specifications' data",
     optSyncDocs: "Sync Practice Exercise '.docs/introduction.md' and '.docs/instructions.md' files",
     optSyncFilepaths: "Populate empty 'files' values in Concept/Practice exercise '.meta/config.json' files",
     optSyncMetadata: "Sync Practice Exercise '.meta/config.json' metadata values",
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -35,7 +35,6 @@ type
       yesFmt*: bool
     of actSync:
       exercise*: string
-      probSpecsDir*: string
       offline*: bool
       update*: bool
       yes*: bool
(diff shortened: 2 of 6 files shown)


ee7:
@ErikSchierboom When both `configlet sync` and `configlet sync --offline` use the cached prob-specs location, I'm thinking that the `--prob-specs-dir` option no longer provides much benefit. What do you think about removing it?

If we want to remove `--prob-specs-dir`, it would be good to do that before releasing the stable `4.0.0` - it's a breaking change that would otherwise require `5.0.0`. We can remove it from the `--help` message, but if someone uses it, output a specific error message explaining that the option was removed (and what to do instead).

ErikSchierboom:
I think removing it would be good, and indeed to be doing this before our major release is best.

ee7:
OK - please test locally.

I'm much happier with this design, given that previously I almost always used `-o -p path/to/prob-specs`.

I have some follow-up refactoring, like refactoring the tests to use the cached prob-specs, but I'll do that in a separate PR.

tests/test_binary.nim
```diff
@@ -839,6 +840,28 @@ proc testsForSync(binaryPath: static string) =
...
+        # Reset local `main` to a previous commit.
+        check git(["-C", psDir, "reset", "--hard",
+                  "0eda2318cb5622532e498559255c8fe141c9d07f"]).exitCode == 0
+
+        # Perform a sync without `--offline`.
+        execAndCheckExitCode(1, syncBase)
+
+        # Check that local HEAD and `main` point to same commit as `origin/main`.
+        let upstreamLatestRef = gitCheck(0, ["-C", psDir, "rev-parse",
+                                            "origin/main"])
```
ee7: We don't need `getNameofRemote` here, because it's guaranteed to be named `origin` during CI.

Locally, a user might override `clone.defaultRemoteName` in their git config, so pulling from `origin/main` would fail.

~I'll add a comment for this~. Edit: Actually, I'll try to make it not CI-only.

I've thought about explicitly using `git clone --origin origin` in `cloneExercismRepo`, but that might surprise someone who has specified a different `clone.defaultRemoteName` in their git config.

src/sync/sync.nim
```diff
@@ -93,41 +89,38 @@ proc syncImpl(conf: Conf): set[SyncKind] =
...
-        let exercises = toSeq findExercises(conf, probSpecsDir)
-        if conf.action.update:
-          updateTests(result, conf, exercises)
-        else:
-          checkTests(result, exercises)
-
-  finally:
-    if conf.action.probSpecsDir.len == 0 and conf.action.scope != {skFilepaths}:
-      removeDir(probSpecsDir)
+  let psExercisesDir = probSpecsDir / "exercises"
```
ErikSchierboom, line 92: Nice! Less indentation.

tests/test_binary.nim
ErikSchierboom: > I've thought about explicitly using git clone --origin origin in cloneExercismRepo, but that might surprise somebody.

Why would that surprise somebody?

README.md
```diff
@@ -26,8 +26,7 @@ Options for fmt:
 
 Options for sync:
   -e, --exercise <slug>        Only operate on this exercise
-  -p, --prob-specs-dir <dir>   Use this 'problem-specifications' directory, rather than cloning temporarily
-  -o, --offline                Do not check that the directory specified by --prob-specs-dir is up to date
+  -o, --offline                Do not check that the cached 'problem-specifications' directory is up to date
```
ErikSchierboom: I wonder if we could shorten this a bit and make it describe the functionality more than the implementation. 
Maybe something like:

```suggestion
  -o, --offline                Do not update the 'problem-specifications' information
```

or 

```suggestion
  -o, --offline                Do not update the 'problem-specifications' data
```

or

```suggestion
  -o, --offline                Do not update the 'problem-specifications' repo
```

src/cli.nim
```diff
@@ -392,6 +386,15 @@ proc parseOption(kind: CmdLineKind, key: string, val: string): Opt =
     if val.len == 0 and result notin optsNoVal and result != optSyncTests:
       showError(&"{formatOpt(kind, key)} was given without a value")
   except ValueError:
+    if keyNormalized in ["p", "probspecsdir"]:
+      const msg = """
```
ErikSchierboom, line 390: Excellent idea to catch this case!

src/sync/probspecs.nim
```diff
@@ -101,59 +102,81 @@ proc getNameOfRemote(probSpecsDir: ProbSpecsDir;
...
+      # Exit if the repo has no such remote.
+      const upstreamHost = "github.com"
+      const upstreamLocation = "exercism/problem-specifications"
+      let remoteName = getNameOfRemote(probSpecsDir, upstreamHost, upstreamLocation)
 
-    if not conf.action.offline:
-      # For now, just exit with an error if the HEAD is not up-to-date with
-      # upstream, even if it's possible to do a fast-forward merge.
+      # `fetch` and `merge` separately, for better error messages.
+      logNormal(&"Running 'git pull' in cached problem-specifications dir...")
```
ErikSchierboom: Maybe have this be more about what is happening functionally?

```suggestion
      logNormal(&"Updating problem-specifications data...")
```

Or something like that.

ErikSchierboom (approved):
It works well! Some minor comments.

tests/test_binary.nim
ee7: Because if somebody deliberately changes the `clone.defaultRemoteName` in their git config to be `foo`, they might reasonably assume that all newly cloned git repos on their machine have the remote called `foo`. It seems a little heavy-handed to override their explicit preference, even if we don't expect them to be running git commands in the cached repo.

And, thinking about this more, I think I don't want us to hard-code `git fetch origin/main` anyway. If somebody has no network connection, and hasn't run an online sync ever (or recently), they should be able to copy another local prob-specs to the cache location and have it stay working for a later online sync (even if the remote for `exercism/problem-specifications` is named e.g. `upstream`). Niche use case for a power user, sure, but it seems too fragile otherwise.

ErikSchierboom: Okay. I wouldn't mind it, but I'm fine with it

README.md
ee7: Good idea, thanks. The previous version was also bad because it only reflected the old behavior of "is it out of date?" rather than "pull from prob-specs".

I pushed this:

```
  -o, --offline                Do not update the cached 'problem-specifications' data
```

where maybe the word "cached" helps convey that `--offline` won't work if you don't have a local prob-specs.

But I don't like how "update" looks next to the rest of the help text:

```
Options for sync:
  -e, --exercise <slug>        Only operate on this exercise
  -o, --offline                Do not update the cached 'problem-specifications' data
  -u, --update                 Prompt to update the seen data that are unsynced
  -y, --yes                    Auto-confirm prompts from --update for updating docs, filepaths, and metadata
      --docs                   Sync Practice Exercise '.docs/introduction.md' and '.docs/instructions.md' files
      --filepaths              Populate empty 'files' values in Concept/Practice exercise '.meta/config.json' files
      --metadata               Sync Practice Exercise '.meta/config.json' metadata values
      --tests [mode]           Sync Practice Exercise '.meta/tests.toml' files.
                               The mode value specifies how missing tests are handled when using --update.
                               Allowed values: c[hoose], i[nclude], e[xclude] (default: choose)
```

where the `--update` text of "prompt to update the seen data" is now confusing - it might imply that prob-specs isn't updated unless `--update` is passed, which is wrong.

I've pushed a commit that changes the `--update` text to "seen track data", but it still doesn't seem great. Maybe we could avoid the word "update" in `--offline`, or the word "update" in `--update`? Thoughts?

One more extreme option is to change `--update` to `--write`...

src/sync/probspecs.nim
ee7: Pushed 9ed7f204e026 with extra "cached", like the help message. Let me know if you hate it.

README.md
```diff
@@ -26,8 +26,8 @@ Options for fmt:
 
 Options for sync:
   -e, --exercise <slug>        Only operate on this exercise
-  -o, --offline                Do not check that the cached 'problem-specifications' directory is up to date
-  -u, --update                 Prompt to update the seen data that are unsynced
+  -o, --offline                Do not update the cached 'problem-specifications' data
+  -u, --update                 Prompt to update the seen track data that are unsynced
```
ErikSchierboom: What does "seen" mean here?

```suggestion
  -u, --update                 Prompt to update the unsynced track data
```

ee7: "prompt to update the unsynced track data" might sound like it always operates on every kind of unsynced data. The "seen" tries to convey that it respects the `--docs`, `--filepaths`, `--metadata`, and `--tests` options - otherwise a user might think those only work without `--update`.

It doesn't help that the help descriptions are like

```
--metadata               Sync Practice Exercise '.meta/config.json' metadata values
```

where it says just "Sync". In general there's a lot of potential confusion from "sync" vs "update".

ErikSchierboom: Ah. I didn't quite get that, but I must admit that I don't know a better option.

ee7: Hmm. Thoughts on this?

```
  -u, --update                 Prompt to update unsynced track data (default: docs, filepaths, metadata, and tests)
```

This tries to convey what "seen" tried to, and it would also document the default syncing scope in the help message.

ee7: Actually, I don't love that either. That's the default syncing scope even without update, so it should be written somewhere else.

ErikSchierboom: Agreed. What about:

> Prompt to update selected unsynced track data

But that implies a selection to be made

ee7: Yeah. 

Let's go with

> "Prompt to update the unsynced track data"

for now, and try to improve later? Maybe nobody notices the problems here as much as us.

ErikSchierboom: 👍  The latter is almost certainly true :)

PR #554, opened 2022-04-13
info(prob_specs_exercises): add `state-of-tic-tac-toe`

ee7:
Upstream: https://github.com/exercism/problem-specifications/commit/3e1cfcbaa6fb

PR #555, opened 2022-04-13
info: add script for writing prob-specs JSON file

ee7:
Running this script will update `src/info/prob_specs_exercises.json`
locally. I ran this script to produce the diff in #554.

We will no longer need this script if we alter `configlet info` to use
the new prob-specs cache, rather than baked-in prob-specs data. But it's
worth having the script in the git history, even if we delete it later.

For background, see commit b8432f601490 (which added the initial JSON
file, created by an older version of this script).

### DIFF
--- a/bin/write_probspecs_info.nim
+++ b/bin/write_probspecs_info.nim
@@ -59,15 +59,20 @@ proc init(T: typedesc[ProbSpecsState], probSpecsDir: string): T =
 
 proc main =
   const repoRootDir = currentSourcePath().parentDir().parentDir()
-  const probSpecsDir = repoRootDir.parentDir() / "problem-specifications"
+  const probSpecsDir = getCacheDir() / "exercism" / "configlet" / "problem-specifications"
   if dirExists(probSpecsDir):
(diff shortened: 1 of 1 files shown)


bin/write_probspecs_info.nim
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env -S nim r --verbosity:0 --skipParentCfg:on
+import std/[algorithm, json, os, osproc, strformat, strutils, times]
+import pkg/jsony
+
```
ee7, line 4: I put the script in `bin/`, rather than `src/info/` - I think in `bin/` it's a little clearer that the file is a script for generating a JSON file, rather than a `.nim` file that's compiled when calling `nimble build`.

But I'm OK with putting it in `src/info/` if you prefer.

ErikSchierboom: No, `bin/` is fine, as we're using that in many repos.

ErikSchierboom (commented):
> [...] if configlet info is altered to use the prob-specs cache rather than baked-in prob-specs data. 

That's a great idea!

ErikSchierboom (approved):

PR #585, opened 2022-05-05
lint(exercises): check `files.editor` field for exercises

ErikSchierboom:
- lint(validators): allow optional array of files
- lint(exercises): check the validity of the `files.editor` field

### DIFF
--- a/src/lint/validators.nim
+++ b/src/lint/validators.nim
@@ -420,24 +420,31 @@ proc hasArrayOfFiles*(data: JsonNode;
                       path: Path;
                       context = "";
                       relativeToPath: Path;
-                      errorAnnotation = ""): bool =
-  if hasArrayOfStrings(data, key, path, context):
-    result = true
-    var processedItems = initHashSet[string](data[key].len)
+                      errorAnnotation = "";
+                      isRequired = true): bool =
+  ## Returns true in any of these cases:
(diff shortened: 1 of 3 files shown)


src/lint/validators.nim
```diff
@@ -420,24 +420,31 @@ proc hasArrayOfFiles*(data: JsonNode;
...
-        let itemStr = item.getStr()
-        if processedItems.containsOrIncl(itemStr):
-          let msg = &"The {q context} array contains duplicate " &
-                    &"{q itemStr} values"
+                      errorAnnotation = "";
+                      isRequired = true): bool =
+  ## Returns true in any of these cases:
+  ## - `hasArrayOfStrings` returns true for `data[key]`.
+  ## - `data` lacks the key `key` and `isRequired` is false.
+  if data.hasKey(key, path, context, isRequired):
```
ee7, line 428: Do we need this line and extra indendation level? Can we use the below?

```Nim
if hasArrayOfStrings(data, key, path, context, isRequired = isRequired)
```

ErikSchierboom: No, because that'll return `true` if the field is not required and not present, which will cause the subsequent `data[key]` call to fail.

ee7 (approved):

Issue #556, opened 2022-04-14
ci: diff output from `lint` and `sync` with previous configlet release


ee7:
Follow-up from https://github.com/exercism/configlet/issues/523#issuecomment-1099162786

I have a script for this, which I've run for all the configlet releases. But it probably needs some tidying up before it works in this repo.

---

Edit 2023-10-02:
I think we can consider this part of the integration tests, but a separate job that's allowed to fail (where it indicates failure if the PR causes any configlet command to produce different output on any track).

And maybe for PRs only (not on push).

PR #589, opened 2022-05-06
docs: document configlet release process

### LINKED ISSUE #523: Document release process
We should document how to build a new release. Ideally, we'd be able to use https://github.com/exercism/configlet/blob/main/bin/release

ee7:
Closes: #523

---

@ErikSchierboom Now that you've gone through the release process, you can review this better :)

I've tried to expand on what I wrote in https://github.com/exercism/configlet/issues/523#issuecomment-1099162786.

To-do:

- [x] Bikeshed the filename
- [x] Explain what to do if a problem is found only after publishing the release. If I recall correctly, a published release cannot be set as draft again, but it can be marked as a pre-release. And that means that it would no longer be downloaded by a configlet CI job, or `fetch-configlet`.
- [x] Consider what we say about force-pushing.


docs/maintaining-configlet.md
```diff
@@ -0,0 +1,76 @@
...
+no longer want to create a release at this time, you can either:
+
+- Delete the draft release, then delete the most recent tag both remotely and
+  locally, and then force-push to `main` to remove exactly the most recent
+  commit (the release commit)
+
+- Or create a PR to fix the problem, merge it, and then begin the release
+  process again. In this case you should not delete the problematic tag, and
+  either delete the problematic release without un-drafting it, or keep it as a
+  draft and edit its release notes to say that it should not be used.
```
ee7: We haven't needed to do this yet. 

We may or may not be comfortable with the first option. It's possible to "get away with it" for low-traffic repos where it's unlikely that somebody fetches the bad commit in the (hopefully short) time period before a force-push (somebody who does receive the bad commit would get an error during a later `git pull`). But we might like to suggest a preference for one or the other.

But I think we should mention the first as an option even if we recommend against it: otherwise, a reader in the distant future might think "oh, I'll just force-push instead", but then forget to delete the release/tag. That's not a disaster, but I don't know what happens to a (draft) GitHub release after the corresponding commit is no longer on `main` - it might increase the chance that somebody fetches "the bad release".

Please read https://git-scm.com/docs/git-tag#_on_re_tagging

docs/maintaining-configlet.md
```diff
@@ -0,0 +1,76 @@
...
+
+> Remaining steps to release:
+>
+> 1. Edit the release notes to contain the list of user-facing changes,
+>    separating by features and bug fixes.
+> 2. Wait for every build job to finish.
+> 3. Check that CI is green.
+> 4. Un-draft the release.
+
+where un-drafting the release means that, from the moment on, it is used in any
```
ErikSchierboom:
```suggestion
where un-drafting the release means that, from that moment on, it is used in any
```

docs/maintaining-configlet.md
ErikSchierboom: > We may or may not be comfortable with the first option.

I'd argue against it. We don't do force pushes anywhere, and I'd rather not make an exception for configlet.
Can't we combine parts of the two suggestions, where we a) delete the release b) create a PR to fix things

ErikSchierboom (commented):
> Feel free to bikeshed the filename, etc.

As everything in this document has to do with releases, I'd prefer the filename to reflect that. Maybe `docs/releases.md` or `docs/publish-releases.md` or something like that.

docs/maintaining-configlet.md
ee7: OK. How about https://github.com/exercism/configlet/pull/589/commits/74ab3b5f3eed1b61a81f2366d6237a259442bbbf?

I realise that it'd be easier to just not mention force-pushing. But I think there are situations (which ideally would be prevented by tests and code review) where it's the least bad option.

Configlet is a bit different from most other active Exercism repos:

- A merged commit in configlet is not immediately deployed
- There are releases
- There are release binaries
- The changes in the configlet repo can be distributed to contributors to every track repo

But is what `releases.md` currenty says about force-pushing actually making an exception for configlet? Or is it actually just discussing a problem that we haven't written a policy about anywhere?

For example, let's say in `exercism/python` we accidentally merge a PR that tries to improve the exercise-checking script. But it turns out that on say, macOS, it means that the script deletes every file from the user's home directory. We can revert that commit, but there isn't a good way to warn every contributor, for all time, never to checkout the bad commit, and that `git bisect` could be dangerous.

Here I'm recalling notorious bugs in:

- [Steam](https://github-redirect.dependabot.com/ValveSoftware/steam-for-linux/issues/3671)
- [Bumblebee](https://github-redirect.dependabot.com/MrMEEE/bumblebee-Old-and-abbandoned/issues/123)
- [Myth II](https://minimaxir.com/2013/06/working-as-intended/)
- [Sierra Utilities](https://arstechnica.com/civis/viewtopic.php?t=479484)

It's just that this `releases.md` should try to help when things go wrong. So I'm trying to mention ways that things can go wrong.

docs/releases.md
```diff
@@ -0,0 +1,98 @@
...
+If we have already published a release, and want to investigate a problem in it, we should mark the release as a prerelease.
+Similar to a draft release, a release marked as a prerelease will not be downloaded by a configlet CI job, or `fetch-configlet`.
+Note that a published release cannot be set as draft again.
+If it turns out that we want the release to be available again, simply undo the marking of the release as prerelease.
+If it turns out that we want to remove the release, follow the steps in the section above to create a new release.
+
+### Should we force-push to `main` to remove the commit of a cancelled release, or a problematic commit?
+
+In general, we do not force-push to `main` in Exercism repositories.
+Force-pushing to `main` is reserved for exceptional circumstances where the benefits outweigh the inconvenience to the user who fetched a later-removed commit or tag, and to the maintainer that must temporarily disable the branch protection rule.
```
ee7: As an aside, as of last week, it is correct to say that a non-admin can bypass branch protection rules:
 
https://github.blog/changelog/2022-08-18-bypass-branch-protections-with-a-new-permission/

docs/maintaining-configlet.md
ee7: That said, I don't think it's important to try to document a process here. I'm also fine with removing the section entirely: https://github.com/exercism/configlet/pull/589/commits/a1a07e2e0f74b795d41b03e573360148bde8510c

Good enough for now?

ee7:
Ping @ErikSchierboom, in case this PR got lost. No rush.

docs/releases.md
```diff
@@ -0,0 +1,91 @@
+# Configlet releases
+
+## How to create a configlet release
+
+1. Check that the repo state is such that creating a release PR makes sense.
+
+1. Run [`bin/bump_version.nim`][bump_version], which also prompts to create a release PR if you have [`gh`][gh] installed on your machine.
+
+1. Ideally, check that `configlet lint` and `configlet sync` do not produce unexpected output on any track - diffing the output from the previous release.
```
ErikSchierboom, line 9: It would be great to have a script for this (checking for any changes in output)

docs/releases.md
```diff
@@ -0,0 +1,91 @@
...
+
+1. Double-check that the repo is in a state such that a release makes sense.
+   Remember that we do not usually force-push to `main`, so any release commit (and tag) will remain permanently.
```
ErikSchierboom: I'd prefer this to be stronger:

```suggestion
   Remember that we do not support force-pushing to `main`, so any release commit (and tag) will remain permanently.
```

ErikSchierboom (approved):
Awesome work. Just a couple of tiny nits.

docs/releases.md
ee7: Yeah. Tracking in https://github.com/exercism/configlet/issues/556.

docs/releases.md
ee7: Done: https://github.com/exercism/configlet/pull/589/commits/010240c2d857ea951580f8eeb9ce44d50dd8a3c4

ee7:
I held off merging this to remind myself to upload legacy release assets, when we were still doing that.

PR #596, opened 2022-05-11
info: refactor

ee7:
I find this more readable overall. See the commit messages for more details.

I've checked that this PR keeps the output of `configlet info` the same on every track.

---

Summary:
- Inline the procedures that are only called once, which can reduce
  cognitive load when reading. We can refactor to common procs later for
  functionality that we need more than once (e.g. in another command).
- Start using `sugar.collect` - it also helps the above inlining.
- Remove a TODO: `configlet info` will soon use the prob-specs cache.
- Remove unneeded generic in `show[A]`. This makes it immediately
  obvious that we only operate on `HashSet[string]`, and fixes the doc
  comment - it said that elements are returned "in alphabetical order",
  which wouldn't be true for e.g. `HashSet[int]`.
- Use a new `toStringSorted` common proc.
- Avoid `s` as name of `HashSet` parameter - reserve `s` for `string`.
- Move type and `init` declarations: make declaration order match usage
  order.
- Move `probSpecsExercises` declaration to the innermost possible scope.

### DIFF
--- a/src/info/info.nim
+++ b/src/info/info.nim
@@ -129,8 +129,9 @@ proc info*(conf: Conf) =
   let trackConfigPath = conf.trackDir / "config.json"
 
   if fileExists(trackConfigPath):
-    let trackConfigContents = readFile(trackConfigPath)
-    let trackConfig = TrackConfig.init(trackConfigContents)
+    let trackConfig = block:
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
LGTM

PR #605, opened 2022-05-17
info: use prob specs cache

### LINKED ISSUE #483: info: use a (cached) `problem-specifications` repo

ee7:
This stops baking the prob-specs data into the binary, and makes
`configlet info` use the cached prob-specs directory like
`configlet sync`.

Advantages:
- The prob-specs data is updated at run time; we no longer need
  commits like https://github.com/exercism/configlet/commit/b54fe48f72908727eac1c5f9bdf6b79256464d00 or https://github.com/exercism/configlet/commit/146d41c184b5bccef117a91cdc56c3691a20aede, and can remove
`prob_specs_exercises.json`
- The size of the configlet executable is reduced by 12 KiB (2.5%)
- Fewer lines of code: we can remove `bin/write_probspecs_info.nim`

Disadvantages:
- The user must now pass `-o` or `---offline` when running
  `configlet info` with no/limited network connectivity.
- `configlet info` is slower, even when using `-o`, because we run a few
  `git` commands to validate the cache repo.

Closes: #483


ErikSchierboom (approved):
Approving the changes in https://github.com/exercism/configlet/commit/15ac68d98bd02d11803efcd9bf22b7e65c47fa89

src/info/info.nim
```diff
@@ -1,6 +1,6 @@
 import std/[algorithm, os, sequtils, sets, strformat, strutils, sugar, terminal]
 import pkg/jsony
-import ".."/[cli, types_track_config]
+import ".."/[cli, sync/probspecs, types_track_config]
```
ee7, line 3: We could refactor later so that the types/procedures for cloning and validating prob-specs are in a common location.

src/cli.nim
```diff
@@ -33,6 +33,8 @@ type
       exerciseFmt*: string
       updateFmt*: bool
       yesFmt*: bool
+    of actInfo:
+      offlineInfo*: bool
     of actSync:
       exercise*: string
       offline*: bool
```
ee7, line 40: `offline` and `offlineInfo` is consistent with e.g. `exercise` and `exerciseFmt`. But maybe it's better to move to `offlineSync` and `exerciseSync` later.

As a reminder, they can't both be called `offline` unless we make `--offline` a global flag. The relevant RFC is [nim-lang/RFCs#368](https://github-redirect.dependabot.com/nim-lang/RFCs/issues/368).

src/sync/probspecs.nim
```diff
@@ -104,6 +104,10 @@ proc getNameOfRemote*(probSpecsDir: ProbSpecsDir;
   showError(&"there is no remote that points to '{location}' at '{host}' in " &
             &"the cached problem-specifications directory: '{probSpecsDir}'")
 
+func isOffline(conf: Conf): bool =
+  (conf.action.kind == actSync and conf.action.offline) or
+      (conf.action.kind == actInfo and conf.action.offlineInfo)
```
ee7, line 109: Similarly, more clunky than usual. It'd be nice if something like this were possible:

```Nim
func isOffline(conf: Conf): bool =
  conf.action.kind in {actInfo, actSync} and conf.action.offline
```

or just 

```Nim
if conf.action.offline
```

ee7 (commented):
I've checked that, with this PR, the output of `configlet info -o` is the same on every track as that with `configlet info` before this PR.

src/cli.nim
ErikSchierboom: Still waiting for the PR to implement that RFC...

src/sync/probspecs.nim
ErikSchierboom: > if conf.action.offline

That would be sweet. We'll just have to wait I guess 🤷

PR #614, opened 2022-06-07
sync(tests): improve error for invalid TOML

### LINKED ISSUE #613: sync: crashes on TOML files but lint shows no problems
Encountered this using configlet version `4.0.0-beta.4`
I run the command `configlet sync --tests --docs --metadata --filepaths` and it crashes with the following sort of error:

```
Checking exercises...
parsetoml.nim(908)       parseValue
Error: unhandled exception: /Users/mark/SRC/exercism/common-lisp/exercises/practice/affine-cipher/.meta/tests.toml(6:16) unexpected character "e" [TomlError]
```

This is because my TOML file has lines like: `description = blah blah` rather than `description = "blah blah"`.

But `configlet lint` reports no problems. 

1. Perhaps sync should not *crash* in this case?
2. Perhaps lint could signal that the TOML file is invalid?

ee7:
A `tests.toml` file might contain invalid TOML. For example:

```toml
[2ee1d9af-1c43-416c-b41b-cefd7d4d2b2a]
description = encode yes
```

where `encode yes` is [invalid because it is unquoted](https://toml.io/en/v1.0.0#string).

Before this commit, `configlet sync` would not handle the TOML parsing
exception:

```console
$ cd /tmp
$ git clone --quiet https://github.com/exercism/common-lisp
$ cd common-lisp
$ git checkout f521b1fb0f04ffc2d5baa6cf0bba37c231cc1bd7
$ bin/fetch-configlet
$ bin/configlet sync --tests
Updating cached 'problem-specifications' data...
Checking exercises...
parsetoml.nim(908)       parseValue
Error: unhandled exception: /tmp/common-lisp/exercises/practice/affine-cipher/.meta/tests.toml(6:16) unexpected character "e" [TomlError]
```

With this commit, configlet still exits immediately, but handles the
exception:

```console
$ configlet sync --tests
Updating cached 'problem-specifications' data...
Checking exercises...

Error: A 'tests.toml' file contains invalid TOML:
/tmp/common-lisp/exercises/practice/affine-cipher/.meta/tests.toml(6:16) unexpected character "e"

The expected 'tests.toml' format is documented in
https://exercism.org/docs/building/configlet/sync#h-tests
```

Note that `configlet lint` does not yet lint `tests.toml` files, and so
invalid TOML does not yet cause `configlet lint` to indicate an error.

Closes: #613

### DIFF
--- a/src/sync/tracks.nim
+++ b/src/sync/tracks.nim
@@ -77,6 +77,9 @@ proc init(T: typedesc[PracticeExerciseTests], testsPath: string): T =
           The expected 'tests.toml' format is documented in
           https://exercism.org/docs/building/configlet/sync#h-tests""".unindent()
         quit 1
+      except CatchableError:
+        stderr.writeLine "Error: " & getCurrentExceptionMsg()
+        quit 1
 
     for uuid, val in tests.getTable():
       if val.hasKey("include"):
--- a/src/sync/tracks.nim
+++ b/src/sync/tracks.nim
@@ -1,4 +1,4 @@
-import std/[algorithm, json, os, sets]
+import std/[algorithm, json, os, sets, strformat, strutils]
 import pkg/parsetoml
 import ".."/cli
 
@@ -65,7 +65,18 @@ proc init(T: typedesc[PracticeExerciseTests], testsPath: string): T =
   ## included and excluded test case UUIDs.
   result = PracticeExerciseTests.init()
   if fileExists(testsPath):
-    let tests = parsetoml.parseFile(testsPath)
+    let tests =
+      try:
+        parsetoml.parseFile(testsPath)
+      except TomlError: # Note that `TomlError` inherits from `Defect`.
+        stderr.writeLine fmt"""
+
+          Error: A 'tests.toml' file contains invalid TOML:
+          {getCurrentExceptionMsg()}
+
+          The expected 'tests.toml' format is documented in
+          https://exercism.org/docs/building/configlet/sync#h-tests""".unindent()
+        quit 1
 
     for uuid, val in tests.getTable():
       if val.hasKey("include"):


src/sync/tracks.nim
```diff
@@ -65,7 +65,18 @@ proc init(T: typedesc[PracticeExerciseTests], testsPath: string): T =
   ## included and excluded test case UUIDs.
   result = PracticeExerciseTests.init()
   if fileExists(testsPath):
-    let tests = parsetoml.parseFile(testsPath)
+    let tests =
+      try:
+        parsetoml.parseFile(testsPath)
+      except TomlError: # Note that `TomlError` inherits from `Defect`.
```
ee7, line 71: Definition: https://github.com/NimParsers/parsetoml/blob/609e03e0a7d6/src/parsetoml.nim#L112

I believe this is a bug in `parsetoml` - I'll suggest a fix in a PR there.

Defects should not be caught. See e.g. https://nim-lang.github.io/Nim/manual.html#exception-handling-exception-hierarchy

ErikSchierboom: Yep, that looks like a bug.

ErikSchierboom (approved):

ee7:
We should also do the same in the other places that we parse TOML:

https://github.com/exercism/configlet/blob/5a2937df47fa6c9917838f1b892187bdd5f41425/src/sync/sync_metadata.nim#L21-L32

https://github.com/exercism/configlet/blob/5a2937df47fa6c9917838f1b892187bdd5f41425/src/sync/exercises.nim#L146-L148

But I'm going to leave those till later because:
- The first will produce an unhandled exception if the user's cached prob-specs dir contains an invalid `metadata.toml` file, or if the values are of the wrong type. The prob-specs CI doesn't currently lint `metadata.toml` files, but I've just opened [exercism/problem-specifications#2061](https://github-redirect.dependabot.com/exercism/problem-specifications/issues/2061).
- The second won't currently produce an unhandled exception for an invalid `tests.toml` file (because the error is raised first in the code path touched by this PR).

ErikSchierboom (approved):

PR #620, opened 2022-06-17
generate: demote headings, and group link reference definitions

### LINKED ISSUE #328: generate: reduce level of other headings by 1

### LINKED ISSUE #626: generate: does not combine link reference definitions at the bottom
Consider this `.md.tpl` file:

```markdown
# Introduction

## Foo

%{concept:foo}

## Bar

%{concept:bar}
```

When more than one placeholder had a link reference definition, `configlet generate` would produce something like:

```markdown
# Introduction

## Foo

Here is a line with a link [Foo][foo].

[foo]: http://www.example.com

## Bar

Here is another line with a link [Bar][bar].

[bar]: http://www.example.com
```

That is, it would not combine link reference definitions at the bottom. It should do so - ordering by first usage, and deduplicating them:

```markdown
# Introduction

## Foo

Here is a line with a link [Foo][foo].

## Bar

Here is another line with a link [Bar][bar].

[foo]: http://www.example.com
[bar]: http://www.example.com
```

ee7:
To-do:
- [x] Decide on desired behavior.
- [x] Implement it.
- [x] Move reference links to the bottom, ordering by usage, and deduplicating them.
- [x] Support `~~~` code blocks?

As an example of something that isn't supported, GFM multi-line block quotes:

```markdown
>>>
This line is part of a multi-line block quote
### This is not a heading in GFM
>>>
```

Closes: #328
Closes: #626

---

This PR will also demote the level of a line that lacks a whitespace after the final `#` character, like:

```markdown
###Some heading
```

But previously we only removed a first top-level heading if the separating space was present.

Edge case that I don't expect to be important: the current implementation will demote a level-6 heading (beginning with `######`) to "a level-7 heading". But a line beginning with 7 `#` characters is actually no longer a heading.

### DIFF
--- a/src/generate/generate.nim
+++ b/src/generate/generate.nim
@@ -5,8 +5,8 @@ proc getConceptSlugLookup(trackDir: Path): Table[string, string] =
   ## Returns a `Table` that maps each concept's `slug` to its `name`.
   let concepts = TrackConfig.init(readFile(trackDir / "config.json")).concepts
   result = initTable[string, string](concepts.len)
-  for `concept` in concepts:
-    result[`concept`.slug] = `concept`.name
+  for con in concepts:
+    result[con.slug] = con.name
 
 func alterHeadings(s: string, linkDefs: var seq[string], h2 = ""): string =
   result = newStringOfCap(s.len)
--- a/src/generate/generate.nim
+++ b/src/generate/generate.nim
@@ -16,7 +16,8 @@ proc writeError(description: string, path: Path) =
   stderr.writeLine(path)
   stderr.write "\n"
 
-func alterHeaders(s: string, title: string, headerLevel: int): string =
+func alterHeaders(s: string, title: string, headerLevel: int,
+                  links: var seq[string]): string =
   # Markdown implementations differ on whether a space is required after the
   # final '#' character that begins the header.
   result = newStringOfCap(s.len)
@@ -40,6 +41,13 @@ func alterHeaders(s: string, title: string, headerLevel: int): string =
         let demotionAmount = if headerLevel in [1, 2]: 1 else: headerLevel - 1
         for _ in 1..demotionAmount:
           result.add '#'
+      elif s.continuesWith("[", i+1):
+        let j = s.find("]:", i+2)
+        if j > i+2 and j < s.find('\n', i+2):
+          var line = ""
+          i += s.parseUntil(line, '\n', i+1)
+          if line notin links:
+            links.add line
       elif s.continuesWith("```", i+1):
         inFencedCodeBlock = not inFencedCodeBlock
       elif s.continuesWith("~~~", i+1):
@@ -52,17 +60,21 @@ func alterHeaders(s: string, title: string, headerLevel: int): string =
   strip result
 
 proc conceptIntroduction(trackDir: Path, slug: string, title: string,
-                         templatePath: Path, headerLevel: int): string =
+                         templatePath: Path, headerLevel: int,
+                         links: var seq[string]): string =
   ## Returns the contents of the `introduction.md` file for a `slug`, but:
   ## - Without a first top-level header.
   ## - Adding a starting a second-level header containing `title`.
   ## - Demoting the level of any other header.
   ## - Without any leading/trailing whitespace.
+  ## - Without any reference links.
+  ##
+  ## Appends reference links to `links`.
   let conceptDir = trackDir / "concepts" / slug
   if dirExists(conceptDir):
     let path = conceptDir / "introduction.md"
     if fileExists(path):
-      result = path.readFile().alterHeaders(title, headerLevel)
+      result = path.readFile().alterHeaders(title, headerLevel, links)
     else:
       writeError(&"File {path} not found for concept '{slug}'", templatePath)
       quit(1)
@@ -80,6 +92,7 @@ proc generateIntroduction(trackDir: Path, templatePath: Path,
 
   var i = 0
   var headerLevel = 1
+  var links = newSeq[string]()
   while i < content.len:
     var conceptSlug = ""
     # Here, we implement the syntax for a placeholder as %{concept:some-slug}
@@ -91,7 +104,7 @@ proc generateIntroduction(trackDir: Path, templatePath: Path,
       if conceptSlug in slugLookup:
         let title = slugLookup[conceptSlug]
         result.add conceptIntroduction(trackDir, conceptSlug, title,
-                                       templatePath, headerLevel)
+                                       templatePath, headerLevel, links)
       else:
         writeError(&"Concept '{conceptSlug}' does not exist in track config.json",
                    templatePath)
@@ -101,6 +114,11 @@ proc generateIntroduction(trackDir: Path, templatePath: Path,
         headerLevel = content.skipWhile({'#'}, i+1)
       result.add content[i]
       inc i
+  if links.len > 0:
+    result.add '\n'
+    for link in links:
+      result.add link
+      result.add '\n'
 
 proc generate*(conf: Conf) =
   ## For every Concept Exercise in `conf.trackDir` with an `introduction.md.tpl`
--- a/src/generate/generate.nim
+++ b/src/generate/generate.nim
@@ -10,20 +10,36 @@ proc writeError(description: string, path: Path) =
   stderr.writeLine(path)
   stderr.write "\n"
 
+func demoteHeaders(s: string): string =
+  ## Demotes the level of any Markdown header in `s` by one. Supports only
+  ## headers that begin with a `#` character.
+  # Markdown implementations differ on whether a space is required after the
+  # final '#' character that begins the header.
+  result = newStringOfCap(s.len)
+  var i = 0
+  while i < s.len:
+    result.add s[i]
+    if s[i] == '\n' and i+1 < s.len and s[i+1] == '#':
+      result.add '#'
+    inc i
+
 proc conceptIntroduction(trackDir: Path, slug: string,
                          templatePath: Path): string =
-  ## Returns the contents of the `introduction.md` file for a `slug`, but
-  ## without any top-level heading, and without any leading/trailing whitespace.
+  ## Returns the contents of the `introduction.md` file for a `slug`, but:
+  ## - Without a first top-level header.
+  ## - Demoting the level of any other header.
+  ## - Without any leading/trailing whitespace.
   let conceptDir = trackDir / "concepts" / slug
   if dirExists(conceptDir):
     let path = conceptDir / "introduction.md"
     if fileExists(path):
       result = readFile(path)
       var i = 0
-      # Strip the top-level heading (if any)
+      # Strip the top-level header (if any)
       if scanp(result, i, *{' ', '\t', '\v', '\c', '\n', '\f'}, "#", +' ',
                +(~'\n')):
         result.setSlice(i..result.high)
+      result = demoteHeaders(result)
       strip result
     else:
       writeError(&"File {path} not found for concept '{slug}'", templatePath)
--- a/src/generate/generate.nim
+++ b/src/generate/generate.nim
@@ -8,8 +8,7 @@ proc getConceptSlugLookup(trackDir: Path): Table[string, string] =
   for `concept` in concepts:
     result[`concept`.slug] = `concept`.name
 
-func alterHeadings(s: string, h2: string, headingLevel: int,
-                   linkDefs: var seq[string]): string =
+func alterHeadings(s: string, linkDefs: var seq[string], h2 = ""): string =
   result = newStringOfCap(s.len)
   var i = 0
   i += s.skipWhitespace()
@@ -19,7 +18,7 @@ func alterHeadings(s: string, h2: string, headingLevel: int,
   # For now, support only spaces.
   if s.continuesWith("# ", i):
     i += s.skipUntil('\n', i)
-  if headingLevel == 1:
+  if h2.len > 0:
     result.add &"## {h2}"
   # Demote other headings.
   var inFencedCodeBlock = false
@@ -31,9 +30,7 @@ func alterHeadings(s: string, h2: string, headingLevel: int,
       # Add a '#' to a line that begins with '#', unless inside a code or HTML block.
       if s.continuesWith("#", i+1) and not (inFencedCodeBlock or
                                             inFencedCodeBlockTildes or inCommentBlock):
-        let demotionAmount = if headingLevel in [1, 2]: 1 else: headingLevel - 1
-        for _ in 1..demotionAmount:
-          result.add '#'
+        result.add '#'
       elif s.continuesWith("[", i+1):
         let j = s.find("]:", i+2)
         if j > i+2 and j < s.find('\n', i+2):
@@ -61,9 +58,8 @@ proc writeError(description: string, path: Path) =
   stderr.writeLine(path)
   stderr.write "\n"
 
-proc conceptIntroduction(trackDir: Path, slug: string, h2: string,
-                         templatePath: Path, headingLevel: int,
-                         linkDefs: var seq[string]): string =
+proc conceptIntroduction(trackDir: Path, slug: string, templatePath: Path,
+                         linkDefs: var seq[string], h2 = ""): string =
   ## Returns the contents of the `introduction.md` file for a `slug`, but:
   ## - Without a first top-level heading.
   ## - Adding a starting a second-level heading containing `h2`.
@@ -76,7 +72,7 @@ proc conceptIntroduction(trackDir: Path, slug: string, h2: string,
   if dirExists(conceptDir):
     let path = conceptDir / "introduction.md"
     if fileExists(path):
-      result = path.readFile().alterHeadings(h2, headingLevel, linkDefs)
+      result = path.readFile().alterHeadings(linkDefs, h2)
     else:
       writeError(&"File {path} not found for concept '{slug}'", templatePath)
       quit(1)
@@ -104,9 +100,9 @@ proc generateIntroduction(trackDir: Path, templatePath: Path,
              "%{", *{' '}, "concept", *{' '}, ':', *{' '},
              +{'a'..'z', '-'} -> conceptSlug.add($_), *{' '}, '}'):
       if conceptSlug in slugLookup:
-        let h2 = slugLookup[conceptSlug]
-        result.add conceptIntroduction(trackDir, conceptSlug, h2,
-                                       templatePath, headingLevel, linkDefs)
+        let h2 = if headingLevel == 2: "" else: slugLookup[conceptSlug]
+        result.add conceptIntroduction(trackDir, conceptSlug, templatePath,
+                                       linkDefs, h2)
       else:
         writeError(&"Concept '{conceptSlug}' does not exist in track config.json",
                    templatePath)
--- a/src/generate/generate.nim
+++ b/src/generate/generate.nim
@@ -7,15 +7,6 @@ proc getConceptSlugLookup(trackDir: Path): Table[string, string] =
   for `concept` in concepts:
     result[`concept`.slug] = `concept`.name
 
-proc writeError(description: string, path: Path) =
-  let descriptionPrefix = description & ":"
-  if colorStderr:
-    stderr.styledWriteLine(fgRed, descriptionPrefix)
-  else:
-    stderr.writeLine(descriptionPrefix)
-  stderr.writeLine(path)
-  stderr.write "\n"
-
 func alterHeadings(s: string, title: string, headingLevel: int,
                    links: var seq[string]): string =
   # Markdown implementations differ on whether a space is required after the
@@ -58,6 +49,15 @@ func alterHeadings(s: string, title: string, headingLevel: int,
       inCommentBlock = false
     inc i
   strip result
+
+proc writeError(description: string, path: Path) =
+  let descriptionPrefix = description & ":"
+  if colorStderr:
+    stderr.styledWriteLine(fgRed, descriptionPrefix)
+  else:
+    stderr.writeLine(descriptionPrefix)
+  stderr.writeLine(path)
+  stderr.write "\n"
 
 proc conceptIntroduction(trackDir: Path, slug: string, title: string,
                          templatePath: Path, headingLevel: int,
--- a/src/generate/generate.nim
+++ b/src/generate/generate.nim
@@ -1,7 +1,7 @@
 import std/[parseutils, strbasics, strformat, strscans, strutils, tables, terminal]
 import ".."/[cli, helpers, types_track_config]
 
-proc getSlugLookup(trackDir: Path): Table[string, string] =
+proc getConceptSlugLookup(trackDir: Path): Table[string, string] =
   let concepts = TrackConfig.init(readFile(trackDir / "config.json")).concepts
   result = initTable[string, string](concepts.len)
   for `concept` in concepts:
@@ -127,7 +127,7 @@ proc generate*(conf: Conf) =
 
   let conceptExercisesDir = trackDir / "exercises" / "concept"
   if dirExists(conceptExercisesDir):
-    let slugLookup = getSlugLookup(trackDir)
+    let slugLookup = getConceptSlugLookup(trackDir)
     for conceptExerciseDir in getSortedSubdirs(conceptExercisesDir):
       let introductionTemplatePath = conceptExerciseDir / ".docs" / "introduction.md.tpl"
       if fileExists(introductionTemplatePath):
--- a/src/generate/generate.nim
+++ b/src/generate/generate.nim
@@ -1,10 +1,6 @@
-import std/[hashes, parseutils, strbasics, strformat, strscans, strutils, sugar,
-            tables, terminal]
+import std/[parseutils, strbasics, strformat, strscans, strutils, sugar, tables,
+            terminal]
 import ".."/[cli, helpers, types_track_config]
-
-proc hash(slug: Slug): Hash {.borrow.}
-proc `==`(x, y: Slug): bool {.borrow.}
-proc add(s: var Slug, c: char) {.borrow.}
 
 proc getConceptSlugLookup(trackDir: Path): Table[Slug, string] =
   ## Returns a `Table` that maps each concept's `slug` to its `name`.
--- a/src/generate/generate.nim
+++ b/src/generate/generate.nim
@@ -2,6 +2,7 @@ import std/[parseutils, strbasics, strformat, strscans, strutils, tables, termin
 import ".."/[cli, helpers, types_track_config]
 
 proc getConceptSlugLookup(trackDir: Path): Table[string, string] =
+  ## Returns a `Table` that maps each concept's `slug` to its `name`.
   let concepts = TrackConfig.init(readFile(trackDir / "config.json")).concepts
   result = initTable[string, string](concepts.len)
   for `concept` in concepts:
--- a/src/generate/generate.nim
+++ b/src/generate/generate.nim
@@ -1,5 +1,11 @@
-import std/[parseutils, strbasics, strformat, strscans, terminal]
-import ".."/[cli, helpers]
+import std/[parseutils, strbasics, strformat, strscans, tables, terminal]
+import ".."/[cli, helpers, types_track_config]
+
+proc getSlugLookup(trackDir: Path): Table[string, string] =
+  let concepts = TrackConfig.init(readFile(trackDir / "config.json")).concepts
+  result = initTable[string, string](concepts.len)
+  for `concept` in concepts:
+    result[`concept`.slug] = `concept`.name
 
 proc writeError(description: string, path: Path) =
   let descriptionPrefix = description & ":"
@@ -10,7 +16,7 @@ proc writeError(description: string, path: Path) =
   stderr.writeLine(path)
   stderr.write "\n"
 
-func alterHeaders(s: string): string =
+func alterHeaders(s: string, title: string): string =
   # Markdown implementations differ on whether a space is required after the
   # final '#' character that begins the header.
   result = newStringOfCap(s.len)
@@ -19,6 +25,7 @@ func alterHeaders(s: string): string =
   # Skip the top-level header (if any)
   if i < s.len and s[i] == '#' and i+1 < s.len and s[i+1] == ' ':
     i += s.skipUntil('\n', i)
+  result.add &"## {title}"
   # Demote other headers
   var inFencedCodeBlock = false
   while i < s.len:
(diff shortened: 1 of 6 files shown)


ee7:
@ErikSchierboom Given [this `introduction.md.tpl`](https://raw.githubusercontent.com/exercism/common-lisp/6803f7519204/exercises/concept/character-study/.docs/introduction.md.tpl), and [this concept `introduction.md`](https://raw.githubusercontent.com/exercism/common-lisp/6803f7519204/concepts/characters/introduction.md), what is the expected generated `introduction.md` file?

Currently, this PR produces the following diff to the [exercise's `introduction.md`](https://raw.githubusercontent.com/exercism/common-lisp/6803f7519204/concepts/characters/introduction.md) in the common-lisp repo, meaning that it contains only level-1 and level-3 headers:
```diff
-## Comparing characters.
+### Comparing characters.
 
 Characters can be compared for equality by `char=` or `char-equal` (the latter being case-insensitive).
 
 Characters can be ordered with functions such as `char<`, `char>`, `char<=`, `char>=` and their case-insensitive versions `char-lessp` and `char-greaterp`.
 
-## Types of characters
+### Types of characters
 
 Characters can be categorized as different types: graphical, alphabetic, alpha-numeric, digit, upper-case or lower-case. 
 A character can be more than one type such as an upper-case alphabetic character.
 There are predicates for each of these types: `graphic-char-p`, `alpha-char-p`, `alphanumericp`, `digit-char-p`, `upper-case-p` and `lower-case-p`.
 
-## Converting character case
+### Converting character case
 
 Some characters can be upper or lower case and can be converted between them with `char-upcase` and `char-downcase`.
```

Is `configlet generate` supposed to add a header with the name of the concept? That is, to generate the second of these files:

- https://raw.githubusercontent.com/exercism/elixir/91ccf91940f3/concepts/binaries/introduction.md
- https://raw.githubusercontent.com/exercism/elixir/91ccf91940f3/exercises/concept/file-sniffer/.docs/introduction.md

does `introduction.md.tpl` contain

```markdown
# Introduction

## Binaries

%{concept:binaries}
```

or this?

```markdown
# Introduction

%{concept:binaries}
```

And if the latter, should we set the contents of the inserted level-2 header from the concept's `name` in the track-level `config.json` file?

ee7:
Pushed cb4788687d771e895f1c5a668a68eaf08e38b350 assuming that the answer is "yes" to both of:

> Is configlet generate supposed to add a header with the name of the concept? 

> should we set the contents of the inserted level-2 header from the concept's `name` in the track-level `config.json` file?

(I'll try to improve some variable names like `title` later).

ErikSchierboom:
> Is configlet generate supposed to add a header with the name of the concept?
> should we set the contents of the inserted level-2 header from the concept's name in the track-level config.json file?

Personally, I think the answer to both should be "yes", but I'll double-check with the other maintainers.

src/generate/generate.nim
```diff
@@ -10,21 +16,43 @@ proc writeError(description: string, path: Path) =
...
+  # Skip the top-level header (if any)
+  if i < s.len and s[i] == '#' and i+1 < s.len and s[i+1] == ' ':
+    i += s.skipUntil('\n', i)
+  result.add &"## {title}"
+  # Demote other headers
+  var inFencedCodeBlock = false
+  while i < s.len:
+    result.add s[i]
+    if s[i] == '\n':
+      # When inside a fenced code block, don't alter a line that begins with '#'
```
ErikSchierboom: The heading could also be in a comment block I think.

src/generate/generate.nim
```diff
@@ -1,5 +1,11 @@
-import std/[strbasics, strformat, strscans, terminal]
-import ".."/[cli, helpers]
+import std/[parseutils, strbasics, strformat, strscans, tables, terminal]
+import ".."/[cli, helpers, types_track_config]
+
+proc getSlugLookup(trackDir: Path): Table[string, string] =
```
ErikSchierboom: Maybe `getConceptSlugLookup`?

ErikSchierboom (approved):
LGTM Let's hold off on merging until other maintainers have chimed in.

angelikatyborska:
> does `introduction.md.tpl` contain
> 
> ```gfm
> # Introduction
> 
> ## Binaries
> 
> %{concept:binaries}
> ```
> 
> or this?
> 
> ```gfm
> # Introduction
> 
> %{concept:binaries}
> ```
> 
> And if the latter, should we set the contents of the inserted level-2 header from the concept's `name` in the track-level `config.json` file?

I always imagined that it would be the first option because it just seems more straight-forward (that merging markdown files doesn't require reading into the track's JSON config) and it's more flexible (you can add any header that you want). Anyway, either option is going to work for the Elixir track because we only ever nest single-concept introductions under a h2 that contains the concept's name. The second option might be nicer if you want to enforce this kind of pattern for all tracks.

verdammelt:
For the Common Lisp track it appears I'd favor the first - the track does not assume, nor necessarily want, a header with the name of the concept to be inserted - simply that the concept introductions "# Introduction" line be removed when inserting.

So I'd prefer what seems to be keeping the status quo. That being said I would be fine with the other style as well.

SleeplessByte:
I agree with the last two comments. I think for JS we kinda expect it too sometimes so things might break if option two is chosen.

ErikSchierboom:
Okay, then we should with option number 1. Could you update @ee7?

ee7:
> So I'd prefer what seems to be keeping the status quo. 

I don't know whether there's really a "status quo" currently, since the common-lisp track is currently the only track that has a `.md.tpl` file: 

```console
$ find . -name '*.md.tpl' | sort
./common-lisp/exercises/concept/character-study/.docs/introduction.md.tpl
./common-lisp/exercises/concept/gigasecond-anniversary/.docs/introduction.md.tpl
./common-lisp/exercises/concept/high-scores/.docs/introduction.md.tpl
./common-lisp/exercises/concept/key-comparison/.docs/introduction.md.tpl
./common-lisp/exercises/concept/larrys-winning-checker/.docs/introduction.md.tpl
./common-lisp/exercises/concept/leslies-lists/.docs/introduction.md.tpl
./common-lisp/exercises/concept/lillys-lasagna/.docs/introduction.md.tpl
./common-lisp/exercises/concept/lillys-lasagna-leftovers/.docs/introduction.md.tpl
./common-lisp/exercises/concept/logans-numeric-partition/.docs/introduction.md.tpl
./common-lisp/exercises/concept/log-levels/.docs/introduction.md.tpl
./common-lisp/exercises/concept/lucys-magnificent-mapper/.docs/introduction.md.tpl
./common-lisp/exercises/concept/pal-picker/.docs/introduction.md.tpl
./common-lisp/exercises/concept/pizza-pi/.docs/introduction.md.tpl
./common-lisp/exercises/concept/reporting-for-duty/.docs/introduction.md.tpl
./common-lisp/exercises/concept/socks-and-sexprs/.docs/introduction.md.tpl
```

Alternatively: https://github.com/search?q=org%3Aexercism+filename%3A*.md.tpl

It'd be slightly disruptive to the common-lisp track, which omits the h2 in the template file when only 1 concept is referenced. But is that a pattern we want?

<details>
  <summary>The contents of those files (click to expand)</summary>
  

#### `character-study/.docs/introduction.md.tpl`
```markdown
# Introduction

%{concept:characters}

```


#### `gigasecond-anniversary/.docs/introduction.md.tpl`
```markdown
# Introduction

## Dates and Times

%{concept:date-time}

## Multiple Values

%{concept:multiple-values}
```


#### `high-scores/.docs/introduction.md.tpl`
```markdown

# Introduction

%{concept:hash-tables}
```


#### `key-comparison/.docs/introduction.md.tpl`
```markdown

# Introduction

%{concept:equality}

```

#### `larrys-winning-checker/.docs/introduction.md.tpl`
```markdown
# Introduction

%{concept:arrays}
```


#### `leslies-lists/.docs/introduction.md.tpl`
```markdown
# Introduction

%{concept:lists}

```


#### `lillys-lasagna/.docs/introduction.md.tpl`
```markdown

# Introduction

%{concept:functions}

```

#### `lillys-lasagna-leftovers/.docs/introduction.md.tpl`
```markdown
# Introduction

%{concept:lambda-list}

## Optional Parameters

%{concept:optional-parameters}

## Keyword Parameters

%{concept:keyword-parameters}

## Rest Parameters

%{concept:rest-parameters}

```


#### `logans-numeric-partition/.docs/introduction.md.tpl`
```markdown
# Introduction

%{concept:reducing}
```


#### `log-levels/.docs/introduction.md.tpl`
```markdown
# Introduction

%{concept:strings}

```


#### `lucys-magnificent-mapper/.docs/introduction.md.tpl`
```markdown
# Introduction

## Mapping

%{concept:mapping}

## Filtering

%{concept:filtering}
```


#### `pal-picker/.docs/introduction.md.tpl`
```markdown
# Introduction

## Truthy And Falsy

%{concept:truthy-and-falsy}

## Conditionals

%{concept:conditionals}
```


#### `pizza-pi/.docs/introduction.md.tpl`
```markdown
# Introduction

In Common Lisp, like many languages, numbers come in a few of types – two of the most basic are:

## Integers

%{concept:integers}

## Floating Point Numbers

%{concept:floating-point-numbers}

## Arithmetic

%{concept:arithmetic}


```


#### `reporting-for-duty/.docs/introduction.md.tpl`
```markdown
# Introduction

%{concept:format-basics}

```


#### `socks-and-sexprs/.docs/introduction.md.tpl`
```markdown
# Introduction

## Comments

%{concept:comments}

## Cons

%{concept:cons}

## Expressions

%{concept:expressions}

## Symbols

%{concept:symbols}
```

</details>

ee7:
> the track does not assume, nor necessarily want, a header with the name of the concept to be inserted - simply that the concept introductions "# Introduction" line be removed when inserting. 

Then `configlet generate` must demote headers by some dynamic amount, right? Because if we don't hard-code inserting a h2, then `configlet generate` must not demote any headers for the below (to avoid going from h1 to h3):

```markdown
# Introduction

%{concept:foo}
```

and presumably should demote the headers in `foo` by two levels for:

```markdown
# Introduction

Blah blah

## h2

Blah blah.

### h3

%{concept:foo}
```

But I'm not sure whether we can support that use robustly with the current placeholder syntax. It's hard for `configlet generate` to guess the intended amount of demotion with something like:

```markdown
# Introduction

Blah blah.

## h2

Blah blah.

### h3

Blah blah.

%{concept:foo}
```

(is `%{concept:foo}` supposed to be under the h3 or not?).

If we want to do that, I think we need to agree on some concise rule for determing the demotion level, and maybe lint the `.tpl` files to avoid content being placed under headers of unexpected level. But would it be simpler and more robust to hard-code a demotion level of 1?

Note that https://exercism.org/docs/building/markdown/markdown contains:

> No heading may descend a level greater than one below the previous (e.g. `##` may only be followed by `###`, not `####`).

verdammelt:
@ee7 common-lisp is the only track using this? did not know.

So by "status quo" I guess I mean:  I would prefer it to continue to do whatever it may be doing now. But again, I can likely work with whatever y'all choose `generate` to do.

ErikSchierboom:
> It'd be slightly disruptive to the common-lisp track, which omits the h2 in the template file when only 1 concept is referenced. But is that a pattern we want?

 The C# and F# tracks also use this: https://github.com/exercism/csharp/blob/main/exercises/concept/log-levels/.docs/introduction.md I think it looks cleaner than an additional h2 heading, but 🤷 

> But would it be simpler and more robust to hard-code a demotion level of 1?

Would this depend on the concept placeholder always being place under an h2 heading? As in: what if the concept placeholder is directly under an h1 heading and the concept's introduction.md has h2 and h3 subheadings? 

> I think for JS we kinda expect it too sometimes so things might break if option two is chosen.

@SleeplessByte Do you perhaps have an example?

src/generate/generate.nim
ee7: Indeed. In https://github.com/exercism/configlet/pull/620/commits/528461562bc17c4579938cfe87e1cac42b173368 I mentioned that it doesn't support HTML blocks, and there are [a lot of them](https://spec.commonmark.org/0.30/#html-blocks) beside comment blocks.

However, in [Exercism's markdown spec](https://github.com/exercism/docs/blob/d334042e8f54ba7dd24ad215648355ff7795359a/building/markdown/markdown.md#comments) we write:

> Prefer Markdown comments instead of HTML comments (e.g. use `[comment]: # (Actual comment...)` rather than `<!-- Actual comment -->`

So do we want to support the below anyway?

```markdown
# Introduction

<!--
# Not a header
-->
```

And if we add `markdownlint` to every track, would we enable [`MD033`](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md033)? If "yes" to the latter, then we probably don't need to avoid demoting `#` inside HTML blocks in this PR.  

The markdown handling in this PR is very naive, but doing it robustly requires a full markdown parser. And sadly, it doesn't seem like there's a mature, robust, and pure-Nim markdown parser that allows:

```text
markdown -> AST -> manipulate to demote headers -> markdown
```

It looks like the best option is to use the [wrapper](https://github.com/zengxs/nim-cmark) for [libcmark](https://github.com/commonmark/cmark), but that will make it more complicated to keep configlet as a dependency-free binary.

ee7: To be more specific, it looks like [soasme/nim-markdown](https://github.com/soasme/nim-markdown) at best only supports

```
markdown -> AST -> manipulate AST -> HTML
```

The best I could do was along the lines of:

```nim
import std/[lists, strbasics, strutils, tables]
import pkg/markdown {.all.}

proc main =
  var s = """
    # Introduction

    Foo.

    ## Some header

    Bar.
  """.unindent()
  s.strip(chars={'\n'})

  # Avoid the overhead from calling the `markdown` proc
  # which uses a non-inline `strip`, and calls `render`.
  let root = Document(doc: s)
  var state = State(references: initTable[string, Reference](), config: initCommonmarkConfig())
  state.parse(root)
  for child in root.children:
    if child of Heading:
      echo child

when isMainModule:
  main()
```

```console
$ nim r foo.nim
<h1>Introduction</h1>
<h2>Some header</h2>
```

There are also performance killers like `token.children.toSeq.map((t: Token) => $t).join("\n")` everywhere in the critical path.

ErikSchierboom: I don't worry much about performance, as it is likely in the range of milliseconds? But the fact that you can't go back to markdown is quite annoying.

> So do we want to support the below anyway?

I think most editors would use that comment format by default. But maybe nobody actually uses headings in comments? 🤷

ee7: > I don't worry much about performance

I meant more like "this is code smell to me". It'd push me further towards "just use libcmark", even if `soasme/nim-markdown` supported markdown -> AST -> markdown.

> I think most editors would use that comment format by default.

Mmm. Probably true.

Supporting `<!-- -->` with the approach in this PR isn't much work. But do we want to support a line beginning with `#` inside any others from https://spec.commonmark.org/0.30/#html-blocks?

ee7:
> The C# and F# tracks also use this: https://github.com/exercism/csharp/blob/main/exercises/concept/log-levels/.docs/introduction.md I think it looks cleaner than an additional h2 heading, but 🤷

So the main difference is in the right panel of: 
- https://exercism.org/tracks/csharp/exercises/log-levels/edit
- https://exercism.org/tracks/elixir/exercises/pacman-rules/edit

Screenshots:



Would it help to have an Exercism-wide preference for one or the other? I think I prefer the Elixir way, with the h2. On the csharp link, there's no immediate signal that the entire introduction is about strings. The Elixir way is also more consistent between an exercise that uses a single placeholder, and an exercise that uses multiple placeholders.

> Would this depend on the concept placeholder always being place under an h2 heading? As in: what if the concept placeholder is directly under an h1 heading and the concept's introduction.md has h2 and h3 subheadings?

One possibility is "always demote by 1, but if the placeholder is not directly under a h2, add an h2 with the name of the concept". But there are many options. I don't know if we can decide what's best without enumerating them.

SleeplessByte:
@ee7 I like the Elixir way more.

ee7:
> common-lisp is the only track using this? did not know.

@verdammelt Yes, `common-lisp` is the early adopter :)

`configlet generate` hasn't really been fully usable yet, due to the deficiency that this PR tries to resolve.

For example, with these:
- [exercises/concept/gigasecond-anniversary/.docs/introduction.md.tpl](https://raw.githubusercontent.com/exercism/common-lisp/6803f7519204/exercises/concept/gigasecond-anniversary/.docs/introduction.md.tpl)
- [concepts/date-time/introduction.md](https://raw.githubusercontent.com/exercism/common-lisp/6803f7519204/concepts/date-time/introduction.md)
- [exercises/concept/gigasecond-anniversary/.docs/introduction.md](https://raw.githubusercontent.com/exercism/common-lisp/6803f7519204/exercises/concept/gigasecond-anniversary/.docs/introduction.md)

The last contains only level-2 headers, but should contain some level-3 headers.

`configlet generate` previously only produced acceptable output when there was only a top-level header in the `.md.tpl` file, like:
- [exercises/concept/character-study/.docs/introduction.md.tpl](https://raw.githubusercontent.com/exercism/common-lisp/6803f7519204/exercises/concept/character-study/.docs/introduction.md.tpl)
- [concepts/characters/introduction.md](https://raw.githubusercontent.com/exercism/common-lisp/6803f7519204/concepts/characters/introduction.md)
- [exercises/concept/character-study/.docs/introduction.md](https://raw.githubusercontent.com/exercism/common-lisp/6803f7519204/exercises/concept/character-study/.docs/introduction.md)

ee7:
Running `configlet generate` with the latest commit in this PR (d7cea656544219201ba24d2f417e1bea69e82d01) produces this diff to the common-lisp track: https://github.com/exercism/common-lisp/pull/659

The behavior is as follows. Thoughts?

#### Placeholder under `#`

Insert a h2 using the concept's name in the track `config.json`, and demote the headers in the `introduction.md` for the `foo` concept by 1:

```markdown
# Introduction

%{concept:foo}
```

#### Placeholder under `##`

Don't insert an extra header, and demote concept's headers by 1:

```markdown
# Introduction

## Foo

%{concept:foo}
```

#### Placeholder under `###`

Don't insert an extra header, and demote concept's headers by 2:

```markdown
# Introduction

## Some header

Blah blah.

### Foo

%{concept:foo}
```

---

One hurdle from https://exercism.org/docs/building/markdown/markdown:

> Beyond the single level-1 (`#`) heading, only level-2 (`##`), level-3 (`###`) and level-4 (`####`) headings may be used.

So even demoting by 1 level is problematic if the concept introduction already contained a level-4 heading. How should we handle that?

src/generate/generate.nim
ee7: Pushed https://github.com/exercism/configlet/pull/620/commits/da1d6951e54cf2c4a8b1400ec68f5ff4e6bc5ad2 to add support for HTML comment blocks

ErikSchierboom:
> Would it help to have an Exercism-wide preference for one or the other? I think I prefer the Elixir way, with the h2. On the csharp link, there's no immediate signal that the entire introduction is about strings. The Elixir way is also more consistent between an exercise that uses a single placeholder, and an exercise that uses multiple placeholders.

Seeing them displayed side-by-side, I now also like the Elixir way better.

> So even demoting by 1 level is problematic if the concept introduction already contained a level-4 heading. How should we handle that?

I don't think we need to support this. The concepts are such a key feature that they should really be their own level-2 headings, not merged into 

> The behavior is as follows. Thoughts?

I think it's great!

ee7:
Latest hurdle: reference links.

Consider this `.tpl.md`:

```markdown
# Introduction

## Docs

%{concept:docs}

## Typespecs

%{concept:typespecs}
```

which inserts:
- [`concepts/docs/introduction.md`](https://raw.githubusercontent.com/exercism/elixir/906544c076d3/concepts/docs/introduction.md)
- [`concepts/typespecs/introduction.md`](https://raw.githubusercontent.com/exercism/elixir/906544c076d3/concepts/typespecs/introduction.md)

Note that the first has this reference link at the end:

```markdown
[markdown]: https://docs.github.com/en/github/writing-on-github/basic-writing-and-formatting-syntax
```

and the second has this:

```markdown
[types]: https://hexdocs.pm/elixir/typespecs.html#types-and-their-syntax
```

So the current implementation of `configlet generate` produces

```markdown
# Introduction

## Docs

Blah blah.

Elixir documentation is written in [**Markdown**][markdown].

Blah blah.

[markdown]: https://docs.github.com/en/github/writing-on-github/basic-writing-and-formatting-syntax

## Typespecs

Blah blah.

### Types

Blah blah.

A full list of all types can be found in the ["Typespecs" section in the official documentation][types].

### Naming arguments

Blah blah.

### Custom types

Blah blah.

[types]: https://hexdocs.pm/elixir/typespecs.html#types-and-their-syntax
```

But I suppose we want all reference links together at the bottom, in order of usage? And de-duplicated if multiple files have the same reference link?

```markdown
# Introduction

## Docs

Blah blah.

Elixir documentation is written in [**Markdown**][markdown].

Blah blah.

## Typespecs

Blah blah.

### Types

Blah blah.

A full list of all types can be found in the ["Typespecs" section in the official documentation][types].

### Naming arguments

Blah blah.

### Custom types

Blah blah.

[markdown]: https://docs.github.com/en/github/writing-on-github/basic-writing-and-formatting-syntax
[types]: https://hexdocs.pm/elixir/typespecs.html#types-and-their-syntax
```

From https://exercism.org/docs/building/markdown/markdown#h-links

> Please use [reference links](https://spec.commonmark.org/0.29/#reference-link), which are defined **at the bottom of the Markdown file**, mapped to a reference slug, and referenced by that slug in the text.

ErikSchierboom:
> Latest hurdle: reference links.

There are more hurdles than I expected! Ideally, yes: putting all the reference links at the bottom would be best. It doesn't matter much for people using the online editor, but CLI users might find it odd. If this is hard to do, I could also do this in the website level, as I could automatically convert reference links to  inline links when serving the markdown files to the CLI.

ee7:
>  Ideally, yes: putting all the reference links at the bottom would be best.

OK, I'll aim to do that.

I've also created https://github.com/exercism/elixir/pull/1151 while testing.

ee7:
I've investigated the `libcmark` option.

Pros:
- We'd actually parse the markdown properly.

Cons:
- It parses into an AST that doesn't contain nodes for reference link definitions. So `markdown -> libcmark -> render as commonmark` converts all reference links into inline links.
- The AST doesn't store some other information, e.g. the delimiter kind used for emphasis or code blocks (so `_` is later rendered as `*`, and an admonition like `~~~~exercism/note` is later rendered as `` ``` exercism/note ``). 
- You can't customize the markdown output format. But the standardized output form means that simple post-processing can be quite robust.
- It would increase the configlet binary size by about 50% (roughly 487 KiB to 727 KiB on Linux).
- I don't know yet whether the Windows build is trivial.

So I've got a working `configlet generate` that uses libcmark and demotes headers, but it writes markdown like this https://github.com/ee7/exercism-elixir/commit/d05c1e7f2668ee48ec5d4d23d01b310f801e84a8

How many of the cons are dealbreakers?

And would it actually be nice for configlet to be able to format markdown? For example, `configlet fmt --markdown` would become possible, which would also resolve many `markdownlint` rule violations. But here the formatting would be a little different from that of `prettier`, unless we have some extra processing step. Using another parsing/formatting library might also be possible.

ee7:
Aside: this PR has reminded of the complexity of markdown parsing.

The [CommonMark spec](https://spec.commonmark.org/0.30/) is very complex. Scroll through examples 253–300 or 350–480 (yes, 130 examples to show how emphasis works) for a taste.

These are also nice reads:
- https://talk.commonmark.org/t/beyond-markdown/2787
- https://talk.commonmark.org/t/issues-we-must-resolve-before-1-0-release-6-remaining/1287

#### Markdown quiz 1: what does the below produce?

```markdown
- [foo]: bar
baz
```

<details>

<summary>Expand for answer</summary>

Probably a list with one item, baz:

- [foo]: bar
baz

The above has the same markdown source. That is, GitHub renders:

```html
<ul>
  <li>
    baz
  </li>
</ul>
```

But [other renderings](https://babelmark.github.io/?text=-+%5Bfoo%5D%3A+bar%0Abaz) include these:

```html
<ul>
  <li>
    [foo]: bar baz
  </li>
</ul>
```

```html
<ul>
  <li>
    <p>
      baz
    </p>
  </li>
</ul>
```

```html
<ul>
  <li>
  </li>
</ul>
<p>
  baz
</p>
```

```html
<ul>
  <li>
  </li>
</ul>
```

```html
<ul>
  <li>
    <p>
      : bar
    </p>
  </li>
</ul>
<p>
  baz
</p>
```

</details>


#### Markdown quiz 2: what is the correct HTML rendering for the below Markdown?

```text
*****Hello*world****
```

<details>

<summary>Expand for answer</summary>

```
<p>**<em><strong>Hello<em>world</em></strong></em></p>
```

That is:

<p>**<em><strong>Hello<em>world</em></strong></em></p>

But [a year ago](https://github.com/commonmark/cmark/commit/34250e12ccebdc6372b8b49c44fab57c72443460), the reference implementation of the CommonMark spec rendered it incorrectly as: 

```
<p>*****Hello<em>world</em>***</p>
```

That is:

<p>*****Hello<em>world</em>***</p>

which, at the time of writing, is also still how GitHub renders it.

(from https://talk.commonmark.org/t/i-dont-understand-how-emphasis-is-parsed/3866)

And at the time of writing, a parser comparison tool [produces](https://babelmark.github.io/?text=*****Hello*world****) 11 (!) different renderings.

</details>

ErikSchierboom:
> It parses into an AST that doesn't contain nodes for reference link definitions. So markdown -> libcmark -> render as commonmark converts all reference links into inline links.

If I'm reading this correctly, we wouldn't be able to retain reference links if we were to attempt to format Markdown? 

> The AST doesn't store some other information, e.g. the delimiter kind used for emphasis or code blocks (so _ is later rendered as *, and an admonition like ~~~~exercism/note is later rendered as ``` exercism/note).

Not the best of ASTs then. 

> You can't customize the markdown output format.

:( Ruby's cmark wrapper allows this, so they must have custom-built it?

> It would increase the configlet binary size by about 50% (roughly 487 KiB to 727 KiB on Linux).

I don't mind this much to be honest.

> I don't know yet whether the Windows build is trivial.

Having a proper Windows build is important.

ee7:
> we wouldn't be able to retain reference links if we were to attempt to format Markdown?

Not without an additional step to add them back in again. See in https://github.com/ee7/exercism-elixir/commit/d05c1e7f2668ee48ec5d4d23d01b310f801e84a8, in particular the transformations of reference links. That commit is the diff produced by a `configlet generate` that uses `libcmark` vs the output of `configlet generate` from this PR.

> Not the best of ASTs then.

Mmm. I imagine that it's deliberate, though - John Macfarlane knows what he's doing.

> :( Ruby's cmark wrapper allows this, so they must have custom-built it?

Do you mean [`commonmarker`](https://github.com/gjtorikian/commonmarker), the wrapper for [`libcmark-gfm`](https://github.com/github/cmark-gfm)?

My statement of "can't customize the markdown output format" was a bit too strong - I meant that the only options to customize rendering are these:

```console
$ man 3 cmark
[...]
Options affecting rendering
     #define CMARK_OPT_SOURCEPOS (1 << 1)

     Include a data-sourcepos attribute on all block elements.

     #define CMARK_OPT_HARDBREAKS (1 << 2)

     Render softbreak elements as hard line breaks.

     #define CMARK_OPT_SAFE (1 << 3)

     CMARK_OPT_SAFE is defined here for API compatibility, but it no longer has
     any effect. "Safe" mode is now the default: set CMARK_OPT_UNSAFE to disable
     it.

     #define CMARK_OPT_UNSAFE (1 << 17)

     Render raw HTML and unsafe links (javascript:, vbscript:, file:, and data:,
     except for image/png, image/gif, image/jpeg, or image/webp mime types).
     By default, raw HTML is replaced by a placeholder HTML comment.
     Unsafe links are replaced by empty strings.

     #define CMARK_OPT_NOBREAKS (1 << 4)

     Render softbreak elements as spaces.
[...]
```

So I believe that you can't make it e.g. prefer `_` over `*` for emphasis, or omit the space between a code fence and language name, or omit indentation before list items.

It looks like the [render options for commonmarker](https://github.com/gjtorikian/commonmarker/blob/558c7275b18a7ae16136c0fc55f444458dd8cc58/README.md#render-options) are the same as the [underlying ones for libcmark-gfm](https://github.com/github/cmark-gfm/blob/ef1cfcb8a31de72bb951d9259bc4ac89e266975b/man/man3/cmark-gfm.3#L829-L1010). But yeah, it looks like [commonmarker exposes a way to use a custom renderer](https://github.com/gjtorikian/commonmarker/blob/558c7275b18a7ae16136c0fc55f444458dd8cc58/README.md#creating-a-custom-renderer). Is that what you're talking about?

Edit:
Looks like "yes": [exercism/website/app/commands/markdown/render_html.rb](https://github.com/exercism/website/blob/c1dce78e906d57a45207a33df7b6fece000d6a03/app/commands/markdown/render_html.rb#L18)

ErikSchierboom:
I was indeed referring to that. But what do you think we should do?

ee7:
> But what do you think we should do?

Well, this PR at least works for https://github.com/exercism/elixir/pull/1151, including the combining of reference links at the bottom when multiple concepts have at least one reference link.

It's just that the current parsing is somewhat fragile, especially to whitespace and linebreaks.

I'm looking into patching `libcmark` at install time.

ee7:
I'll do the `cmark` change in a follow-up PR. This PR is already long, and a separate commit will help if we needed to revert to this non-`cmark` implementation for some reason.

Erik: the total diff since your last approval is https://github.com/exercism/configlet/compare/cb4788687d771e895f1c5a668a68eaf08e38b350...607f369b84761d308af66c6b74d87ee618f61cbf

src/generate/generate.nim
ee7: Done: https://github.com/exercism/configlet/pull/620/commits/d1a1ab1b7859ed4b726f3c6c052dd2e10f93eb84

src/generate/generate.nim
```diff
@@ -1,5 +1,54 @@
...
+import std/[parseutils, strbasics, strformat, strscans, strutils, sugar, tables,
+            terminal]
+import ".."/[cli, helpers, types_track_config]
+
+proc getConceptSlugLookup(trackDir: Path): Table[Slug, string] =
+  ## Returns a `Table` that maps each concept's `slug` to its `name`.
+  let concepts = TrackConfig.init(readFile(trackDir / "config.json")).concepts
+  collect:
+    for con in concepts:
+      {con.slug.Slug: con.name}
```
ee7, line 10: The `.Slug` is needed here because the concept's `slug` is not yet a `Slug`.

https://github.com/exercism/configlet/blob/d39169d76a2cbb01a6bdc30bd7684dca7c906d81/src/types_track_config.nim#L43-L46

We should handle changing to `slug: Slug` in a separate PR, because it's not just a refactoring (depending on which parsing hooks are visible to jsony at parse time, it would add a parse-time error if a slug value in the JSON file is not `kebab-case`).

But note that, for now, `configlet generate` does not validate that a `slug` value is kebab-case. I think that it should, but it doesn't matter much for now: the placeholder matching does enforce kebab-case for the concept value.

ee7:
@ErikSchierboom Building configlet from the latest commit in this PR, running `configlet generate` on the common-lisp track produces the penultimate state in https://github.com/exercism/common-lisp/pull/659. Could you check that PR to see if you're happy with how configlet behaves? Please review individual commits in that PR, to see configlet's behavior when the placeholder is both under and not under a level-2 heading.

`common-lisp` is still the only track that uses a `.tpl` file.

ErikSchierboom:
That looks great to me. Thanks a lot!

PR #623, opened 2022-06-26
sync: remove "Updating tests..." message

### LINKED ISSUE #622: sync: reports "Updating tests..." when it's not
Noticed that when using `sync` to update *documents*, the output indicates that configlet is updating tests:

```console
~/git/c[sync]❯ ./bin/configlet --version
4.0.0-beta.4

~/git/c[sync]❯ ./bin/configlet sync -o -e crypto-square -u
Checking exercises...
[warn] docs: instructions unsynced: crypto-square
sync the above docs ([y]es/[n]o)? y
Updated the docs for 1 Practice Exercise
Updating tests...                                                                    # No its not!
The `crypto-square` exercise has up-to-date docs, filepaths, metadata, and tests!
```

Perhaps this refers to some tests other than those in the exercise?
In which case probably a re-wording of the statement could be useful.

Thanks!

ee7:
It was confusing that configlet always printed "Updating tests" when
tests were part of the syncing scope, even if every test was up to date.

Before this commit:

```console
$ configlet sync -o -e hello-world -u
Checking exercises...
Updating tests...
The `hello-world` exercise has up-to-date docs, filepaths, metadata, and tests!
```

With this commit:

```console
$ configlet sync -o -e hello-world -u
Checking exercises...
The `hello-world` exercise has up-to-date docs, filepaths, metadata, and tests!
```

The "Updating tests..." message was trying to say "OK, we'll start
checking tests now, and if we find anything, we'll prompt to update".
But because all the tests encountered are up to date, there is no
further output to ask "Do you want to include this test case?".

Furthermore, configlet doesn't print things like

```text
Checking docs...
Checking filepaths...
Checking metadata...
```

as it enters the other syncing scopes, and the "Updating tests..."
message is a leftover from the `canoncial_data_syncer` days when the
application only did syncing, was destructive by default (we had
--check, not --update), and only operated on tests. So let's remove the
message. But later we could consider adding a "Checking foo..." message
for each syncing scope at the highest verbosity level.

Closes: #622

### DIFF
--- a/src/sync/sync_tests.nim
+++ b/src/sync/sync_tests.nim
@@ -135,8 +135,6 @@ proc syncIfNeeded(exercise: Exercise, conf: Conf): bool =
 
 proc updateTests*(seenUnsynced: var set[SyncKind], conf: Conf,
                   exercises: seq[Exercise]) =
-  logNormal("Updating tests...")
-
   var everyExerciseIsSynced = true
 
   for exercise in exercises:
--- a/tests/test_binary.nim
+++ b/tests/test_binary.nim
@@ -57,7 +57,6 @@ proc testsForSync(binaryPath: static string) =
     syncOfflineUpdateTests = &"{syncOfflineUpdate} --tests"
 
     header = "Checking exercises..."
-    headerUpdateTests = "Updating tests..."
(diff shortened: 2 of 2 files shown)


ErikSchierboom (approved):
Nice

PR #629, opened 2022-07-08
completions: add completions for bash and fish

### LINKED ISSUE #615: shell tab completions
I've been building a track, so I've been using configlet a lot. With the plethora of subcommands and options, I figured tab completion would be handy. My shell is fish, so that's where I started.

But then I thought "how to install the completions"? How about this: add an option to `fetch-configlet` to download a shell-specific completions file, and emit instructions on how to source it into the user's shell.

Example:
```console
$ bin/fetch-configlet -c fish
To enable completions: source bin/configlet-completions.fish
$ bin/fetch-configlet -c bash
To enable completions: . bin/configlet-completions.bash
$ ls -1 bin
configlet
configlet-completions.bash
configlet-completions.fish
fetch-configlet
...
```

Do you think this would be useful?

glennj:
Not being a zsh or PS user, those completions are not included.

Changes to fetch-configlet to detect user's shell and print instructions will follow in a separate PR.

Closes #615

### DIFF
--- a/completions/configlet.bash
+++ b/completions/configlet.bash
@@ -1,0 +1,104 @@
+# completions for the configlet command, bash flavour
+
+# remove any prior completions
+complete -r bin/configlet configlet 2>/dev/null
+# and install this one
+complete -F _configlet_completion_ configlet
+
+# @(pattern1|pattern2|...) is bash extended pattern matching meaning
+# "one of pattern1 or pattern2 or ..."
+# ref: https://www.gnu.org/software/bash/manual/bash.html#Pattern-Matching
+
+_configlet_completion_() {
+    local global_opts='-h --help --version -t --track-dir -v --verbosity'
+    local cur=${COMP_WORDS[COMP_CWORD]}
+    local prev=${COMP_WORDS[COMP_CWORD - 1]}
+
+    # Check for global options that takes a value
+    if _configlet_complete_global_option_; then
+        return
+    fi
+
+    local i
+    for ((i = 1; i < COMP_CWORD; i++)); do
+        if [[ ${COMP_WORDS[i]} == @(lint|generate|info|uuid|fmt|sync) ]]; then
+            "_configlet_complete_${COMP_WORDS[i]}_"
+            return
+        fi
+    done
+
+    _configlet_complete_options_ "fmt generate info lint sync uuid $global_opts"
+}
+
+_configlet_complete_global_option_() {
+    case $prev in
+        '-v' | '--verbosity')
+            _configlet_complete_options_ "quiet normal detailed"
+            return 0
+            ;;
+        '-t' | '--track-dir')
+            # Complete a directory based on what the user's typed so far
+            mapfile -t COMPREPLY < <(compgen -A directory -- "$cur")
+            return 0
+            ;;
+    esac
+    return 1
+}
+
+_configlet_complete_lint_() {
+    _configlet_complete_options_ "$global_opts"
+}
+
+_configlet_complete_generate_() {
+    _configlet_complete_options_ "$global_opts"
+}
+
+_configlet_complete_info_() {
+    _configlet_complete_options_ "-o --offline $global_opts"
+}
+
+_configlet_complete_uuid_() {
+    _configlet_complete_options_ "-n --num $global_opts"
+}
+
+_configlet_complete_fmt_() {
+    case $prev in
+        '-e' | '--exercise')
+            _configlet_complete_slugs_ "practice" "concept"
+            ;;
+        *)
+            _configlet_complete_options_ "-e --exercise -u --update -y --yes $global_opts"
+            ;;
+    esac
+}
+
+_configlet_complete_sync_() {
+    case $prev in
+        '--tests')
+            _configlet_complete_options_ "choose include exclude"
+            ;;
+        '-e' | '--exercise')
+            _configlet_complete_slugs_ "practice"
+            ;;
+        *)
+            _configlet_complete_options_ "-e --exercise -o --offline -u --update -y --yes --docs --metadata --tests $global_opts"
+            ;;
+    esac
+}
+
+# Note that configlet expects to be called from the track's root dir.
+_configlet_complete_slugs_() {
+    local subdir
+    mapfile -t COMPREPLY < <(
+        for subdir in "$@"; do
+            if [[ -d "./exercises/$subdir" ]]; then
+                ( cd "./exercises/$subdir" && compgen -A directory -- "$cur" )
+            fi
+        done
+    )
+}
+
+_configlet_complete_options_() {
+    local choices=$1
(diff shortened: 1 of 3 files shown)


ErikSchierboom:
Thanks! I'll let ee7 review this, as he is much more well-versed into this than I am.

scripts/fetch-configlet
```diff
@@ -49,6 +49,30 @@ get_download_url() {
     cut -d'"' -f4
 }
 
+# get the user's shell (as portably as possible)
+# and print out the command to enable tab completion
+emit_completion_command() {
+    local shell file
+    if command -v getent >/dev/null; then
+        shell=$(getent passwd "${USER}" | cut -d: -f6)
```
ee7:
```suggestion
        shell=$(getent passwd "${USER}" | cut -d: -f7)
```

The format for `/etc/passwd` is the below, right?
```
account:password:UID:GID:GECOS:directory:shell
```

So we need the seventh field to get the shell?

For example, on Arch Linux:

```console
$ getent passwd "${USER}"
myusername:x:1000:1000::/home/myusername:/bin/zsh
$ getent passwd "${USER}" | cut -d: -f6
/home/myusername
$ getent passwd "${USER}" | cut -d: -f7
/bin/zsh 
```

On OpenBSD:

```console
$ getent passwd "${USER}"
myusername:*:1000:1000:myusername:/home/myusername:/bin/ksh
```

scripts/fetch-configlet
```diff
@@ -49,6 +49,30 @@ get_download_url() {
...
+    elif command -v finger >/dev/null; then
+        shell=$(finger "${USER}" | sed -nE '/.*Shell: / { s///; s/ .*//; p; }')
+    elif [[ -f /etc/passwd ]]; then
+        shell=$(awk -F: -v u="${USER}" '$1 == u {print $6}' /etc/passwd)
```
ee7:
```suggestion
    elif [[ -f /etc/passwd ]]; then
        shell=$(awk -F: -v u="${USER}" '$1 == u {print $7}' /etc/passwd)
```

Ditto.

completions/configlet.bash
```diff
@@ -0,0 +1,104 @@
...
+    return 1
+}
+
+_configlet_complete_lint_() {
+    _configlet_complete_options_ "$global_opts"
+}
+
+_configlet_complete_generate_() {
+    _configlet_complete_options_ "$global_opts"
+}
```
ee7: For later: `configlet generate` will (soon-ish) have the same non-destructive-by-default behavior as `configlet fmt`, and so support the same options (`-u`, `-y`, `-e`) . So we should remember to update the completions accordingly.

completions/configlet.bash
```diff
@@ -0,0 +1,104 @@
...
+_configlet_complete_sync_() {
+    case $prev in
+        '--tests')
+            _configlet_complete_options_ "choose include exclude"
+            ;;
+        '-e' | '--exercise')
+            _configlet_complete_slugs_ "practice"
+            ;;
+        *)
+            _configlet_complete_options_ "-e --exercise -o --offline -u --update -y --yes --docs --metadata --tests $global_opts"
```
ee7:
```suggestion
            _configlet_complete_options_ "-e --exercise -o --offline -u --update -y --yes --docs --filepaths --metadata --tests $global_opts"
```

Was missing `--filepaths`?

completions/configlet.fish
```diff
@@ -0,0 +1,35 @@
...
+complete -c configlet -n "__fish_seen_subcommand_from fmt" -s e -l exercise -d "exercise slug" \
+    -x -a "(find ./exercises/{concept,practice} -maxdepth 1 -mindepth 1 -type d -printf '%P\n' | sort)"
+
+# sync subcommand
+complete -c configlet -n "__fish_use_subcommand" -a sync -d "Check or update Practice Exercise docs, metadata, and tests"
+complete -c configlet -n "__fish_seen_subcommand_from sync" -s o -l offline -f -d "Do not update proc-spec cache"
+complete -c configlet -n "__fish_seen_subcommand_from sync" -s u -l update -d "Write changes"
+complete -c configlet -n "__fish_seen_subcommand_from sync" -s y -l yes -d "Auto-confirm update"
+complete -c configlet -n "__fish_seen_subcommand_from sync"      -l docs -d "Sync docs only"
+complete -c configlet -n "__fish_seen_subcommand_from sync"      -l metadata -d "Sync metadata only"
```
ee7:
```suggestion
complete -c configlet -n "__fish_seen_subcommand_from sync"      -l docs -d "Sync docs only"
complete -c configlet -n "__fish_seen_subcommand_from sync"      -l filepaths -d "Sync filepaths only"
complete -c configlet -n "__fish_seen_subcommand_from sync"      -l metadata -d "Sync metadata only"
```

Ditto.

ee7 (commented):
Thanks for working on this. Very cool.

First pass at a review. Great that it's even got slug completion, but does it find the slugs at "tab time" or at source time? The latter would be a problem when working on multiple tracks. (I haven't tested yet, sorry).

~I see that [`db12f3b..10dd1e0`](https://github.com/exercism/configlet/compare/db12f3b5b38287e2c79f38c0bec261ded0d4358c..10dd1e032b22c0196dfcd609ca19c590ef8b7104) removed the shell detection - are you leaning towards not trying to detect it?~

Edit: Oh, from https://github.com/exercism/configlet/issues/615#issuecomment-1181410546 I guess you'll make a separate PR for the `fetch-configlet` changes.

glennj:
> does it find the slugs at "tab time" or at source time? 

At tab time.

completions/configlet.bash
glennj: Woopsie

completions/configlet.bash
glennj: When that happens, we can just dispatch:
```bash
 _configlet_complete_generate_() {
    _configlet_complete_fmt_
}
```

scripts/fetch-configlet
glennj: Good catch. Will fix in next PR.

ee7 (approved):
Looks great. Thanks again, and sorry again for the delay in reviewing.

I hadn't used fish for a long time. Looks like they've improved compatibility a lot lately (e.g. support for `&&` and `||`, `&>`, and `foo=bar echo $foo`). And the completion approach for fish seems nice. Clear and self-documenting.

I took the liberty of pushing a few trivial commits (please yell at me if you object), and I've opened a bunch of follow-up issues. Minor things like https://github.com/exercism/configlet/issues/637 (I don't know if it's easy to address) and #642.

Merging.

glennj:
Thanks @ee7!

ee7:
Released at last: https://github.com/exercism/configlet/releases/tag/4.0.0-beta.6

PR #643, opened 2022-08-04
completions: make `uuid -n` expect an argument

### LINKED ISSUE #636: completions: make `uuid -n` expect an argument

ee7:
Follow-up PR from #629.

@glennj does this look OK to you?

---

Before this commit, tab completion after `-n` would suggest global
options, even though `-n` without an argument is an error. For the fish
completions:

```console
$ configlet uuid -n[Tab]
-nh  (Show help)  -nt  (Select a track directory)  -nv  (Verbose level)
```

This commit removes that completion for both the bash and fish
completion scripts.

The fish [docs for `complete`](https://fishshell.com/docs/current/cmds/complete.html) describe the relevant options as:

> `-f` or `--no-files`
> This completion may not be followed by a filename.
> 
> `-r` or `--require-parameter`
> This completion must have an option argument, i.e. may not be followed by another option.
> 
> `-x` or `--exclusive`
> Short for `-r` and `-f`.

Closes: #636

### DIFF
--- a/completions/configlet.bash
+++ b/completions/configlet.bash
@@ -58,7 +58,14 @@ _configlet_complete_info_() {
 }
 
 _configlet_complete_uuid_() {
-  _configlet_complete_options_ "-n --num $global_opts"
+  case $prev in
+    '-n' | '--num')
+      return 0 # Don't suggest a global option if -n has no argument.
+      ;;
+    *)
+      _configlet_complete_options_ "-n --num $global_opts"
+      ;;
+  esac
 }
 
 _configlet_complete_fmt_() {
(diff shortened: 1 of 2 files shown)


ErikSchierboom (approved):
I'm approving as the code owner. I'll let you all decide whether or not the contents are good :)

ee7:
I've tested that this does at least stop suggesting global options with `-n [Tab]` or `--num [Tab]`. So I'll merge. Yell at me if it's broken somehow.

Ideally, we could print something like "specify the number of uuids", instead of making tab completion do nothing. But let's leave that for later.

glennj:
Or we could seed the completions with something like "1", "2", "3", "..." to indicate that an integer is expected.

ee7:
I agree that'd be better. I tried it the time with e.g. 1, 5, and 10. But I gave up when the completions appeared in the order `1 10 5`.

PR #647, opened 2022-08-06
fetch-configlet: print instructions for installing completions

### LINKED ISSUE #635: fetch-configlet: detect user's shell and print instructions
Follow-up from https://github.com/exercism/configlet/pull/629.

For fish, I think the instructions should advise the user to run

```shell
configlet completion -s fish > ~/.config/fish/completions/configlet.fish
```

ee7:
Closes: #635

---

@glennj I've taken the liberty of taking your commit from https://github.com/glennj/exercism-configlet/commit/e55d7757efca6371be0fcd0541d3416d4e7f9104 to begin this PR, which should be attributed to you.

The first commit in this PR is just that commit, but rebased on `main`. Feel free to push to this PR's branch.

This PR is a follow-up to https://github.com/exercism/configlet/pull/629, where these changes were originally.

### DIFF
--- a/scripts/fetch-configlet
+++ b/scripts/fetch-configlet
@@ -70,7 +70,8 @@ emit_completion_command() {
   file="bin/completions/configlet.$(basename "${shell}")"
   if [[ -f "${file}" && -r "${file}" ]]; then
     case "${shell}" in
-      bash | fish) echo "Enable tab completions: source ${file}" ;;
+      bash | fish)
+        echo "Enable tab completions: source ${file}" ;;
       zsh)
         : # is there more to do here than the above?
         ;;
--- a/scripts/fetch-configlet
+++ b/scripts/fetch-configlet
@@ -70,9 +70,15 @@ emit_completion_command() {
   file="bin/completions/configlet.$(basename "${shell}")"
   if [[ -f "${file}" && -r "${file}" ]]; then
     case "${shell}" in
-      bash | fish)
+      bash)
         echo "To enable tab completions, run:"
         echo "    configlet completion -s ${shell} | source"
+        ;;
+      fish)
+        echo "To enable tab completions for the current session, run:"
+        echo "    configlet completion -s ${shell} | source"
+        echo "To enable tab completions for every session, run:"
+        echo "    configlet completion -s ${shell} > ~/.config/fish/completions/configlet.fish"
         ;;
       zsh)
         : # is there more to do here than the above?
--- a/scripts/fetch-configlet
+++ b/scripts/fetch-configlet
@@ -54,9 +54,9 @@ get_download_url() {
 emit_completion_command() {
   local shell file
 
-  if command -v getent >/dev/null; then
+  if command -v getent > /dev/null; then
     shell=$(getent passwd "${USER}" | cut -d: -f7)
-  elif command -v finger >/dev/null; then
+  elif command -v finger > /dev/null; then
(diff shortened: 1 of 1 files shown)


ee7:
It's cool that we try to detect a user's shell. But I have these questions:

1. Is there a way to reliably detect the user's current shell when it is different from their default shell?
2. And if not, what do we think about "don't try to detect the user's shell, and just print the instructions for every supported shell"?

For example, a user might use fish as an interactive shell only, as [discussed on the Arch wiki](https://wiki.archlinux.org/title/fish#System_integration). Right now, for such a user,

```
getent passwd "${USER}" | cut -d: -f7
```

may return

```text
/bin/bash
```

in which case the user sees instructions for a different shell. Or it may return, say:

```text
/bin/zsh
```

in which case the user wouldn't see any instructions for installing completions, even though they're available.

Related [recent post of Glenn's](https://unix.stackexchange.com/questions/352316/finding-out-the-default-shell-of-a-user-within-a-shell-script#comment1338508_352319) on unix.stackexchange.com:

> `$SHELL` _is_ the user's login shell. My login shell is fish: if I unset the SHELL variable and then launch bash, SHELL is `/path/to/fish`

ErikSchierboom:
> And if not, what do we think about "don't try to detect the user's shell, and just print the instructions for every supported shell"?

If there isn't a reliable, fool-proof way to detect the user's shell, I think we should print the instructions for every supported shell.

glennj:
> 1. Is there a way to reliably detect the user's current shell when it is different from their default shell?

No, unless we parse `ps` to find the shell in the process hierarchy.

> 2. And if not, what do we think about "don't try to detect the user's shell, and just print the instructions for every supported shell"?

I think this was my original suggestion: let the user tell you what shell they want:
```sh
configlet completions -s fish
```
Perhaps if the `-s` option is not provided, list the available shells for which we have completions.

ee7:
> If there isn't a reliable, fool-proof way to detect the user's shell, I think we should print the instructions for every supported shell.

Let's do that.

> No, unless we parse `ps` to find the shell in the process hierarchy.

Hah. I think that'd be trying to be much too clever.

> I think this was my original suggestion: let the user tell you what shell they want:
> 
> ```shell
> configlet completions -s fish
> ```

We'll have this in the next configlet release (but with the subcommand called `completion`, not `completions` - see https://github.com/exercism/configlet/commit/d10dcd6ddf19b9f5441c0c944bc6f3486a2052f5).

The help message will contain

```text
Usage:
  configlet [global-options] <command> [command-options]

Commands:
  completion  Output a completion script for a given shell
  fmt         Format the exercise '.meta/config.json' files
  generate    Generate Concept Exercise 'introduction.md' files from 'introduction.md.tpl' files
  info        Print some information about the track
  lint        Check the track configuration for correctness
  sync        Check or update Practice Exercise docs, metadata, and tests from 'problem-specifications'.
              Check or populate missing 'files' values for Concept/Practice Exercises from the track 'config.json'.
  uuid        Output new (version 4) UUIDs, suitable for the value of a 'uuid' key

Options for completion:
  -s, --shell <shell>          Choose the shell type (required)
                               Allowed values: b[ash], f[ish], z[sh]

[...]

```

So the main remaining step is to make `fetch-configlet` explain how to install completions. I'd like for us to explain how to install the bash and zsh completions on a not-just-for-this-session basis.

I think this PR doesn't block the configlet release, though.

PR #652, opened 2022-08-06
cli, completion: compress some `const` strings

### LINKED ISSUE #651: consider compressing large strings
It's easy to compress large strings at compile time and uncompress them at run time. This isn't that useful for configlet, but we could compress at least the embedded:

- help message
- completion scripts

Right now, this would reduce the binary size by only a couple of percent at best. But it's easy, has few disadvantages, and there may be other good candidates for compression later (at least the zsh and powershell completion scripts). So maybe it's worth doing.

ee7:
This uses [`supersnappy`](https://github.com/guzba/supersnappy), which is a zero-dependency pure-Nim
implementation of Google's [Snappy](https://github.com/google/snappy).

With it, it is straightforward to compress our larger `const` strings
at compile time and uncompress them at run time. For now, this commit
compresses:

- The `configlet --help` message
- The completion scripts

`supersnappy` is less than 500 lines of code, which is why compressing
these strings (which aren't so large) produces a slightly smaller
configlet binary, despite the added dependency.

We may end up adding [`zippy`](https://github.com/guzba/zippy) as a dependency if configlet learns
to make network requests (e.g. for the online `configlet lint` checks)
so that we can use gzip. But we can't use `zippy` at compile time, so it
isn't useful for this commit, even though the compression is better.
`zippy` is also a bigger dependency (about 40 KiB), so it wouldn't
produce a net saving for this commit anyway.

[`nim-snappy`](https://github.com/status-im/nim-snappy) is an alternative Snappy implementation, but it would
[add dependencies](https://github.com/status-im/nim-snappy/blob/7cb2e57a5861/snappy.nimble) on `faststreams`, `unittest2`, and `stew`.

Closes: #651

### DIFF
--- a/configlet.nimble
+++ b/configlet.nimble
@@ -19,10 +19,11 @@ bin           = @["configlet"]
 
 # Dependencies
 requires "nim >= 1.6.6"
-requires "cligen#b962cf8bc0be847cbc1b4f77952775de765e9689"    # 1.5.19 (2021-09-13)
-requires "jsony#2a2cc4331720b7695c8b66529dbfea6952727e7b"     # 1.1.3  (2022-01-03)
-requires "parsetoml#9cdeb3f65fd10302da157db8a8bac5c42f055249" # 0.6.0  (2021-06-07)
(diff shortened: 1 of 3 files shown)


ErikSchierboom (approved):
LGTM

Issue #642, opened 2022-08-04
completions: for `-e`, suggest only slugs that are in the `config.json` file


ee7:
Perhaps:
- if `jq` is installed, complete the slugs by parsing the track `config.json` file.
- otherwise, fall back to looking at directories in `exercises/{concept,practice}`. But note that `configlet fmt|generate|sync -e foo` won't work if `foo` isn't in the track `config.json` file.

ee7:
I've thought of the smarter way to fix this, which avoids the dependency on `jq`.

Is there a program that can parse JSON, that we know is installed on every user's machine at configlet completion time? Yes. Configlet itself.

We could add a subcommand that we omit from the `configlet --help` message:

```shell
configlet __complete
```

such that tab completion of an exercise slug:

```shell
configlet sync --track-dir foo --exercise [Tab]
```

internally calls:

```shell
configlet __complete sync --track-dir foo --exercise
```

which parses the `config.json` file in `foo` and returns the appropriate slugs (for both Concept and Practice Exercises for `fmt`, and just Practice Exercises for `sync`).

Low priority for now.

ErikSchierboom:
That _is_ a much better option. Neat.

Issue #637, opened 2022-08-04
completions: suggests incorrect `-e` slugs when `-t` is used


ee7:
Each completion script cleverly suggests exercise slugs when the `-e, --exercise` option is used. However, they always search the current directory, even when the `-t, --track-dir` is used. This means that it can e.g. suggest a slug that doesn't exist in the track directory that configlet is operating on.

I guess I'll call this a "bug". Can we look at the argument of `-t` when completing the argument of `-e`?

glennj:
Here's an example of a fish completion that respects previous args: https://unix.stackexchange.com/questions/722020/fish-completion-that-depends-on-other-flags

But as we're thinking about getting configlet itself to handle completions, this issue is no longer about fish/bash/zsh

ee7:
Thanks. For the fish completions I have this lying around in a branch from a couple of months ago:

```fish
function __fish_configlet_get_track_dir
  # If -t/--track-dir is on the command line, echoes its argument.
  set -l cmdline (commandline -p)
  if set -l match (string match -r -- "(-t|--track-dir) +([^ ]+)" $cmdline)
    echo $match[3]
    return 0
  else
    return 1
  end
end
```

which may or may not work. I didn't try the bash or zsh version, and yes, it might not be worth spending the time on, given:

> But as we're thinking about getting configlet itself to handle completions, this issue is no longer about fish/bash/zsh

To others: Glenn is referring to https://github.com/exercism/configlet/issues/642#issuecomment-1208263475.

Yes, I still think that's the most elegant way to handle completions in the long run. It means that a user can install a meta-completion script only once, and get updated completions for free every time that configlet is updated. That's very nice for a program like configlet, which is not installed by a package manager.

PR #653, opened 2022-08-07
completion: add zsh

### LINKED ISSUE #633: completions: add zsh
Follow-up from https://github.com/exercism/configlet/pull/629.

ee7:
Add a zsh completion script and the ability to run

```shell
configlet completion --shell zsh
```

to write that script to stdout.

This commit also bumps the version of the shellcheck action, which is
necessary to prevent the action from running `shellcheck` on the new zsh
script. Shellcheck does not support zsh, so CI would indicate an error
otherwise.

The bump must be to the latest commit - we already used the latest
release (1.1.0). An [upstream commit](https://github.com/ludeeus/action-shellcheck/commit/ceeca77f6538) that claims to exclude zsh
scripts seems to be insufficient - we need to use the `ignore_names`
input that [does not exist in the latest release](https://github.com/ludeeus/action-shellcheck/commit/c2b45ddc5f85).

Closes: #633

---

It seems like every tutorial for this is pretty bad.

Some links for the reviewer:

- [zsh completions for `pacman`](https://gitlab.archlinux.org/pacman/pacman/-/blob/master/scripts/completion/zsh_completion.in).
- [zsh completions for `git`](https://github.com/zsh-users/zsh/blob/master/Completion/Unix/Command/_git).
- https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org
- If you have zsh completions installed, it may or may not be helpful to look in `/usr/share/zsh/functions/Completion/Unix/` and `/usr/share/zsh/site-functions` for the completions for a program you know well.
- [`man zshcompsys`](https://man.archlinux.org/man/zshcompsys.1.en). Good luck.
- For installing `zsh` and enabling completions: https://wiki.archlinux.org/title/zsh#Command_completion

---

To test this PR, copy `configlet.zsh` to some directory, naming it `_configlet`:

```shell
mkdir ~/configlet-comp
cp completions/configlet.zsh ~/configlet-comp/_configlet
```

(Don't put it under `/tmp`, or you might get an error message about an insecure completions folder).

Then add that directory to your `fpath`. In `.zshrc`:

```shell
fpath=(~/configlet-comp $fpath)
```

And make sure that [completions are also enabled](https://wiki.archlinux.org/title/zsh#Command_completion) in that file:

```shell
autoload -Uz compinit promptinit
compinit
```

Then start a new zsh shell, and `configlet [Tab]` should work.

The underlying issue is that the default `fpath` includes just `/usr/local/share/zsh/site-functions` and directories in `/usr/share/zsh`. We can't just write, as we can for bash:

```shell
configlet completion -s bash | source
```

or add the completions to some standard user completions directory, like for fish:

```shell
configlet completion -s fish > ~/.config/fish/completions/configlet.fish
```

So I think we have to provide instructions to make the user save the output of `configlet completion --shell zsh` to either:

1. `/usr/local/share/zsh/site-functions/_configlet`
1. `/usr/share/zsh/site-functions/_configlet`
1. `~/foo/_configlet`, explaining how to add `~/foo` to their `$fpath`

where we probably should do the last, because:
- e.g. generally only packages installed with the package manager should put stuff in `/usr/share/zsh/site-functions`.
- the first two require root permissions.

### DIFF
--- a/completions/configlet.zsh
+++ b/completions/configlet.zsh
@@ -49,7 +49,7 @@ _configlet() {
     (fmt)
       _arguments "${_arguments_options[@]}" \
           "$_configlet_global_opts[@]" \
-          {-e,--exercise}'+:[exercise slug]' \
+          '(-e --exercise)'{-e+,--exercise=}'[exercise slug]:' \
           {-u,--update}'[Write changes]' \
           {-y,--yes}'[Auto-confirm update]' \
       ;;
@@ -61,7 +61,7 @@ _configlet() {
     (sync)
       _arguments "${_arguments_options[@]}" \
           "$_configlet_global_opts[@]" \
-          {-e,--exercise}'+:[exercise slug]' \
+          '(-e --exercise)'{-e+,--exercise=}'[exercise slug]:' \
           {-o,--offline}'[Do not update prob-specs cache]' \
           {-u,--update}'[Write changes]' \
           {-y,--yes}'[Auto-confirm update]' \
--- a/completions/configlet.zsh
+++ b/completions/configlet.zsh
@@ -73,7 +73,7 @@ _configlet() {
     (uuid)
       _arguments "${_arguments_options[@]}" \
           "$_configlet_global_opts[@]" \
-          {-n,--num}'+:[How many UUIDs]' \
+          '(-n --num)'{-n+,--num=}'[How many UUIDs]:' \
       ;;
   esac
 }
--- a/completions/configlet.zsh
+++ b/completions/configlet.zsh
@@ -14,11 +14,15 @@ _configlet() {
   local line
   local curcontext="$curcontext"
 
+  _configlet_global_opts=(
+    {-h,--help}'[Show help]'
+    '--version[Show version information]'
+    {-t,--track-dir}'+:[Select a track directory]'
+    {-v,--verbosity}'[Verbosity level]: :(quiet normal detailed)'
+  )
+
   _arguments "${_arguments_options[@]}" \
-      {-h,--help}'[Show help]' \
-      '--version[Show version information]' \
-      {-t,--track-dir}'+:[Select a track directory]' \
-      {-v,--verbosity}'[Verbosity level]: :(quiet normal detailed)' \
+      "$_configlet_global_opts[@]" \
       ":: :_configlet_commands" \
       "*::: :->configlet"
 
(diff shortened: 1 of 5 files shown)


completions/configlet.zsh
```diff
@@ -0,0 +1,114 @@
...
+    slugs+=( ${${slug_paths#./exercises/practice/}%-*-*} )
+    compadd "$@" -a slugs
+  }
+
+  _configlet_complete_practice_exercise_slug() {
+    local -a cmd slugs slug_paths
+    slug_paths=(./exercises/practice/*(/))
+    slugs=( ${${slug_paths#./exercises/practice/}%-*-*} )
+    compadd "$@" -a slugs
+  }
```
ee7: Yes, this could use a refactor.

Note that this can suggest invalid slugs, because it:

- doesn't respect `--track-dir` (see #637)
- or read the track `config.json` file (see #642)

Let's handle that later - we should fix it for the bash and fish completion scripts too.

completions/configlet.zsh
```diff
@@ -0,0 +1,114 @@
...
+          {-y,--yes}'[Auto-confirm update]' \
+          '--docs[Sync docs only]' \
+          '--filepaths[Populate .meta/config.json "files" entry]' \
+          '--metadata[Sync metadata only]' \
+          '--tests[For auto-confirming]: :(choose include exclude)' \
+      ;;
+    (uuid)
+      _arguments "${_arguments_options[@]}" \
+          "$_configlet_global_opts[@]" \
+          '(-n --num)'{-n+,--num=}'[How many UUIDs]:' \
```
ee7, line 107: > [`-optname+`](https://man.archlinux.org/man/zshcompsys.1.en#optname~2)
> The first argument may appear immediately after optname in the same word, or may appear as a separate word after the option. For example, `-foo+:...` specifies that the completed option and argument will look like either `-fooarg` or `-foo arg`.
> 
> [`-optname=`](https://man.archlinux.org/man/zshcompsys.1.en#optname~3)
> The argument may appear as the next word, or in same word as the option name provided that it is separated from it by an equals sign, for example `-foo=arg` or `-foo arg`.

completions/configlet.zsh
```diff
@@ -0,0 +1,114 @@
...
+}
+
+_configlet() {
+  typeset -a _arguments_options
+
+  if is-at-least 5.2; then
+    _arguments_options=(-s -S -C)
+  else
+    _arguments_options=(-s -C)
+  fi
```
ee7, line 29: From `man zshcompsys`:

> The options of `_arguments` have the following meanings:
> 
> [`-s`](https://man.archlinux.org/man/zshcompsys.1.en#s)
> Enable option stacking for single-letter options, whereby multiple single-letter options may be combined into a single word. For example, the two options `-x` and `-y` may be combined into a single word `-xy`. By default, every word corresponds to a single option name (`-xy` is a single option named `xy`).
> Options beginning with a single hyphen or plus sign are eligible for stacking; words beginning with two hyphens are not.
> 
> Note that `-s` after `--` has a different meaning, which is documented in the segment entitled `Deriving spec forms from the help output'.
> 
> [`-C`](https://man.archlinux.org/man/zshcompsys.1.en#C)
> Modify the `curcontext` parameter for an action of the form `->state`. This is discussed in detail below.
> 
> [`-S`](https://man.archlinux.org/man/zshcompsys.1.en#S)
> Do not complete options after a `--` appearing on the line, and ignore the `--`. For example, with `-S`, in the line
> 
> ```text
> foobar -x -- -y
> ```
> 
> the `-x` is considered an option, the `-y` is considered an argument, and the `--` is considered to be neither.

completions/configlet.zsh
```diff
@@ -0,0 +1,114 @@
...
+
+  local line
+  local curcontext="$curcontext"
```
ee7, line 32: From `man zshcompsys`:

> The option `-C` tells `_arguments` to modify the `curcontext` parameter for an action of the form `->state`. This is the standard parameter used to keep track of the current context. Here it (and not the context array) should be made local to the calling function to avoid passing back the modified value and should be initialised to the current value at the start of the function:
> 
> ```text
> local curcontext="$curcontext"
> ```

ErikSchierboom (approved):
Just tested it and it works nicely!

completions/configlet.zsh
ee7: Refactored somewhat in https://github.com/exercism/configlet/pull/653/commits/833522bc5c4bec947f8cdb5357ea2cd99e173deb

PR #656, opened 2022-08-14
completions(fish): remove invalid file completions

ee7:
With the fish completions, writing a command or option that does not
take an argument and then pressing tab would suggest files and
directories:

```console
$ bin/configlet sync [Tab]
bin/                config.json      docs/       LICENSE   README.md
CODE_OF_CONDUCT.md  config/          exercises/  mix.exs   reference/
concepts/           CONTRIBUTING.md  HELLO.md    mix.lock
```

even though adding such an argument would produce an error.

Remove those completions across the whole command, and then add them
only where we want them (directory completions for `-t`/`--track-dir`). This
also improves the completions for `-t`/`--track-dir`, since it would
previously suggest both files and directories.

---

Docs:
- https://fishshell.com/docs/current/cmds/complete.html
- https://fishshell.com/docs/current/completions.html

### DIFF
--- a/completions/configlet.fish
+++ b/completions/configlet.fish
@@ -1,28 +1,32 @@
+# disable file completions everywhere
+complete -c configlet -f
+
 # global options
-complete -c configlet -s h -l help -f -d "Show help"
-complete -c configlet      -l version -f -d "Show version info"
-complete -c configlet -s t -l track-dir -d "Select a track directory"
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
I know nothing really about fish (_insert corny joke here_), so I'll just trust on your expertise.

PR #659, opened 2022-08-18
completions(bash, fish): add zsh completion for -s/--shell

ee7:
Commit https://github.com/exercism/configlet/commit/e4c86435d284c4ba7da76ae00cc80387ceb865dc added a completion script for zsh, but forgot to add
zsh as a completion for the value of -s/--shell in the bash and fish
completion scripts.

---

"Some completion scripts had incomplete completions for `completion`. Complete them."


ErikSchierboom (approved):
:D

PR #664, opened 2022-09-27
generate: allow a concept slug to contain digits

### LINKED ISSUE #663: Generate: slugs with numbers break generation
Hi!
I was trying to do the equivalent of [this PR](https://github.com/exercism/elixir/pull/1151) in the Elm track, but I ran into an issue.
We have concepts named `basics-1` and `basics-2`, which makes `configlet generate` break.

For the concept exercise `lucians-luscious-lasagna`, I wrote the `introduction.md.tpl` as

```
# Introduction

%{concept: basics-1}
```

But after generation the `introduction.md` is

```
# Introduction

1}
```

The culprit probably is
https://github.com/exercism/configlet/blob/d31e569714c1562eeca76004d5b49d3500ddfc5d/src/generate/generate.nim#L100-L102

It seems that it would be easy enough to support numbers, is that an option?

ee7:
A concept slug is [allowed to contain digits](https://github.com/exercism/docs/blob/ba3eba34d334/building/configlet/lint.md#glossary):

> The `concepts[].slug` value must be a kebab-case string with length <= 255
>
> [...]
>
> kebab-case string: a string that [...] must match the regular expression: `^[a-z0-9]+(-[a-z0-9]+)*$`

but `configlet generate` did not support such a slug.

With this commit, `configlet generate` now behaves as expected for an
`introduction.md.tpl` file like:

```markdown
# Introduction

%{concept: basics-1}
```

Fixes: #663

### DIFF
--- a/src/generate/generate.nim
+++ b/src/generate/generate.nim
@@ -93,13 +93,24 @@ proc generateIntroduction(trackDir: Path, templatePath: Path,
   var headingLevel = 1
   var linkDefs = newSeq[string]()
   while i < content.len:
-    var conceptSlug = Slug ""
+    var cs = "" # Buffer for a concept slug that may not be valid.
     # Here, we implement the syntax for a placeholder as %{concept:some-slug}
     # where we allow spaces after the opening brace, around the colon,
-    # and before the closing brace. The slug must be in kebab-case.
+    # and before the closing brace.
+    #
+    # The slug must be in kebab-case, matching the regular expression:
+    #
+    #   ^[a-z0-9]+(-[a-z0-9]+)*$
+    #
+    # `configlet lint` enforces this for slugs in the track config.json file.
     if scanp(content, i,
              "%{", *{' '}, "concept", *{' '}, ':', *{' '},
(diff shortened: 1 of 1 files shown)


src/generate/generate.nim
```diff
@@ -93,13 +93,24 @@ proc generateIntroduction(trackDir: Path, templatePath: Path,
...
+    #
+    # The slug must be in kebab-case, matching the regular expression:
+    #
+    #   ^[a-z0-9]+(-[a-z0-9]+)*$
+    #
+    # `configlet lint` enforces this for slugs in the track config.json file.
     if scanp(content, i,
              "%{", *{' '}, "concept", *{' '}, ':', *{' '},
-             +{'a'..'z', '-'} -> conceptSlug.add($_), *{' '}, '}'):
+             +{'a'..'z', '0'..'9', '-'} -> cs.add($_), *{' '}, '}'):
```
ee7, line 108: Be permissive here to keep the pattern simple, and get a slightly better error message if a user typos a `-` in their placeholder.

ErikSchierboom (approved):

PR #667, opened 2022-10-06
github: set draft release title and notes

ee7:
Before this commit, the draft release was created without a title or release notes. To give a better starting point for a maintainer preparing a release, set the title to the current tag, and set the initial notes to a list of commits since the previous official release.

The checkout Action only fetches one commit by default, so also change the build workflow to fetch all git history.

The `gh release create` command does have a `--generate-notes` option, but the implementation in this commit gives full control over the format, and includes commit refs.

---

Tested here: https://github.com/ee7/exercism-configlet/releases/tag/github-set-draft-release-title-and-notes

### DIFF
--- a/.github/bin/publish-release
+++ b/.github/bin/publish-release
@@ -7,6 +7,8 @@ if ! gh release view "${build_tag}"; then
   # tag for the previous official configlet release:
   # - is the most recent (or second most recent) local tag.
   # - exists locally (so we support creating test releases on a forked repo).
+  # The `gh release create` command does have a `--generate-notes` option,
+  # but the below gives full control over the format, and includes commit refs.
   previous_release_sha="$(gh api graphql --jq 'recurse | strings' -f query='
     {
       repository(owner: "exercism", name: "configlet") {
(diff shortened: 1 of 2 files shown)


.github/bin/publish-release
```diff
@@ -3,7 +3,24 @@
...
+  previous_release_sha="$(gh api graphql --jq 'recurse | strings' -f query='
+    {
+      repository(owner: "exercism", name: "configlet") {
+        latestRelease {
+          tagCommit {
+            oid
+          }
+        }
+      }
+    }')"
```
ee7, line 21: It looks like this is the best way to get the ref of the commit corresponding to the latest official configlet release.

With the REST endpoint it seems like it requires two requests:

```shell
previous_release="$(gh api repos/exercism/configlet/releases/latest --jq '.tag_name')"
previous_release_sha="$(gh api "repos/exercism/configlet/commits/${previous_release}" --jq '.sha')"
```

And there we have to be careful to get the sha corresponding to the commit itself, rather the annotated tag (which is an object with its own sha).

ErikSchierboom (approved):
Great improvement!

PR #678, opened 2022-10-18
helpers: refactor common `walkDir` functionality

ee7:
Follow-up to 2a04e8b6f933


ErikSchierboom (approved):
Nice

ee7:
Note that Nim 2.0 will include its own:

```nim
type
  Path* = distinct string
```

(See https://github.com/nim-lang/Nim/commit/3c12b72168d9fe9916f471fa17d2c24c52b33066).

Issue #361, opened 2021-06-17
fetch-configlet.ps1: don't require Windows


ee7:
From https://github.com/exercism/problem-specifications/pull/1799#discussion_r653396492:
> In the future, we could consider modifying the PowerShell script to also download the binaries for non-Windows OS'es (PowerShell is cross-platform nowadays).

This seems like a good idea - it's more complex to explain to users that the bash script works on macOS/Linux/Windows, but the PowerShell script is Windows-only.

Might be a good idea to test these scripts in CI first, see: https://github.com/exercism/configlet/issues/121

PR #690, opened 2022-10-26
.github: workflows: add workflow to test `fetch-configlet` scripts

### LINKED ISSUE #121: fetch-configlet scripts: Test during CI
We could run these scripts in CI and check that they actually download the latest release to a set location.

ee7:
The fetch-configlet scripts were previously not tested during CI. Test them.

The `fetch-configlet.ps1` script currently always fetches the Windows release asset:

https://github.com/exercism/configlet/blob/7dd6784061745316df5fe71588db5c8a61477103/scripts/fetch-configlet.ps1#L16

so this workflow only tests that script on Windows. (Issue #361 tracks this low-priority problem).

Closes: #121

### DIFF
--- a/.github/workflows/fetch-configlet.yml
+++ b/.github/workflows/fetch-configlet.yml
@@ -14,11 +14,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [linux, macOS, windows]
+        include:
+          - os: linux
+            runs-on: ubuntu-22.04
+
+          - os: mac
+            runs-on: macos-12
+
+          - os: windows
+            runs-on: windows-2022
 
     name: fetch-configlet-${{ matrix.os }}
-    runs-on: ${{ matrix.os == 'linux' && 'ubuntu-22.04' ||
-                (matrix.os == 'macOS' && 'macos-12' || 'windows-2022') }}
+    runs-on: ${{ matrix.runs-on }}
     steps:
       - name: Checkout
         uses: actions/checkout@e2f20e631ae6d7dd3b768f56a5d2af784dd54791
--- a/.github/workflows/fetch-configlet.yml
+++ b/.github/workflows/fetch-configlet.yml
@@ -1,0 +1,44 @@
+name: Fetch configlet
+
+on:
+  push:
+    paths:
+      - scripts/fetch-configlet*
+  pull_request:
+    paths:
+      - scripts/fetch-configlet*
+  workflow_dispatch:
+
+jobs:
+  fetch_configlet:
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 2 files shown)


ee7:
See workflow logs:

- [Linux fetch-configlet](https://github.com/exercism/configlet/actions/runs/3331595621/jobs/5511543416#step:4:8)
- [macOS fetch-configlet](https://github.com/exercism/configlet/actions/runs/3331595621/jobs/5511543546#step:4:8)
- [Windows fetch-configlet](https://github.com/exercism/configlet/actions/runs/3331595621/jobs/5511543671#step:4:8)
- [Windows fetch-configlet.ps1](https://github.com/exercism/configlet/actions/runs/3331595621/jobs/5511543671#step:5:7)

.github/workflows/fetch-configlet.yml
```diff
@@ -0,0 +1,44 @@
...
+jobs:
+  fetch_configlet:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [linux, macOS, windows]
+
+    name: fetch-configlet-${{ matrix.os }}
+    runs-on: ${{ matrix.os == 'linux' && 'ubuntu-22.04' ||
+                (matrix.os == 'macOS' && 'macos-12' || 'windows-2022') }}
```
ErikSchierboom: What is the reason you do it like this an not have the matrix specifc the runners? That you can use `matrix.os == 'windows'` later on?

ee7: If we use 

```yaml
strategy:
  matrix:
    os: [ubuntu-22.04, macos-12, windows-2022]

  runs-on: ${{ matrix.os }}
```

then we either have to later use conditions like

```
  if: matrix.os == 'windows-2022'
```

(which means there's no single source of truth for the windows version), or

```
  if: runner.os == 'Windows'
```

which is less clear because:

- we don't set `runner.os` ourselves
- `runner.os` can't be used as a condition everywhere in a workflow
- `runner.os` can be confused with `matrix.os`, which I was lazy in naming

But the other reason is consistency: I was leaning towards making the other workflows use the same matrix logic.

Thoughts?

ErikSchierboom: See https://github.com/exercism/configlet/pull/692/files/5e7dffebd8f7490df372d3fd4cbd05e03416a2b9#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957f for my thoughts

ee7: Either way, I shouldn't have proposed a new matrix usage in this PR. Please take another look.

ErikSchierboom (approved):

Issue #130, opened 2021-01-22
Build: Restructure workflows


ee7:
In the long-term I think we should make some deeper change. For example, make the current job in `build.yml` contain only the "Create artifact" and "Publish release" steps, but depend on an earlier job that runs the tests. Currently, we assume that the `tests.yml` build process is identical to that in `build.yml` - the job in `tests.yml` does not test the exact binary that is released (even if it could be bit-identical).

_Originally posted by @ee7 in https://github.com/exercism/canonical-data-syncer/issues/129#issuecomment-765322244_

ee7:
Reusable workflows may be good for the refactoring we want.

We wrote our main testing and build workflows a relatively long time ago, considerably before the time that GitHub added reusable workflows, and made them more composable. In particular, the ability to call a reusable workflow from a matrix or another reusable workflow is very recent.

Timeline:

- [2019-11-11 GitHub Actions is released](https://github.blog/changelog/2019-11-11-github-actions-is-generally-available/)
- [2021-10-05 reusable workflows are beta](https://github.blog/changelog/2021-10-05-github-actions-dry-your-github-actions-configuration-by-reusing-workflows/)
- [2021-11-24 reusable workflows are stable](https://github.blog/changelog/2021-11-24-github-actions-reusable-workflows-are-generally-available/)
- [2022-01-25 reusable workflows can be referenced locally](https://github.blog/changelog/2022-01-25-github-actions-reusable-workflows-can-be-referenced-locally/)
- [2022-06-10 inputs unified across manual and reusable workflows](https://github.blog/changelog/2022-06-10-github-actions-inputs-unified-across-manual-and-reusable-workflows/)
- [2022-08-22 reusable workflows can be called from a matrix, and other reusable workflows](https://github.blog/changelog/2022-08-22-github-actions-improvements-to-reusable-workflows-2/)
- [2022-10-13 structure of reusable workflows is better shown in navigation](https://github.blog/changelog/2022-10-13-design-improvements-to-github-actions-navigation/)

PR #692, opened 2022-10-27
.github(workflows): refactor matrix

ee7:
I think this is a bit cleaner. Thoughts?

I use this trick quite often, including in the `exercism/nim` repo [here](https://github.com/exercism/nim/blob/90155db2f99b2d76e99ac2dc18ca8f7967f60e62/.github/workflows/exercises.yml#L8-L26). But it's possible that there's now a more idiomatic way than both.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,26 +11,19 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target:
+        include:
           - os: linux
             arch: 64bit
+
           - os: mac
             arch: 64bit
+
           - os: windows
             arch: 64bit
-        include:
-          - target:
-              os: linux
-            builder: ubuntu-22.04
-          - target:
-              os: mac
-            builder: macos-12
-          - target:
-              os: windows
-            builder: windows-2022
 
-    name: "${{ matrix.target.os }}-${{ matrix.target.arch }}"
-    runs-on: ${{ matrix.builder }}
+    name: "${{ matrix.os }}-${{ matrix.arch }}"
+    runs-on: ${{ matrix.os == 'linux' && 'ubuntu-22.04' ||
+                (matrix.os == 'mac' && 'macos-12' || 'windows-2022') }}
     steps:
       - name: Checkout code
         uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
@@ -38,7 +31,7 @@ jobs:
           fetch-depth: 0 # Allows using `git log` to set initial release notes.
 
       - name: On Linux, install musl
-        if: matrix.target.os == 'linux'
+        if: matrix.os == 'linux'
         run: ./.github/bin/linux-install-build-tools
 
       - name: Install Nim
@@ -56,8 +49,8 @@ jobs:
         shell: bash
         run: ./.github/bin/create-artifact
         env:
-          OS: ${{ matrix.target.os }}
-          ARCH: ${{ matrix.target.arch }}
+          OS: ${{ matrix.os }}
+          ARCH: ${{ matrix.arch }}
 
       - name: Publish release
         shell: bash
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,17 +13,19 @@ jobs:
       matrix:
         include:
           - os: linux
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -11,34 +11,27 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target:
+        include:
           - os: linux
             arch: 64bit
+
           - os: mac
             arch: 64bit
```
ee7: We'll change these to `macOS` and `x86_64` later. Right now, it would cause the release assets to be created with different names, which would break the fetch-configlet scripts.

.github/workflows/tests.yml
```diff
@@ -16,32 +16,25 @@ jobs:
...
-          - target:
-              os: windows
-            builder: windows-2022
-
-    name: "${{ matrix.target.os }}-${{ matrix.target.arch }}"
-    runs-on: ${{ matrix.builder }}
+
+    name: "${{ matrix.os }}-${{ matrix.arch }}"
+    runs-on: ${{ matrix.os == 'linux' && 'ubuntu-22.04' ||
+                (matrix.os == 'mac' && 'macos-12' || 'windows-2022') }}
```
ErikSchierboom: Why not add an additional `runner` key? I don't really like this approach

ee7: I haven't found an approach that I love yet.

We could do

```yaml
    strategy:
      fail-fast: false
      matrix:
        include:
          - os: linux
            arch: 64bit
            runner: ubuntu-22.04

          - os: mac
            arch: 64bit
            runner: macos-12

          - os: windows
            arch: 64bit
            runner: windows-2022
```

but that seems ugly because:

1. We have to ensure that we don't accidentally write some mismatch like:

```yaml
os: linux
arch: x86_64
runner: macos-12
```

2. When we want to add more jobs to the matrix, there's no single source of truth for the Ubuntu version in every Linux job:

```yaml
    strategy:
      fail-fast: false
      matrix:
        include:
          - os: linux
            arch: x86_64
            runner: ubuntu-22.04

          - os: mac
            arch: x86_64
            runner: macos-12

          - os: windows
            arch: x86_64
            runner: windows-2022

          - os: linux
            arch: cross-compile-arm64
            runner: ubuntu-22.04
```

3. Nitpicking the name, `matrix.runner` could be confused with the [`runner` context](https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context)

ErikSchierboom: > We have to ensure that we don't accidentally write some mismatch like:

But that's true of your suggestion solution too.

> When we want to add more jobs to the matrix, there's no single source of truth for the Ubuntu version in every Linux job:

That's true, but I find that unlikely. You could even see it as a feature, where you _can_ deviate if you want to.

> Nitpicking the name, matrix.runner could be confused with the [runner context](https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context)

Valid point. Maybe rename to `runs_on`

ee7: OK, let's try the layout from my previous post. Please review again.

> But that's true of your suggestion solution too.

True, but we always need to check only one place.
 
> That's true, but I find that unlikely. 

We'll build for more platforms eventually. GitHub Actions only provides `x86_64` right now for free when not self-hosting, so assuming that we don't start using another CI provider, we need cross-compilation. Cross-compilation requires a different toolchain to the native build, so it seems reasonable to do it in parallel in a separate job.

There might be a good refactoring involving reusable workflows, but I think adding a new job to the matrix is reasonable during a first implementation, right?

> You could even see it as a feature, where you _can_ deviate if you want to.

I suppose. But I see **that** as more unlikely.

> Valid point. Maybe rename to `runs_on`

Good idea. Changed to `runs-on`, to match the style.

ErikSchierboom (approved):

.github/workflows/tests.yml
ee7: @ee7 wrote:
> There might be a good refactoring involving reusable workflows,

I'll track later workflow refactoring in https://github.com/exercism/configlet/issues/130.

ee7:
I have double-checked the build workflow by triggering it on my fork.

Issue #325, opened 2021-05-11
Add common proc to write errors


ee7:
We should add an exported `writeError` proc or similar.

The one in `cli.nim` also writes the help text, which isn't always what we want.

Issue #123, opened 2021-01-15
Refactor: Raise exceptions, rather than using a deep `quit`


ee7:
This isn't just a Nim thing.

Background:
https://forum.nim-lang.org/t/4042

As an example, the below code doesn't echo:
```Nim
try:
  quit(0)
finally:
  echo "hi"
```

(Try in the Nim playground [here](https://play.nim-lang.org/#ix=2M5W)).

For one better design, see [`check_uuids.nim`](https://github.com/exercism/nim/blob/be1382016edd67eb04f35a4468286517135b6bd2/_test/check_uuids.nim#L349-L355) from the `exercism/nim` repo.

PR #699, opened 2022-10-28
sync: improve error handling

ee7:
Replace:

- `showError` with `errorAndHelp`
- (`stderr.writeLine msg` then `quit 1`) with `error`

Refs: #123
Refs: #325

---

Working towards the approach that I use elsewhere, including:

https://github.com/exercism/configlet/blob/b6fff14959b095b77c982d32fbc3941d909a1845/bin/bump_version.nim#L17-L19

### DIFF
--- a/src/cli.nim
+++ b/src/cli.nim
@@ -2,6 +2,16 @@ import std/[os, parseutils, strformat, strutils, terminal]
 import pkg/[cligen/parseopt3, supersnappy]
 
 type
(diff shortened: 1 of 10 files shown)


ErikSchierboom (approved):
Looks good!

ErikSchierboom:
@ee7 Is this something you'd still like to do? It looks like a nice improvement

ee7:
It's probably an improvement overall, but please take over if you want it.

Issue #24, opened 2020-10-09
Support more platforms


ErikSchierboom:
In particular, we should be supporting:

- Linux 32-bit
- Windows 32-bit
- ARM

ee7:
All we can say from the current configlet download data is that it's downloaded about 3x more often on macOS than Windows. We can't easily know the Linux numbers, since the vast majority are CI downloads.

But it's interesting to look at the stats for [the most recent release of `exercism/cli`](https://github.com/exercism/cli/releases/tag/v3.0.13), which gives us download data from the last 2.5 years. 

```shell
gh release view v3.0.13 --repo exercism/cli --json assets \
  | jq -r '.assets | .[] | [.name,.downloadCount] | @tsv' \
  | column -t -N 'Asset name,Download count' -R 2 -o ' | '
```

| Asset name                            | Download count |
| ------------------------------------- | --------------:|
| exercism-3.0.13-darwin-i386.tar.gz    |           1613 |
| exercism-3.0.13-darwin-x86_64.tar.gz  |           3187 |
| exercism-3.0.13-freebsd-i386.tar.gz   |             56 |
| exercism-3.0.13-freebsd-x86_64.tar.gz |            275 |
| exercism-3.0.13-linux-armv5.tar.gz    |            211 |
| exercism-3.0.13-linux-armv6.tar.gz    |            726 |
| exercism-3.0.13-linux-i386.tar.gz     |            342 |
| exercism-3.0.13-linux-ppc64.tar.gz    |            153 |
| exercism-3.0.13-linux-x86_64.tar.gz   |          24161 |
| exercism-3.0.13-openbsd-i386.tar.gz   |             52 |
| exercism-3.0.13-openbsd-x86_64.tar.gz |            212 |
| exercism-3.0.13-windows-i386.zip      |            683 |
| exercism-3.0.13-windows-x86_64.zip    |           7053 |
| exercism-freebsd-32bit.tgz            |             44 |
| exercism-freebsd-64bit.tgz            |            155 |
| exercism-linux-32bit.tgz              |            225 |
| exercism-linux-64bit.tgz              |          21820 |
| exercism-linux-ppc64.tgz              |             70 |
| exercism-mac-32bit.tgz                |            191 |
| exercism-mac-64bit.tgz                |           3938 |
| exercism-openbsd-32bit.tgz            |             33 |
| exercism-openbsd-64bit.tgz            |            218 |
| exercism-windows-32bit.zip            |           1907 |
[...]

PR #705, opened 2022-11-05
.github, fetch-configlet: rename release assets

### LINKED ISSUE #363: build: change release asset names
Currently, the release assets are named:
```
configlet-linux-64bit.tgz
configlet-mac-64bit.tgz
configlet-windows-64bit.zip
```

See: [configlet releases](https://github.com/exercism/configlet/releases) and [`.github/bin/create-artifact`](https://github.com/exercism/configlet/blob/main/.github/bin/create-artifact).

But this doesn't allow https://github.com/exercism/configlet/issues/24 very neatly. If we add `arm64` then just `64bit` is not enough, and `64bit-arm` is a bad suffix in my opinion (just `arm` can refer to the 32-bit architecture, and `64bit` isn't explicitly `x86-64`).

Some possible improvements:
- Use `x86-64` instead of `64bit`
- `macOS` instead of `mac`
- Include the configlet version in the asset name

In the long-term, I'd suggest something like:
```
configlet_4.1.0_Linux_arm64.tgz
configlet_4.1.0_Linux_i386.tgz
configlet_4.1.0_Linux_powerpc64.tgz
configlet_4.1.0_Linux_x86-64.tgz
configlet_4.1.0_macOS_arm64.tgz
configlet_4.1.0_macOS_x86-64.tgz
configlet_4.1.0_Windows_i386.zip
configlet_4.1.0_Windows_x86-64.zip
configlet_4.1.0_FreeBSD_x86-64.tgz   
configlet_4.1.0_OpenBSD_x86-64.tgz  
```

This has a few things that I don't really expect to release for, just to illustrate.

This would also require changes to the `fetch-configlet` scripts.

We can also consider:
- Adding `_musl` to the linux asset names.
- Using `.tar.gz` instead of `.tgz`
- ~Using `.tar.xz`~
- ~Using zstd~
- The choice between `x86-64`, `x86_64`, `amd64` and `x64`
- The choice between `arm64` and `aarch64`
- The choice between `i386`, `i686` and `x86`

For inspiration, we can look how bigger projects refer to architectures.

Assorted examples:
- https://github.com/cli/cli/releases
- https://ziglang.org/download/0.8.0/release-notes.html#Tier-2-Support
- https://ziglang.org/download/
- https://man.sr.ht/builds.sr.ht/compatibility.md
- https://github.com/BurntSushi/ripgrep/releases
- https://github.com/VSCodium/vscodium/releases
- https://github.com/syncthing/syncthing/releases
- https://github.com/gohugoio/hugo/releases
- https://github.com/electron/electron/releases

ee7:
Summary:

- Add version string
- Delimit with underscore, not hyphen
- Rename `64bit` to `x86-64`
- Rename `32bit` to `i386` (note that we do not have 32-bit releases yet anyway)
- Rename `tgz` to `tar.gz`
- Rename `mac` to `macos`

The configlet release assets have always had architecture-ambiguous names:

```text
configlet-linux-64bit.tgz
configlet-mac-64bit.tgz
configlet-windows-64bit.zip
```

This naming format is old (added by https://github.com/exercism/configlet/commit/d4c6e26836a5d74e6fa9698498c96fee7fdee0a5 in the `canonical-data-syncer` days), and was ultimately inheirited from:

- the `exercism/configlet-v2` releases (https://github.com/exercism/v2-configlet/releases)
- the `exercism/cli` releases (see https://github.com/exercism/cli/issues/700#issuecomment-529013140 and [cli-v3.0.12](https://github.com/exercism/cli/releases/tag/v3.0.12)).

The ambiguous names haven't mattered much so far, because configlet has only released for `x86-64`. However, I'm getting closer to adding releases for other 64bit architectures, so we finally need to change the names.

For the rationale being the new naming format, see my (hopefully sufficiently exhaustive) bikeshedding in https://github.com/exercism/configlet/issues/363#issuecomment-1292500893.

Before this commit, the release assets were named like:

```text
configlet-linux-64bit.tgz
configlet-mac-64bit.tgz
configlet-windows-64bit.zip
configlet_4.0.0-beta.7_checksums_sha256.txt
```

With this commit, the next release will have assets named:

```text
configlet_4.0.0-beta.8_checksums_sha256.txt
configlet_4.0.0-beta.8_linux_x86-64.tar.gz
configlet_4.0.0-beta.8_macos_x86-64.tar.gz
configlet_4.0.0-beta.8_windows_x86-64.zip
```

where the archive names now match the naming format for the existing checksums file.

Closes: #363
Refs: #24

---

I've tried to convince myself that the naming format with underscores everywhere but architecture and version string (where semver mandates a hyphen):

```
configlet_4.0.0-beta.7_linux_x86-64.tar.gz
```

is a better format than hyphens everywhere but architecture:

```
configlet-4.0.0-beta.7-linux-x86_64.tar.gz
```

or hyphens everywhere:

```
configlet-4.0.0-beta.7-linux-x86-64.tar.gz
```

or underscores everywhere possible:

```
configlet_4.0.0-beta.7_linux_x86_64.tar.gz
```

Even though I don't love that:

-  It ignores an existing convention that "hyphens separate words, and underscores delimit within a word" (try double clicking on the x86 in `x86_64` and `x86-64` if you aren't aware of this - the first selects the whole thing, the second doesn't). Note that double-clicking `beta` in `4.0.0-beta.7` won't select the whole version, and the same is true of `4.0.0_beta.7` even if semver permitted that, so us using underscores everywhere possible doesn't achieve a naming format where double-clicking on part of the name always performs tokenization. 
- `uname -m` returns `x86_64`, not `x86-64`. It is still referred to as `x86-64` in plenty of other places (e.g. https://en.wikipedia.org/wiki/X86-64)
- If we pretend that prerelease versions don't exist, and applications with more than one word don't exist, I probably prefer:

```
configlet-4.0.0-linux-x86_64.tar.gz
```
 
to this PR's:

```
configlet_4.0.0_linux_x86-64.tar.gz
```

---

To allow the `fetch-configlet` workflow to pass, I have added assets with the new naming format to the [configlet 4.0.0-beta.7 release](https://github.com/exercism/configlet/releases/tag/4.0.0-beta.7) - they're copies of the existing assets, with the same checksums.

Before we create a release that includes this PR, we need to:

- update the `fetch-configlet` script used by the org-wide configlet workflow.
- and create and merge a PR on every track to update the track-level `fetch-configlet{.ps1}` scripts. We can do this mass PR either:
   - manually
   - or by adding `fetch-configlet` to `org-wide-files` (tracked in https://github.com/exercism/org-wide-files/issues/262)

But for some period of time, we should keep assets with the old naming format in releases. Then an out-of-date `fetch-configlet` script will continue to work.

### DIFF
--- a/.github/bin/create-artifact
+++ b/.github/bin/create-artifact
@@ -10,7 +10,7 @@ case "${OS}" in
     artifact_file="${artifacts_dir}/${binary_name}-${OS}-${ARCH}.zip"
     7z a "${artifact_file}" "${binary_name}.exe"
     ;;
-  linux | mac)
+  linux | macos)
     artifact_file="${artifacts_dir}/${binary_name}-${OS}-${ARCH}.tgz"
     tar -cvzf "${artifact_file}" "${binary_name}"
     ;;
--- a/.github/bin/create-artifact
+++ b/.github/bin/create-artifact
@@ -11,7 +11,7 @@ case "${OS}" in
     7z a "${artifact_file}" "${binary_name}.exe"
     ;;
   linux | macos)
-    artifact_file="${artifacts_dir}/${binary_name}-${OS}-${ARCH}.tgz"
+    artifact_file="${artifacts_dir}/${binary_name}-${OS}-${ARCH}.tar.gz"
     tar -cvzf "${artifact_file}" "${binary_name}"
     ;;
 esac
--- a/.github/bin/create-artifact
+++ b/.github/bin/create-artifact
@@ -7,11 +7,11 @@ binary_name='configlet'
 
 case "${OS}" in
   windows)
-    artifact_file="${artifacts_dir}/${binary_name}-${OS}-${ARCH}.zip"
+    artifact_file="${artifacts_dir}/${binary_name}_${OS}_${ARCH}.zip"
     7z a "${artifact_file}" "${binary_name}.exe"
     ;;
   linux | macos)
-    artifact_file="${artifacts_dir}/${binary_name}-${OS}-${ARCH}.tar.gz"
+    artifact_file="${artifacts_dir}/${binary_name}_${OS}_${ARCH}.tar.gz"
     tar -cvzf "${artifact_file}" "${binary_name}"
     ;;
 esac
--- a/.github/bin/create-artifact
+++ b/.github/bin/create-artifact
@@ -4,14 +4,15 @@ artifacts_dir='artifacts'
 mkdir -p "${artifacts_dir}"
 
 binary_name='configlet'
+build_tag="${GITHUB_REF_NAME}"
 
 case "${OS}" in
   windows)
-    artifact_file="${artifacts_dir}/${binary_name}_${OS}_${ARCH}.zip"
+    artifact_file="${artifacts_dir}/${binary_name}_${build_tag}_${OS}_${ARCH}.zip"
     7z a "${artifact_file}" "${binary_name}.exe"
     ;;
   linux | macos)
-    artifact_file="${artifacts_dir}/${binary_name}_${OS}_${ARCH}.tar.gz"
+    artifact_file="${artifacts_dir}/${binary_name}_${build_tag}_${OS}_${ARCH}.tar.gz"
     tar -cvzf "${artifact_file}" "${binary_name}"
     ;;
 esac
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,15 +14,15 @@ jobs:
         include:
           - os: linux
             runs-on: ubuntu-22.04
-            arch: 64bit
+            arch: x86-64
 
           - os: macos
             runs-on: macos-12
-            arch: 64bit
+            arch: x86-64
 
           - os: windows
             runs-on: windows-2022
-            arch: 64bit
+            arch: x86-64
 
     name: "${{ matrix.os }}-${{ matrix.arch }}"
     runs-on: ${{ matrix.runs-on }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
             runs-on: ubuntu-22.04
             arch: 64bit
 
-          - os: mac
+          - os: macos
             runs-on: macos-12
             arch: 64bit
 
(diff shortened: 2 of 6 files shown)


ErikSchierboom (approved):
> With this commit, the next release will have assets named:

Those look good to me.
We can't merge this until we've handled the syncing of the scripts, right?

ee7:
> We can't merge this until we've handled the syncing of the scripts, right?

We can - it's just that we need to manually add assets with the old names before undrafting the release.

I could have made the build workflow automatically upload the duplicate assets, but:

1. I'm hoping we won't go many releases without updating the fetch scripts
1. I think I'd prefer to remove the assets with the old name when the release is no longer the latest one
1. (Partly continuing from point 2), I think I'd prefer for the checksums file to only reference the newly named assets

I'll also remove the **newly** named assets from `4.0.0-beta.7` after we release `4.0.0-beta.8`.

ErikSchierboom:
Okay sure

ee7:
@ee7 wrote:

> I'll also remove the **newly** named assets from `4.0.0-beta.7` after we release `4.0.0-beta.8`.

Done.

`fetch-configlet` only ever fetches assets from the latest release. So the point is that every release will eventually only use one archive naming format, with the checksums file referencing only the assets present, with assets uploaded only by GitHub Actions. Rather than the strange looking:

```shell
gh release view -R exercism/configlet \
    --json assets \
    --jq '.assets.[] | {updatedAt,name}' \
    | jq
```

```json
{
  "name": "configlet-linux-64bit.tgz",
  "updatedAt": "2022-10-27T21:00:06Z"
}
{
  "name": "configlet-mac-64bit.tgz",
  "updatedAt": "2022-10-27T21:00:35Z"
}
{
  "name": "configlet-windows-64bit.zip",
  "updatedAt": "2022-10-27T21:01:34Z"
}
{
  "name": "configlet_4.0.0-beta.7_checksums_sha256.txt",
  "updatedAt": "2022-10-27T21:01:50Z"
}
{
  "name": "configlet_4.0.0-beta.7_linux_x86-64.tar.gz",
  "updatedAt": "2022-11-05T15:34:31Z"
}
{
  "name": "configlet_4.0.0-beta.7_macos_x86-64.tar.gz",
  "updatedAt": "2022-11-05T15:34:31Z"
}
{
  "name": "configlet_4.0.0-beta.7_windows_x86-64.zip",
  "updatedAt": "2022-11-05T15:34:31Z"
}
```

PR #715, opened 2022-12-09
fmt, lint: support `representer` key in exercise `config.json`

ErikSchierboom:
See https://github.com/exercism/docs/commit/24881b8ed14b9ad03aa0f0a1d36e1309b4b5ff5a

@ee7 I've not bothered with individual commits, as I was unsure how to structure it. I'm happy for you to tweak it as desired.

### DIFF
--- a/src/lint/concept_exercises.nim
+++ b/src/lint/concept_exercises.nim
@@ -3,10 +3,10 @@ import ".."/helpers
 import "."/validators
 
 proc hasValidRepresenter(data: JsonNode; path: Path): bool =
-  const s = "representer"
-  if data.hasKey(s):
-    if hasObject(data, s, path):
-      result = hasInteger(data[s], "version", path, s, isRequired = true,
+  const k = "representer"
+  if data.hasKey(k):
+    if hasObject(data, k, path):
+      result = hasInteger(data[k], "version", path, k, isRequired = true,
                           allowed = 1..int.high)
 
 proc hasValidFiles(data: JsonNode; path, exerciseDir: Path): bool =
--- a/src/lint/concept_exercises.nim
+++ b/src/lint/concept_exercises.nim
@@ -7,7 +7,7 @@ proc hasValidRepresenter(data: JsonNode; path: Path): bool =
   if data.hasKey(k):
     if hasObject(data, k, path):
       result = hasInteger(data[k], "version", path, k, isRequired = true,
-                          allowed = 1..int.high)
+                          allowed = 1..1000)
   else:
     result = true
 
--- a/src/lint/concept_exercises.nim
+++ b/src/lint/concept_exercises.nim
@@ -8,6 +8,8 @@ proc hasValidRepresenter(data: JsonNode; path: Path): bool =
     if hasObject(data, k, path):
       result = hasInteger(data[k], "version", path, k, isRequired = true,
                           allowed = 1..int.high)
+  else:
+    result = true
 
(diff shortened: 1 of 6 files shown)


ErikSchierboom:
Side note: the trailing whitespace error messages are slightly unhelpful in that they don't list the actual line numbers:

```shell
There is trailing whitespace on the below lines:
src/types_exercise_config.nim:    
src/types_exercise_config.nim:    # `exercism/problem-specifications` repo. 
```

ee7 (commented):
Don't merge yet - currently, this makes `configlet lint` exit with an exit code of 1 on every track (or nearly every track). I think because the representer key needs to be optional.

I'll look at it.

Edit: I pushed https://github.com/exercism/configlet/pull/715/commits/f63cfaf5cd8ec55e7b4afb09ae30c74df6c0be58

src/lint/concept_exercises.nim
```diff
@@ -2,6 +2,15 @@ import std/[json, os]
 import ".."/helpers
 import "."/validators
 
+proc hasValidRepresenter(data: JsonNode; path: Path): bool =
+  const k = "representer"
+  if data.hasKey(k):
+    if hasObject(data, k, path):
+      result = hasInteger(data[k], "version", path, k, isRequired = true,
+                          allowed = 1..int.high)
```
ee7: The error message is a bit funny:

> The value of `representer.version` is 0, but it must be between 1 and 9223372036854775807 (inclusive):
> /path/to/track/exercises/practice/foo/.meta/config.json

Lots of runway for bumping the representer version.

Should we decrease the upper bound, and if so, specify it exactly?

For example, should we write `allowed = 1..2`, and increase the upper bound when we increase the maximum representer version in the spec?

If the maximum version in the spec is `n`, I think we want an error for an exercise config file that uses `n+1`. That shouldn't automatically imply the latest valid representer version, because then there's a silent and possibly breaking change when the maximum representer version is increased?

ErikSchierboom: Well, unlike the test runner, this version number is not something we enforce, but something that maintainers choose for their representers. The C# representer could be at version 5, but the Nim representer could be at version 99. Maybe we can just check if the number is > 0? I'd prefer to to have people use negative versions.

ee7: Ah.

I agree on disallowing negative version numbers. But I think I'm against `"version": 0` too - I was just checking that it was handled. I think it should start at `1`, for consistency with the [test runner version](https://github.com/exercism/docs/blob/ff94eefb7bad9c6c7e3596e1070bb3f2053720c6/building/tooling/test-runners/interface.md#version) that you mentioned.

How about something like `allowed = 1..100`? Or `allowed = 1..1000`?

ee7: I don't expect it to matter, but that would also stop someone from writing a version number like `20221214`. Good idea?

ee7: Also, running `configlet fmt` on a file that contains `"representer": {}` will produce

```json
  "representer": {
    "version": 0
  },
```

So it's probably a good thing for `configlet lint` to error for a version value of 0.

ErikSchierboom: `1..1000` is wide enough I think

ee7: OK, thanks.

I pushed https://github.com/exercism/configlet/pull/715/commits/74dd03c69f88266addbfd3626b71963c8ca1c7f8

ee7 (approved):

PR #761, opened 2023-06-14
create: add command for creating articles and approaches

ErikSchierboom:
- create: add command
- cli: add --exercise option for create command
- cli: add --approach <slug> option for create command
- cli: pass approach and exercise values
- cli: write basic approach
- Write files
- Fix failing import
- Format
- Add support for creating articles

PR #770, opened 2023-07-24
create, fmt, sync: move formatting code into fmt directory

ErikSchierboom:
@ee7 I found it a bit hard to work with the formatting code which seems to be defined all over the place. This PR is an initial attempt to move as much of the formatting code into the `src/fmt` dir.


ee7:
Resolved merge conflicts.

But the below tests failed in https://github.com/exercism/configlet/pull/770/commits/c5f58317dd6878ebb49b1a28833ba8329d1fac0b, and still fail in the current state:

- test_binary.nim
- test_fmt.nim
- test_sync.nim

ee7:
In `test_fmt`:

```nim
      test "with every Elixir Concept Exercise":
        for exerciseDir in getSortedSubdirs(conceptExercisesDir.Path):
          let exerciseConfigPath = joinPath(exerciseDir.string, ".meta", "config.json")
          let exerciseConfig = parseFile(exerciseConfigPath, ConceptExerciseConfig)
```

Running that test produces:

```text
JSON parsing error:
/foo/exercism-configlet/tests/.test_elixir_track_repo/exercises/concept/basketball-website/.meta/config.json(6, 19): Expected { but got [ instead.
```

It looks like it expects contributors to be a JSON object, not a JSON array.

```json
{
  "blurb": "Learn about Access Behaviour by helping extract deeply nested data for the basketball team's website.",
  "authors": [
    "neenjaw"
  ],
  "contributors": [
    "angelikatyborska",
    "NobbZ"
  ],
  "files": {
    "solution": [
      "lib/basketball_website.ex"
    ],
    "test": [
      "test/basketball_website_test.exs"
    ],
    "exemplar": [
      ".meta/exemplar.ex"
    ]
  },
  "language_versions": ">=1.10"
}
```

I didn't immediately see why, because `ConceptExerciseConfig` is defined as:

```nim
  ConceptExerciseConfig* = object
    originalKeyOrder*: seq[ExerciseConfigKey]
    authors*: seq[string]
    contributors*: Option[seq[string]]
```

and this PR doesn't touch that definition.

ee7:
With https://github.com/exercism/configlet/pull/770/commits/f924547f837b55d4882ac313804e220027366cdc, running the below file does not produce an error.

```nim
import std/os
import pkg/jsony
import exec, helpers, types_exercise_config

proc main =
  const testsDir = currentSourcePath().parentDir()
  const trackDir = testsDir / ".test_elixir_track_repo"

  # Setup: clone a track repo, and checkout a known state
  setupExercismRepo("elixir", trackDir,
                    "8447eaeb5ae8bdd0ae94383e6ec5bcfa21a7f993") # 2021-10-28

  const conceptExercisesDir = joinPath(trackDir, "exercises", "concept")

  let exerciseConfigPath = joinPath(conceptExercisesDir, "basketball-website",
                                    ".meta", "config.json")
  let exerciseConfig = parseFile(exerciseConfigPath, ConceptExerciseConfig)
  echo exerciseConfig.contributors

main()
```

So the problem lies in the difference between the above file and `test_fmt.nim`.

But the above code does error if we add `fmt/exercises` to the imports, or even `from fmt/exercises import nil` (!).

The error still occurs when we make `fmt/exercises.nim` contain exactly the below (note that the proc isn't even exported!)

```Nim
import ".."/types_exercise_config

proc formatExerciseConfigFile(exerciseKind: ExerciseKind;
                              configPath: string): string =
  ## Parses the `.meta/config.json` file at `configPath` and returns it in the
  ## canonical form.
  let exerciseConfig = ExerciseConfig.init(exerciseKind, configPath)
```

But the error doesn't occur if we don't import `fmt/exercises`, and copy the `formatExerciseConfigFile` definition into the top code block in this comment.

ee7:
Pushed https://github.com/exercism/configlet/pull/770/commits/a30acec1a18e62975fdfa4338b00deade219dd4e, which resolves the errors.

Here's a smaller reduction of the problem:

```nim
import std/options
from helpers import parseFile

type
  MyObject = object
    foo: Option[int]

proc main =
  const exerciseConfigPath = "/tmp/config.json"
  writeFile(exerciseConfigPath, """{"foo": 1}""")
  let exerciseConfig = parseFile(exerciseConfigPath, MyObject)
  echo exerciseConfig.foo

main()
```

this produces

```text
JSON parsing error:
/tmp/config.json(1, 9): Expected { but got 1 instead.
```

but if we add `import pkg/jsony`, we get:

```text
some(1)
```

ErikSchierboom:
@ee7 Is it due to templating that we don't get this as a compile error?

ErikSchierboom:
And thanks for figuring it out!

ee7:
> @ee7 Is it due to templating that we don't get this as a compile error?

Not exactly. I think it's eventually due to some compile-time logic in jsony (in particular, jsony's implementation of hooks), and how that interacts with Nim's generics.

Instead of `import pkg/jsony`, it is also sufficient to write `from pkg/jsony import parseHook`.

From the [jsony README](https://github.com/treeform/jsony/blob/4fa3a9b52649e31783ab2165b5476d5737c8842c/README.md#has-hooks):

> Note: that hooks need to be exported to where you are parsing the json so that the parsing system can pick them up.

And from [treeform/jsony#8](https://www.github.com/treeform/jsony/issues/8):

> it seems like due to Nim visibility constraint with generics `parseHook` must be exported or Nim won't use the proper one.

ee7:
But yes, particularly nasty. It wasn't obvious to me even though I fixed essentially the same bug 6 weeks ago in https://github.com/exercism/configlet/pull/761.

I think the long-term fix is to always export jsony in files that export routines that use jsony. Tracking in https://github.com/exercism/configlet/issues/817.

ee7:
See the better reduction below, which simulates the missing `parseHook`s by only importing `fromJson`.

```nim
import std/options
from jsony import fromJson

type
  MyObject = object
    foo: Option[int]

proc main =
  let myObject = fromJson("""{"foo": 1}""", MyObject)
  echo myObject.foo

main()
```

Output:

```text
/tmp/foo.nim(12) foo
/tmp/foo.nim(9) main
/home/foo/.nimble/pkgs2/jsony-1.1.5-6aeb83e7481ca8686396a568096054bc668294df/jsony.nim(590) fromJson
/home/foo/.nimble/pkgs2/jsony-1.1.5-6aeb83e7481ca8686396a568096054bc668294df/jsony.nim(468) parseHook
/home/foo/.nimble/pkgs2/jsony-1.1.5-6aeb83e7481ca8686396a568096054bc668294df/jsony.nim(377) parseObjectInner
/home/foo/.nimble/pkgs2/jsony-1.1.5-6aeb83e7481ca8686396a568096054bc668294df/jsony.nim(435) parseHook
/home/foo/.nimble/pkgs2/jsony-1.1.5-6aeb83e7481ca8686396a568096054bc668294df/jsony.nim(49) eatChar
/home/foo/.nimble/pkgs2/jsony-1.1.5-6aeb83e7481ca8686396a568096054bc668294df/jsony.nim(30) error
Error: unhandled exception: Expected { but got 1 instead. At offset: 8 [JsonError]
```

The error is resolved if we either:

- change `foo` from an `Option[int]` to `int`
- or change `from jsony import fromJson` to `from jsony import fromJson, parseHook` 

So it looks like if the `parseHook` for an `Option` isn't available where the JSON is parsed (which is more complicated when the proc that calls `fromJson` is generic, and in a separate file), jsony ends up expecting `foo` to be a JSON object.

I'll think about whether I should create an issue in the jsony repo for this.

ee7 (approved):
Our lack of integration tests for `fmt` are painful here. But I've checked that this PR does not change the output of `configlet fmt` or `configlet sync` on any track.

Moving formatting stuff to `fmt` sounds good to me. The formatting logic was only distributed as it was because `sync` was implemented before `fmt`.

I don't know how to check the exact factoring without starting from the same point and trying to do the same thing myself and checking the diff with this PR. But it looks like an improvement to me.

ErikSchierboom:
You happy to merge this?

ee7:
Yes, I'm merging now.

Issue #478, opened 2021-12-15
fmt: operate on every kind of JSON file on a track


ee7:
`configlet fmt` currently only operates on Concept Exercise and Practice Exercise `.meta/config.json` files.

It should also operate on the other JSON files:
- [ ] The track `config.json` file
- [ ] Concept `.meta/config.json` files
- [ ] Concept `links.json` files

ee7:
When `fmt` learns to format concept `.meta/config.json` files, I think we'll need to decide between:
- Adding a `--concept` option
- Renaming the `fmt` command's `--exercise` option to `--slug`, which can then be used for formatting an exercise or a concept (although this is probably bad if a concept can have the same name as an exercise).

We can also consider wheter `configlet fmt` should eventually grow some scoping options like `sync`, e.g.:
- `--track`
- `--exercises`
- `--concepts`

With the default of operating on all of them, as sync does.

ErikSchierboom:
I'll take a look at this

PR #772, opened 2023-07-31
fmt: support formatting track config

### LINKED ISSUE #479: tests: add "end-to-end" tests for `fmt`
The initial implementation in 4a88c133fe98 added only the tests in [`tests/test_sync.nim`](https://github.com/exercism/configlet/blob/4a88c133fe98709db4b0112d1f3d2cf96c1fd80e/tests/test_sync.nim).

ErikSchierboom:
- fmt: move all approach and article formatting code to `fmt` dir
- Extract exercises logic
- Fix newlines
- Fix tests import
- fmt(refactor): create separate DocumentKind values for concept and practice exercises
- fmt: support formatting track config

Refs: https://github.com/exercism/configlet/issues/478
Closes: https://github.com/exercism/configlet/issues/479


ee7:
CI failures look similar to https://github.com/exercism/configlet/pull/770#issuecomment-1719284811.

src/types_track_config.nim
```diff
@@ -17,14 +41,20 @@ type
...
     status*: Status
 
   PracticeExercise* = object
     slug*: Slug
+    name*: string
+    uuid*: string
     practices*: HashSet[string]
     prerequisites*: HashSet[string]
+    difficulty*: int
+    topics*: HashSet[string]
```
ee7: I think at least some of the CI failures are because the Nim track has `null` as a `topics` value.

https://github.com/search?q=org%3Aexercism+%22topics%22%3A+null+path%3Aconfig.json&type=code

Needs to be an Option, I think. Or we remove all the null.

tests/test_binary.nim
```diff
@@ -1008,6 +1008,99 @@ proc testsForCompletion(binaryPath: string) =
...
+    fmtUpdate = &"{fmtBase} --update"
+    header = fmt"""
+      Found 0 Concept Exercises and 68 Practice Exercises in {trackDir}/config.json
+      Looking for exercises that lack a formatted '.meta/config.json', '.approaches/config.json'
+      or '.articles/config.json' file...
+      The below paths are relative to '{trackDir}'
+    """.unindent().strip(trailing = true)
+
+  suite "fmt, when the track `config.json` file is not formatted (prints the expected output, and exits with 1)":
+    test "-t foo":
```
ee7:
```suggestion
  suite "fmt, when the track `config.json` file is not formatted":
    test "prints the expected output, and exits with 1":
```

Or whatever.

tests/test_binary.nim
```diff
@@ -1008,6 +1008,99 @@ proc testsForCompletion(binaryPath: string) =
...
+        Not formatted: exercises/practice/space-age/.meta/config.json
+        Not formatted: exercises/practice/spiral-matrix/.meta/config.json
+        Not formatted: exercises/practice/sublist/.meta/config.json
+        Not formatted: exercises/practice/sum-of-multiples/.meta/config.json
+        Not formatted: exercises/practice/triangle/.meta/config.json
+        Not formatted: exercises/practice/twelve-days/.meta/config.json
+        Not formatted: exercises/practice/two-fer/.meta/config.json
+        Not formatted: exercises/practice/word-count/.meta/config.json
+        Not formatted: exercises/practice/yacht/.meta/config.json
+      """.unindent()
```
ee7: Looks like this test fails on Windows. Maybe from line endings. Maybe here.

ee7: Edit: oh, it's because of the dir separators.

tests/test_binary.nim
```diff
@@ -367,7 +367,8 @@ proc testsForSync(binaryPath: static string) =
       +    "example": []
       +  "blurb": "Diffie-Hellman key exchange.",
     """.unindent()
-    let configPath = joinPath("exercises", "practice", "diffie-hellman", ".meta", "config.json")
+    let configPath = joinPath("exercises", "practice", "diffie-hellman",
+        ".meta", "config.json")
```
ErikSchierboom: This was auto-formatted by my Nim plugin in VS Code. I can revert, but it will do this every single time

tests/test_binary.nim
```diff
@@ -1008,6 +1012,285 @@ proc testsForCompletion(binaryPath: string) =
...
+  const formattedTrackDir = testsDir / ".test_nim_track_repo"
+  const unformattedTrackDir = testsDir / ".test_elixir_track_repo"
+
+  # Setup: clone track repo, and checkout a known state (formatted)
+  setupExercismRepo("nim", formattedTrackDir,
+                    "ea91acb3edb6c7bc05dd3b050c0a566be6c3329e") # 2022-01-22
+
+  # Setup: clone track repo, and checkout a known state (unformatted)
+  setupExercismRepo("elixir", unformattedTrackDir,
+                    "07448c4f870c15f8191196a2b01e8bf09708b8ce") # 2022-01-11
```
ErikSchierboom, line 1021: I'm using both the Elixir and Nim repo here, with Nim being the one where all the config files (except the track `config.json` file) are formatted, and Elixir having none of the formatted.

tests/test_binary.nim
ee7: Are you using the nimsaem extension? That's the preferred and more maintained one.

ErikSchierboom: Let me double check!

ErikSchierboom: I am using that extension

ee7: Hmm. Well, I think you should be able to disable the formatting-on-save. Alternatively, you could add something like `--maxLineLen:100` as an argument to nimpretty.

tests/test_binary.nim
```diff
@@ -1008,6 +1012,286 @@ proc testsForCompletion(binaryPath: string) =
...
+
+  suite "fmt, when the track `config.json` file is not formatted":
+    test "prints the expected output, and exits with 1":
+      const expectedOutput = fmt"""
+        {unformattedHeader}
+        Looking for exercises that lack a formatted '.meta/config.json', '.approaches/config.json'
+        or '.articles/config.json' file...
+        The below paths are relative to '{unformattedTrackDir}'
+        Not formatted: config.json
+        Not formatted: exercises/concept/basketball-website/.meta/config.json
```
ee7: Tests still fail on Windows because all these have `\` instead of `/`, like:

```
Not formatted: exercises\concept\basketball-website\.meta\config.json
```

tests/test_binary.nim
```diff
@@ -1008,6 +1012,286 @@ proc testsForCompletion(binaryPath: string) =
...
+      execAndCheck(1, cmd, expectedOutput)
+
+  suite "fmt, for an exercise that is not formatted (prints the expected output, and exits with 1)":
+    test "-e bob":
+      const expectedOutput = fmt"""
+        {unformattedHeader}
+        Looking for exercises that lack a formatted '.meta/config.json', '.approaches/config.json'
+        or '.articles/config.json' file...
+        The below paths are relative to '{unformattedTrackDir}'
+        Not formatted: exercises/practice/bob/.meta/config.json
```
ee7: Same.

tests/test_binary.nim
ee7: Re-stated in https://github.com/exercism/configlet/pull/772#discussion_r1334293663

ee7 (commented):
Remaining windows failure:

```
[Suite] fmt, with --update, without --yes, for an exercise that is formatted (no diff, and exits with 0)
D:\a\configlet\configlet\tests\test_binary.nim(1246, 21): Check failed: exitCode == 0
exitCode was 1
[FAILED] -e bob
```

ErikSchierboom:
> Remaining windows failure:

Weird. Can we output more information?

ee7:
> Weird. Can we output more information?

Yes. Currently it's discarding the command output:

```nim
let exitCode = execCmdEx(&"{fmtUpdateFormatted} --yes -e bob")[1]
```

If you want to debug, you can write something ike:

```nim
let (outp, exitCode) = execCmdEx(&"{fmtUpdateFormatted} --yes -e bob")
echo outp
```

ee7:
With https://github.com/exercism/configlet/pull/772/commits/75c48c9059719e748677b6146926684b6bb01d25 the debug output (https://github.com/exercism/configlet/actions/runs/6274523193/job/17040081783) is:

```text
Found 0 Concept Exercises and 68 Practice Exercises in D:\a\configlet\configlet\tests\.test_nim_track_repo\config.json
Looking for exercises that lack a formatted '.meta/config.json', '.approaches/config.json'
or '.articles/config.json' file...
The below paths are relative to 'D:\a\configlet\configlet\tests\.test_nim_track_repo'
Not formatted: exercises\practice\bob\.meta\config.json
Formatted 1 file
```

ErikSchierboom:
Ah, this is what it looks like on my machine:

```
Found 0 Concept Exercises and 68 Practice Exercises in /home/erik/exercism/configlet/tests/.test_nim_track_repo/config.json
Looking for exercises that lack a formatted '.meta/config.json', '.approaches/config.json'
or '.articles/config.json' file...
The `bob` exercise config is already formatted!
```

It _must_ be newlines, right?

ee7:
> It _must_ be newlines, right?

I guess so.

ErikSchierboom:
All green!

tests/test_binary.nim
```diff
@@ -1008,6 +1008,287 @@ proc testsForCompletion(binaryPath: string) =
...
+        --- exercises/practice/bob/.meta/config.json
+        +++ exercises/practice/bob/.meta/config.json
+        -  "blurb": "Bob is a lackadaisical teenager. In conversation, his responses are very limited.",
+        +  "blurb": "Bob is a lackadaisical teenager. In conversation, his responses are very limited.",
+      """.unindent().replace("\p", "\n")
+      let configPath = "exercises" / "practice" / "bob" / ".meta" / "config.json"
+      let trackDir = unformattedTrackDir
+      testDiffThenRestore(trackDir, expectedDiff, configPath)
+
+  suite "fmt, with --update, without --yes, for an exercise that is formatted (no diff, and exits with 0)":
```
ee7:
```suggestion
  suite "fmt, with --update --yes, for an exercise that is formatted (no diff, and exits with 0)":
```

tests/test_binary.nim
```diff
@@ -1008,6 +1008,287 @@ proc testsForCompletion(binaryPath: string) =
...
+      """.unindent().replace("\p", "\n")
+      execAndCheck(1, &"{fmtBaseUnformatted} -e foo", expectedOutput)
+
+  suite "fmt, with --update, without --yes, for an exercise that is not formatted (no diff, and exits with 1)":
+    test "-e bob":
+      let exitCode = execCmdEx(&"{fmtUpdateUnformatted} -e bob")[1]
+      check exitCode == 1
+      checkNoDiff(unformattedTrackDir)
+
+  suite "fmt, with --update, for an exercise that is not formatted (no diff, and exits with 0)":
```
ee7:
```suggestion
  suite "fmt, with --update, for an exercise that is not formatted (diff, and exits with 0)":
```

tests/test_binary.nim
```diff
@@ -1008,6 +1008,287 @@ proc testsForCompletion(binaryPath: string) =
...
+        -    "typing/static",
+        -    "typing/strong",
+        -    "execution_mode/compiled",
+        -    "platform/windows",
+        -    "platform/mac",
+        +    "platform/mac",
+        +    "platform/windows",
+        +    "typing/static",
+        +    "typing/strong",
+      """.unindent().replace("\p", "\n")
```
ee7, line 1285: For later: maybe we can make `testDiffThenRestore` handle the platform-specific newlines.

tests/test_binary.nim
```diff
@@ -1008,6 +1008,287 @@ proc testsForCompletion(binaryPath: string) =
...
+        +    "execution_mode/compiled",
+        -    "typing/static",
+        -    "typing/strong",
+        -    "execution_mode/compiled",
+        -    "platform/windows",
+        -    "platform/mac",
+        +    "platform/mac",
+        +    "platform/windows",
+        +    "typing/static",
+        +    "typing/strong",
```
ee7, line 1284: For later: improve this diff and others. The test could be stronger than it currently is, because the current diff is removing lines of context.

ee7 (commented):
I've run `configlet fmt` on every track. The only JSON parsing error is:

```text
JSON parsing error:
./config.json(20, 26): Expected } but got . instead.
```

for the (deactivated) clojurescript track [here](https://github.com/exercism/clojurescript/blob/3bbaf7e8cbb08ba000ccd351b93b5dd25c13f4f7/config.json#L20).

ee7 (commented):
`fmt -uy` makes this change to a track `config.json`: 

```diff
   "exercises": {
-    "concept": [],
     "practice": [
```

and:

```diff
-  "concepts": [],
```

I'm fine with that, because e.g. `configlet fmt -uy` already removes an empty `contributors`. OK with you? There isn't test coverage for this case, because the Nim and Elixir tracks lack such empty arrays.

ee7:
And [this](https://github.com/exercism/elixir/blob/2e344ce5196eb7991fe2158da25170a61ef1130e/config.json#L693) key/value pair doesn't do anything, right?

Edit: see https://github.com/exercism/elixir/pull/827

> - For the automated reordering, I'm using `jq` because Elixir JSON parsers all parse objects into maps, losing key order information, and thus producing very bad git diffs when encoding the file again.
> - To ensure that hello-world is always first, I added a custom `order` field to that exercise only. It will be used before `slug`

We don't want to support custom key/value pairs in either a track-level or exercise-level `config.json` file, right?

ee7 (approved):
LGTM - as far as I can tell, it produces correct results in my testing. Aside: one of the main things that stood out was that the Rust track had *concept* exercises [with `difficulty`](https://github.com/exercism/rust/blob/5ee312deca68291989b8ef0ee10ee9abc311e3ba/config.json#L42).

IIRC we should be able to refactor the serialization code later to use jsony's `dumpHook`.

Happy for me to merge?

ee7:
And I think we can say that this closes https://github.com/exercism/configlet/issues/479. We can improve the coverage and make things stricter later.

ErikSchierboom:
> I'm fine with that, because e.g. configlet fmt -uy already removes an empty contributors. OK with you? There isn't test coverage for this case, because the Nim and Elixir tracks lack such empty arrays.

Yep, that's fine with me!

> And [this](https://github.com/exercism/elixir/blob/2e344ce5196eb7991fe2158da25170a61ef1130e/config.json#L693) key/value pair doesn't do anything, right?
> Edit: see https://github.com/exercism/elixir/pull/827
> 
> For the automated reordering, I'm using jq because Elixir JSON parsers all parse objects into maps, losing key order information, and thus producing very bad git diffs when encoding the file again.
To ensure that hello-world is always first, I added a custom order field to that exercise only. It will be used before slug
We don't want to support custom key/value pairs in either a track-level or exercise-level config.json file, right?

> For the automated reordering, I'm using jq because Elixir JSON parsers all parse objects into maps, losing key order information, and thus producing very bad git diffs when encoding the file again.

I don't entirely understand this, because the exercises are an array in the source. I think we'll have to work with the Elixir people to figure out an alternative approach.

> LGTM - as far as I can tell, it produces correct results in my testing. Aside: one of the main things that stood out was that the Rust track had concept exercises [with difficulty](https://github.com/exercism/rust/blob/5ee312deca68291989b8ef0ee10ee9abc311e3ba/config.json#L42).

Yeah, that's just wrong (although nothing fails) :)

PR #777, opened 2023-08-07
nimble, uuid: generate UUIDs via `std/sysrand`, not `pragmagic/uuids`

### LINKED ISSUE #424: uuid: use best practices for UUID generation

ee7:
Replace two Nimble dependencies with our own code.

Before this commit, configlet would generate a UUID with the Nimble packages [pragmagic/uuids][1] and [pragmagic/isaac][2] (itself a port of a [C implementation][3]).

However, the `uuids` package doesn't currently work on Windows with Nim 2.0 (due to Nim [removing][4] some functions for a deprecated Win32 API). I've [created a PR for the package][5], but:

- That PR may not be sufficient.
- I want to move away from these dependencies anyway (see #424).
- Our tests for UUID generation are good enough that we can confidently change the implementation.

`std/sysrand` has been around for 2.5 years, but still [warns][6] that it hasn't been formally audited and doesn't guarantee safety for cryptographic purposes. But it's probably good enough for our purposes: there's more use, review, and maintenance of std/sysrand than pragmagic/isaac anyway (whose most recent commit was in 2017, and ISAAC is questionable as a CSPRNG).

In the tests, add a test for distinctness, and test more UUIDs. The `configlet uuid` command sets a limit of 1000 UUIDs, so let's assert that nothing strange happens above that number. Pick the number such that the genUuid tests take on the order of 100 ms.

We can't make this a much higher number because we compile the tests in debug mode, and `configlet uuid` is intended for generating a small number of UUIDs. But I tested locally that the new implementation could produce 500M distinct valid version 4 UUIDs.

As a future optimization we could either read more than 16 bytes each time from the system CSPRNG, or once again use the system CSPRNG to seed a different PRNG.

Other benefits of this change:

- It makes configlet more portable. Previously, `configlet uuid` would [only work on Windows and platforms with `/dev/urandom`][7].
- We remove a gotcha from the nimble file version pinning.
- We can rename from `genUUID` to `genUuid` fit our style elsewhere.

Closes: #424

[1]: https://github.com/pragmagic/uuids/blob/8cb8720b567c/uuids.nim
[2]: https://github.com/pragmagic/isaac/blob/45a5cbbd54ff/src/isaac.nim
[3]: https://burtleburtle.net/bob/c/readable.c
[4]: https://github.com/nim-lang/Nim/commit/612abda4f40b
[5]: https://www.github.com/pragmagic/uuids/pull/9
[6]: https://github.com/nim-lang/Nim/blob/v2.0.0/lib/std/sysrand.nim#L10-L17
[7]: https://github.com/pragmagic/uuids/blob/8cb8720b567c/uuids/urandom.nim#L41-L62

---

See also the individual commit messages.

### DIFF
--- a/src/uuid/uuid.nim
+++ b/src/uuid/uuid.nim
@@ -1,17 +1,28 @@
-import std/[strformat, sysrand]
+import std/[strformat, random, sysrand]
 import ".."/logger
 
 type
   Uuid* = array[16, byte]
 
-proc genUuid*: Uuid {.noinit.} =
-  ## Returns a version 4 UUID, using the system CSPRNG as the source of randomness.
-  if urandom(result):
-    result[6] = (result[6] and 0x0f) or 0x40 # Set version to 4
-    result[8] = (result[8] and 0x3f) or 0x80 # Set variant to 1
-  else:
+var r = block:
+  var seed: array[8, byte]
+  if not urandom(seed):
     stderr.writeLine "uuid: error: failed to generate UUID"
     quit 1
+  initRand(cast[int64](seed))
+
+proc genUuid*: Uuid {.noinit.} =
+  ## Returns a version 4 UUID, using the system CSPRNG as the source of randomness.
+  var a = rand(r, uint32.high) # Can't use uint64.high
+  copyMem(result[0].addr, a.addr, 4)
+  a = rand(r, uint32.high)
+  copyMem(result[4].addr, a.addr, 4)
+  a = rand(r, uint32.high)
+  copyMem(result[8].addr, a.addr, 4)
+  a = rand(r, uint32.high)
+  copyMem(result[12].addr, a.addr, 4)
+  result[6] = (result[6] and 0x0f) or 0x40 # Set version to 4
+  result[8] = (result[8] and 0x3f) or 0x80 # Set variant to 1
 
 func `$`*(u: Uuid): string =
   ## Returns the canonical string representation for the given UUID `u`.
--- a/src/uuid/uuid.nim
+++ b/src/uuid/uuid.nim
@@ -1,28 +1,17 @@
-import std/[strformat, random, sysrand]
+import std/[strformat, sysrand]
 import ".."/logger
 
 type
   Uuid* = array[16, byte]
 
-var r = block:
-  var seed: array[8, byte]
-  if not urandom(seed):
-    stderr.writeLine "uuid: error: failed to generate UUID"
-    quit 1
-  initRand(cast[int64](seed))
-
 proc genUuid*: Uuid {.noinit.} =
   ## Returns a version 4 UUID, using the system CSPRNG as the source of randomness.
-  var a = rand(r, uint32.high) # Can't use uint64.high
-  copyMem(result[0].addr, a.addr, 4)
-  a = rand(r, uint32.high)
-  copyMem(result[4].addr, a.addr, 4)
-  a = rand(r, uint32.high)
-  copyMem(result[8].addr, a.addr, 4)
-  a = rand(r, uint32.high)
-  copyMem(result[12].addr, a.addr, 4)
-  result[6] = (result[6] and 0x0f) or 0x40 # Set version to 4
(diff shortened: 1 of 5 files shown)


src/uuid/uuid.nim
```diff
@@ -1,7 +1,30 @@
...
+  ## Returns the canonical string representation for the given UUID `u`.
+  result = newString(36)
+  result[8] = '-'
+  result[13] = '-'
+  result[18] = '-'
+  result[23] = '-'
+  for i, j in [0, 2, 4, 6, 9, 11, 14, 16, 19, 21, 24, 26, 28, 30, 32, 34]:
+    const hex = "0123456789abcdef"
+    result[j + 0] = hex[u[i] shr 4]
+    result[j + 1] = hex[u[i] and 0x0f]
```
ee7, line 26: Compare with [public domain Go implementation in skeeto/uuid](https://github.com/skeeto/uuid/blob/5a4e5674e3920c4a4d81b9129fee38fb623ac7e4/uuid.go#L60-L71).

(I like that guy's [blog](https://nullprogram.com/index/)).

There's also a [Zig port](https://github.com/dmgk/zig-uuid/blob/9895f72f67d463b6bd616cc63dcfe70d12dc53b9/uuid.zig) of that code.

Alternatively:

- Rust: https://docs.rs/uuid/latest/src/uuid/lib.rs.html
- Python: https://github.com/python/cpython/blob/3.12/Lib/uuid.py

ErikSchierboom (approved):
Lovely!

PR #557, opened 2022-04-15
.github, README: sign release archives with minisign

ee7:
Since commit f7c12253077d, each configlet release has a checksums file. The assets for the [most recent release](https://github.com/exercism/configlet/releases/tag/4.0.0-beta.12) were:

```text
configlet_4.0.0-beta.12_checksums_sha256.txt
configlet_4.0.0-beta.12_linux_x86-64.tar.gz
configlet_4.0.0-beta.12_macos_x86-64.tar.gz
configlet_4.0.0-beta.12_windows_x86-64.zip
```

But confirming a checksum match only checks for accidental corruption. Add signing with [minisign][1] to the release process, such that the next release will have a `.minisig` file for each archive:

```text
configlet_4.0.0-beta.13_checksums_sha256.txt
configlet_4.0.0-beta.13_linux_x86-64.tar.gz
configlet_4.0.0-beta.13_linux_x86-64.tar.gz.minisig
configlet_4.0.0-beta.13_macos_x86-64.tar.gz
configlet_4.0.0-beta.13_macos_x86-64.tar.gz.minisig
configlet_4.0.0-beta.13_windows_x86-64.zip
configlet_4.0.0-beta.13_windows_x86-64.zip.minisig
```

From the [minisign docs][2]:

> Minisign is a dead simple tool to sign files and verify signatures.
> It is portable, lightweight, and uses the highly secure Ed25519 public-key signature system.

For example, to verify the above Linux x86-64 release archive, run:

```shell
minisign -Vm configlet_4.0.0-beta.13_linux_x86-64.tar.gz -P RWQGj6DTXgYLhKvWJMGtbDUrZerawUcyWnti9MGuWMx7VDW9DqZn2tMZ
```

where the argument to -P is the configlet minisign public key. Alternatively, you can download the `configlet-minisign.pub` file and run:

```shell
minisign -Vm configlet_4.0.0-beta.13_linux_x86-64.tar.gz -p configlet-minisign.pub
```

Minisign is an implementation of OpenBSD's signify protocol. For more background, see posts on [signify][3] and [problems with PGP][4].

An alternative would be to [sign releases with an SSH key][5], but minisign is more focused, always uses Ed25519, and is simpler. Especially for verification.

Some other projects that sign releases with minisign or signify:

- [libsodium][6]
- [OpenBSD][7]
- [Void Linux][8]
- [Zig][9]
- [LibreSSL][10]

[...]

PR #784, opened 2023-08-09
README: add example minisign output, and clarify

ee7:
Part of this was already approved in #557, but I [removed it](https://github.com/exercism/configlet/pull/557/commits/7f0668f9410b46841dca64b8e9e92c6a94d84954) before merging because I didn't want to use a placeholder filename, and wanted this timestamp to match the one from the release.

The wording change from "contains" to "begins with" isn't important. But I guess the previous wording wasn't strictly true: a filename could contain "Signature and comment signature verified".

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,12 @@ minisign -Vm configlet_4.0.0-beta.13_linux_x86-64.tar.gz -P RWQGj6DTXgYLhKvWJMGt
 where the above argument to `-P` is the configlet public key.
 
 The above command has verified the release archive if (and only if) the command's output contains `Signature and comment signature verified`.
+For example:
+
+```text
+Signature and comment signature verified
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,9 +32,15 @@ Then run a `minisign` command in that directory:
 minisign -Vm configlet_4.0.0-beta.13_linux_x86-64.tar.gz -P RWQGj6DTXgYLhKvWJMGtbDUrZerawUcyWnti9MGuWMx7VDW9DqZn2tMZ
 ```
 
-where the above argument to `-P` is the configlet public key.
+where the above argument to `-P` is the [configlet public key](https://github.com/exercism/configlet/blob/009dc9df9d947e71ff039ac2af0f82315dbf9073/configlet-minisign.pub).
```
ee7, line 35: Would be better as a reference link, but we don't use them in this file currently.

ErikSchierboom (approved):

Issue #789, opened 2023-08-16
build: cross compilation script duplicates some logic


ee7:
Follow-up from https://github.com/exercism/configlet/pull/460#discussion_r1293937435

PR #792, opened 2023-08-16
.github: separate release creation from uploading assets

### LINKED ISSUE #551: build(publish-release): race between `release view` and `release create`
Follow-up from something I noticed in #550. There's a race condition here:

https://github.com/exercism/configlet/blob/4ddb23ddc59e48345c479ec8d43af05b4894cd42/.github/bin/publish-release#L5-L9

@ee7 wrote:
> Note that we still have another race condition: a separate job can
create a release in the time between `gh release view` indicating
"there is no such release" and `gh release create` running. We should
address that later, but it hasn't caused a problem yet - the Linux job
consistently finishes significantly faster than the others. And if we
encounter such a failure in the meantime, the workaround is simple: just
restart the failing job.

One possible way to resolve this is to ensure that `gh release create` can only run once:
1. Make each job use [`actions/upload-artifact`](https://github.com/actions/upload-artifact) only.
2. Add a dependent job that uses [`actions/download-artifact`](https://github.com/actions/download-artifact), and then runs `gh release create`.

ee7:
Before this commit, the `.github/bin/publish-release` script handled both:
- creating the draft release
- uploading assets to the release

with the intention that the first build job to finish would create the release with one asset, and each later job would upload an asset.

The logic was like:
```shell
if ! gh release view "${build_tag}"; then
  gh release create [...]
else
  gh release upload [...]
fi
```
But this had a race. For example, if two build jobs finished at nearly the same moment, they could each create a draft release. This race was almost never encountered in practice, and was easily resolved by deleting a draft release and restarting a build job, but it's worth resolving.

Add an initial build job to create a draft release that has no assets, and make the later build jobs only upload to that release.

This fixes the race, and makes it easier to share logic between the native build jobs and the cross-compiling build jobs.

Fixes: #551
Refs: #789

---

To-do:

- [x] Test build job
- [x] Follow-up PR: rename build scripts (do separately to ensure that git can track the rename)
- [x] Follow-up PR: make cross-compilation job run in parallel with native build jobs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,24 @@ on:
   workflow_dispatch:
 
 jobs:
+  create-empty-release:
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
+        with:
+          fetch-depth: 0 # Allows using `git log` to set initial release notes.
+
+      - name: Create empty release
+        shell: bash
+        run: ./.github/bin/publish-release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
   build:
+    needs: [create-empty-release]
     strategy:
       fail-fast: false
       matrix:
@@ -31,8 +48,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
-        with:
-          fetch-depth: 0 # Allows using `git log` to set initial release notes.
 
       - name: On Linux, install musl
         if: matrix.os == 'linux'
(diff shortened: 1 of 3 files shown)


ee7:
I manually triggered this PR's build workflow. It worked, and created [this draft release](https://github.com/exercism/configlet/releases/tag/untagged-8c1e65d0a6d4da66d4ab).

~CI indicates failure due to rate limiting in the macos fetch-configlet job.~ Edit: It worked after some restarts.

.github/workflows/build.yml
```diff
@@ -7,7 +7,24 @@ on:
...
+      contents: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
+        with:
+          fetch-depth: 0 # Allows using `git log` to set initial release notes.
+
+      - name: Create empty release
+        shell: bash
+        run: ./.github/bin/publish-release
```
ErikSchierboom, line 22: Minor semantic comment: is this publishing a release or creating one?

ErikSchierboom (approved):

.github/workflows/build.yml
ErikSchierboom: Nevermind, just saw: https://github.com/exercism/configlet/pull/794/files

Issue #764, opened 2023-07-04
aarch64 build target for release


neenjaw:
Hey @ee7 and @ErikSchierboom,

I switched platforms from ubuntu to macOS for my daily driver and noticed that configlet can't be fetched using the included repo script any more.  Reading the documentation, I didn't see any method described to build it from source myself as a very poor `nim` user.

Could you advise?

Thanks

ErikSchierboom:
This has been an open issue for a while IIRC. You should be able to use the x86 version by downloading it from an individual release. That said, this is not optimal and something we should fix.

neenjaw:
No rush on it, I've found a way to circumvent this need 😄 

https://github.com/exercism/php/blob/a88fadea963597af0c8980faef89cd0ad6ca01bb/.github/workflows/configlet-generate.yml

ee7:
I've been planning to add aarch64 (AKA arm64) assets for some time. But in the meantime, I thought that `fetch-configlet` already downloaded the x86_64 asset when running on aarch64 macOS.

It looks like that, here:

https://github.com/exercism/configlet/blob/18d9098d990151db2ab78ea4914328083b997ff5/scripts/fetch-configlet#L26-L30

@neenjaw can you confirm that https://github.com/exercism/configlet/blob/main/scripts/fetch-configlet doesn't work on your machine? And if it doesn't, what happens?

Maybe it was just an intermittent connectivity issue? Or you were using an old fetch-configlet script? There was a breaking change in the asset naming scheme last year (see https://github.com/exercism/configlet/commit/57e6c5fdb3f3e3939b9584adbe667c6e1890e59f), and 6 months after that I stopped supporting old fetch-configlet scripts.

neenjaw:
So using the script at the link you shared:

```
curl https://raw.githubusercontent.com/exercism/configlet/main/scripts/fetch-configlet | bash
```

It seems to download configlet as expected but I still get a `bash: line 86: ./bin/configlet: Bad CPU type in executable` error which crashes with error code 127.

I think the fetch configlet script is still downloading the inappropriate version. I added this line above line 30 in the `fetch-configlet` script:

```diff
+    *arm*)  arch='arm64'  ;;
     *).     arch='x86-64' ;;
```

Which appears to work as expected now to fetch a runnable configlet

ee7:
Hey @neenjaw, thanks for following up.

Indeed, we haven't yet updated the `fetch-configlet` script in either this repo or any track repo. Some rationale:

[...]

Issue #122, opened 2021-01-15
fetch-configlet scripts: Add ARM support


ee7:
Similar to issue https://github.com/exercism/canonical-data-syncer/issues/24, but for the fetch scripts (which already have 32-bit support).

The v2 `configlet` [released ARM binaries](https://github.com/exercism/configlet/releases/tag/v3.9.2), but the [fetch-configlet script there](https://github.com/exercism/configlet/blob/86bf086c6fbb0486a43a4f499baffb5a8f2a8278/scripts/fetch-configlet) never supported downloading them.

This will become a more important issue as more people start using the latest ARM-based Mac machines with the Apple M1 chip.

PR #805, opened 2023-08-22
fetch-configlet: support fetching arm64 assets

ee7:
The [most recent configlet release](https://github.com/exercism/configlet/releases/tag/4.0.0-beta.14) includes pre-built binaries for:

- [arm64 Linux][1]
- [arm64 macOS][2]
- [arm64 Windows][3]
- [riscv64 Linux][4]

Support arm64 in the bash fetch-configlet script.

Refs: #122
Follow-up from: https://github.com/exercism/configlet/issues/764#issuecomment-1685080166

[1]: https://github.com/exercism/configlet/commit/0e8d6659e43d
[2]: https://github.com/exercism/configlet/commit/f280445978c1
[3]: https://github.com/exercism/configlet/commit/3824299df1dd
[4]: https://github.com/exercism/configlet/commit/a962b181b54e

---

I'll change the logic in this script in a later PR. I think we shouldn't hardcode the available binaries in the fetch scripts. Instead, we should try to download the configlet release for the user's machine, and error if the asset does not exist. Then we can add assets to the configlet releases without updating the fetch scripts on every track. That'll also make the fetch script support riscv64 Linux.

I suggest delaying the Exercism-wide PRs until we've done that.

### DIFF
--- a/scripts/fetch-configlet
+++ b/scripts/fetch-configlet
@@ -24,10 +24,11 @@ get_download_url() {
   local latest='https://api.github.com/repos/exercism/configlet/releases/latest'
   local arch
   case "$(uname -m)" in
-    x86_64) arch='x86-64' ;;
-    *686*)  arch='i386'   ;;
-    *386*)  arch='i386'   ;;
-    *)      arch='x86-64' ;;
+    aarch64|arm64) arch='arm64'  ;;
+    x86_64)        arch='x86-64' ;;
+    *686*)         arch='i386'   ;;
+    *386*)         arch='i386'   ;;
+    *)             arch='x86-64' ;;
   esac
   local suffix="${os}_${arch}.${ext}"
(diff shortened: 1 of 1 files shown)


ee7:
[CI failure](https://github.com/ee7/exercism-configlet/actions/runs/5935872454/job/16095123411) should be resolved by https://github.com/exercism/configlet/pull/806.

ee7:
@ErikSchierboom I'll leave updating the PowerShell script to you. But let's do that in a separate PR.

Currently:

https://github.com/exercism/configlet/blob/dca4119e22667d562270f4ca4f7ab545ee79e302/scripts/fetch-configlet.ps1#L15-L16

I think it should also probably have the same logic change that I suggest in the top post. But it's less important.

I suspect that we don't have a huge number of non-Windows PowerShell users, but we could consider resolving https://github.com/exercism/configlet/issues/361 at the same time.

ErikSchierboom (approved):

PR #809, opened 2023-08-22
fetch-configlet: support more platforms

ee7:
Allow `fetch-configlet` to:

- download the existing riscv64 Linux asset
- support future assets for e.g. FreeBSD/NetBSD/OpenBSD
- support low-priority, but theoretically possible future assets for powerpc/powerpc64/powerpc64le/sparc64/etc

---

Working towards my suggestion in https://github.com/exercism/configlet/pull/805:

>I think we shouldn't hardcode the available binaries in the fetch scripts. Instead, we should try to download the configlet release for the user's machine, and error if the asset does not exist. Then we can add assets to the configlet releases without updating the fetch scripts on every track. That'll also make the fetch script support riscv64 Linux.

### DIFF
--- a/scripts/fetch-configlet
+++ b/scripts/fetch-configlet
@@ -22,13 +22,15 @@ get_download_url() {
   local os="$1"
   local ext="$2"
   local latest='https://api.github.com/repos/exercism/configlet/releases/latest'
+  local unamem
+  unamem="$(uname -m)"
   local arch
-  case "$(uname -m)" in
-    aarch64|arm64) arch='arm64'  ;;
-    x86_64)        arch='x86-64' ;;
-    *686*)         arch='i386'   ;;
-    *386*)         arch='i386'   ;;
-    *)             arch='x86-64' ;;
+  case "${unamem}" in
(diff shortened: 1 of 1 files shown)


scripts/fetch-configlet
```diff
@@ -22,12 +22,18 @@ get_download_url() {
...
+  unamem="$(uname -m)"
   local arch
-  case "$(uname -m)" in
-    x86_64) arch='x86-64' ;;
-    *686*)  arch='i386'   ;;
-    *386*)  arch='i386'   ;;
-    *)      arch='x86-64' ;;
+  case "${unamem}" in
+    aarch64|arm64) arch='arm64'       ;;
+    x86_64)        arch='x86-64'      ;;
```
ee7, line 30:
```suggestion
    aarch64|arm64) arch='arm64'       ;;
    arm*)          arch='arm'         ;;
    x86_64)        arch='x86-64'      ;;
```

Maybe.

ErikSchierboom (approved):

PR #822, opened 2023-10-05
lint, fmt: support `tags` for approaches and concepts

ErikSchierboom:
- Lint tags for approaches
- Lint tags for concepts
- Don't check tags for articles
- Try fmt tags

---

For background, see the [forum thread][1].

[1]: https://forum.exercism.org/t/proposal-adding-tag-detection-to-analyzers/6931/7

### DIFF
--- a/src/fmt/approaches.nim
+++ b/src/fmt/approaches.nim
@@ -1,4 +1,4 @@
-import std/[json, strformat]
+import std/[json, options, strformat]
 import ".."/[helpers, sync/sync_common, types_approaches_config]
 
 func approachesConfigKeyOrderForFmt(e: ApproachesConfig): seq[ApproachesConfigKey] =
@@ -23,6 +23,20 @@ func addApproachesIntroduction(result: var string;
   result.addNewlineAndIndent(indentLevel)
   result.add "},"
 
+func addApproachTags(result: var string; val: ApproachConfigTags; indentLevel = 2) =
+  result.addNewlineAndIndent(indentLevel)
+  escapeJson("tags", result)
+  result.add ": {"
+  if val.all.len > 0:
+    result.addArray("all", val.all, indentLevel + 1)
+  if val.`any`.len > 0:
+    result.addArray("any", val.`any`, indentLevel + 1)
+  if val.`not`.len > 0:
+    result.addArray("not", val.`not`, indentLevel + 1)
+  result.removeComma()
+  result.addNewlineAndIndent(indentLevel)
+  result.add "},"
+
 func addApproach(result: var string; val: ApproachConfig; indentLevel = 1) =
   ## Appends the pretty-printed JSON for an `approach` object with value `val` to
   ## `result`.
@@ -35,6 +49,10 @@ func addApproach(result: var string; val: ApproachConfig; indentLevel = 1) =
   result.addArray("authors", val.authors, indentLevel + 1)
   if val.contributors.len > 0:
     result.addArray("contributors", val.contributors, indentLevel + 1)
+  if val.tags.isSome():
+    let tags = val.tags.get()
+    if tags.all.len + tags.`any`.len + tags.`not`.len > 0:
+      result.addApproachTags(tags, indentLevel + 1)
   result.removeComma()
   result.addNewlineAndIndent(indentLevel)
   result.add "},"
--- a/src/lint/approaches_and_articles.nim
+++ b/src/lint/approaches_and_articles.nim
@@ -30,6 +30,30 @@ proc hasValidIntroduction(data: JsonNode, path: Path): bool =
               "corresponding introduction file at the below location"
     result.setFalseIfFileMissingOrEmpty(introductionPath, msg)
 
+proc hasValidTags(data: JsonNode; path: Path): bool =
+  const k = "tags"
+  if data.hasKey(k):
+    if hasObject(data, k, path):
+      let checks = [
+        hasArrayOfStrings(data[k], "all", path, isRequired = false, uniqueValues = true),
+        hasArrayOfStrings(data[k], "any", path, isRequired = false, uniqueValues = true),
+        hasArrayOfStrings(data[k], "not", path, isRequired = false, uniqueValues = true),
+      ]
+      result = allTrue(checks)
+      if result:
+        var anyAllLen = 0
+
+        if data[k].hasKey("all"):
+          anyAllLen += data[k]["all"].len
+
+        if data[k].hasKey("any"):
+          anyAllLen += data[k]["any"].len
+
+        if anyAllLen == 0:
+          result.setFalseAndPrint("There must be at least one element in the `all` or `any` fields in the `tags` object", path)
+  else:
+    result = true
+
 proc isValidApproachOrArticle(data: JsonNode, context: string,
                               path: Path): bool =
   if isObject(data, context, path):
@@ -41,6 +65,7 @@ proc isValidApproachOrArticle(data: JsonNode, context: string,
       hasArrayOfStrings(data, "authors", path, context, uniqueValues = true),
       hasArrayOfStrings(data, "contributors", path, context,
                         isRequired = false),
+      hasValidTags(data, path),
     ]
     result = allTrue(checks)
     if result:
--- a/src/lint/approaches_and_articles.nim
+++ b/src/lint/approaches_and_articles.nim
@@ -65,7 +65,7 @@ proc isValidApproachOrArticle(data: JsonNode, context: string,
         result.setFalseIfFileMissingOrEmpty(snippetPath, msg)
 
 proc isValidApproach(data: JsonNode, context: string, path: Path): bool =
-  result = isValidApproachOrArticle(data, context, path, checkTags = true)
+  isValidApproachOrArticle(data, context, path, checkTags = true)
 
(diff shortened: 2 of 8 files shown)


src/fmt/approaches.nim
```diff
@@ -35,6 +49,10 @@ func addApproach(result: var string; val: ApproachConfig; indentLevel = 1) =
   result.addArray("authors", val.authors, indentLevel + 1)
   if val.contributors.len > 0:
     result.addArray("contributors", val.contributors, indentLevel + 1)
+  if val.tags.isSome():
+    let tags = val.tags.get()
+    if tags.all.len + tags.`any`.len + tags.`not`.len > 0:
+      result.addApproachTags(tags, indentLevel + 1)
```
ErikSchierboom, line 55: Let me know if you know of a more elegant way to do this:

src/lint/approaches_and_articles.nim
```diff
@@ -41,6 +41,7 @@ proc isValidApproachOrArticle(data: JsonNode, context: string,
       hasArrayOfStrings(data, "authors", path, context, uniqueValues = true),
       hasArrayOfStrings(data, "contributors", path, context,
                         isRequired = false),
+      if checkTags: hasValidTags(data, path) else: true,
```
ErikSchierboom: We currently only support tags for approaches, and not for articles

src/types_approaches_config.nim
```diff
@@ -1,3 +1,5 @@
+import std/options
+import pkg/jsony
```
ee7:
```suggestion
```

Now exported from helpers, which is imported in this file.

src/types_track_config.nim
```diff
@@ -78,10 +78,16 @@ type
     indentSize*: int
     highlightjsLanguage*: string
 
+  ConceptTags* = object
+    all*: seq[string]
+    `any`*: seq[string]
+    `not`*: seq[string]
```
ee7, line 84: The fields here are the same as `ApproachConfigTags`. Should we have a single source of truth?

src/lint/approaches_and_articles.nim
```diff
@@ -63,6 +64,12 @@ proc isValidApproachOrArticle(data: JsonNode, context: string,
                   "there is no corresponding snippet file at the below location"
         result.setFalseIfFileMissingOrEmpty(snippetPath, msg)
 
+proc isValidApproach(data: JsonNode, context: string, path: Path): bool =
+  result = isValidApproachOrArticle(data, context, path, checkTags = true)
```
ee7:
```suggestion
  isValidApproachOrArticle(data, context, path, checkTags = true)
```

Let's avoid `result =` for clear one-liners.

src/lint/approaches_and_articles.nim
```diff
@@ -63,6 +64,12 @@ proc isValidApproachOrArticle(data: JsonNode, context: string,
...
+
+proc isValidArticle(data: JsonNode, context: string, path: Path): bool =
+  result = isValidApproachOrArticle(data, context, path, checkTags = false)
```
ee7:
```suggestion
  isValidApproachOrArticle(data, context, path, checkTags = false)
```

src/types_track_config.nim
ErikSchierboom: We could. What would we call that file?

src/fmt/approaches.nim
ee7: We could iterate over the fields at compile time and generate something equivalent. Something like this (untested): 

```nim
for field in tags.fields:
  if field.len > 0:
    result.addApproachTags(tags, indentLevel + 1)
    break
```

But we didn't do that anywhere else. So let's not do it here. I hope for all this custom serialization mess to be handled much more elegantly in the future.

src/types_track_config.nim
ErikSchierboom: Feel free to extract it to a separate file.

ee7: >  file

I meant having a single (renamed) object for both.

Maybe this?:

- remove the `ConceptTags` definition
- rename remaining `ConceptTags` instances to `AnalyzerTags`
- rename all `ApproachConfigTags` instances to `AnalyzerTags`?

 Or whatever you think of this as conceptually.

ee7: Oh, I see. The definitions were in different files.

ee7: Well, I think the options are:

1. The status quo
2. Add a separate `types_common.nim` 
3. Make `types_track_config` contain:
 
    ```nim
    from types_approaches_config import AnalyzerTags
    ```

Let's just stick with option 1 for now?

ErikSchierboom: I think option 2 is fine. But we can do that as a follow-up 🤷

ErikSchierboom: Let's do it as a follow-up. There might be more types that we can re-use.

src/lint/validators.nim
```diff
@@ -568,6 +568,30 @@ proc hasInteger*(data: JsonNode; key: string; path: Path; context = "";
...
+        var anyAllLen = 0
+
+        if data[k].hasKey("all"):
+          anyAllLen += data[k]["all"].len
+
+        if data[k].hasKey("any"):
+          anyAllLen += data[k]["any"].len
+
+        if anyAllLen == 0:
+          result.setFalseAndPrint("There must be at least one element in the `all` or `any` fields in the `tags` object", path)
```
ee7:
```suggestion
          const msg = "There must be at least one element in the `all` " &
                      "or `any` fields in the `tags` object"
          result.setFalseAndPrint(msg, path)
```

ee7 (approved):
No diff to output of `configlet lint` on any track.

Diff in output of `configlet fmt` on exactly one track:

#### csharp
```diff
-The below paths are relative to '/home/foo/exercism-tracks/csharp'
-Not formatted: config.json
-Not formatted: exercises/practice/bank-account/.approaches/config.json
-Not formatted: exercises/practice/bob/.approaches/config.json
-Not formatted: exercises/practice/collatz-conjecture/.approaches/config.json
-Not formatted: exercises/practice/darts/.approaches/config.json
-Not formatted: exercises/practice/difference-of-squares/.approaches/config.json
-Not formatted: exercises/practice/diffie-hellman/.approaches/config.json
-Not formatted: exercises/practice/gigasecond/.approaches/config.json
-Not formatted: exercises/practice/grade-school/.approaches/config.json
-Not formatted: exercises/practice/grains/.approaches/config.json
-Not formatted: exercises/practice/hamming/.approaches/config.json
-Not formatted: exercises/practice/isogram/.approaches/config.json
-Not formatted: exercises/practice/leap/.approaches/config.json
-Not formatted: exercises/practice/pangram/.approaches/config.json
-Not formatted: exercises/practice/parallel-letter-frequency/.approaches/config.json
-Not formatted: exercises/practice/perfect-numbers/.approaches/config.json
-Not formatted: exercises/practice/phone-number/.approaches/config.json
-Not formatted: exercises/practice/pig-latin/.approaches/config.json
-Not formatted: exercises/practice/protein-translation/.approaches/config.json
-Not formatted: exercises/practice/proverb/.approaches/config.json
-Not formatted: exercises/practice/raindrops/.approaches/config.json
-Not formatted: exercises/practice/reverse-string/.approaches/config.json
-Not formatted: exercises/practice/rna-transcription/.approaches/config.json
-Not formatted: exercises/practice/series/.approaches/config.json
-Not formatted: exercises/practice/sieve/.approaches/config.json
-Not formatted: exercises/practice/two-fer/.approaches/config.json
-Not formatted: exercises/practice/yacht/.approaches/config.json
+Every exercise config is already formatted!

```

src/lint/track_config.nim
```diff
@@ -135,6 +135,7 @@ proc isValidConcept(data: JsonNode; context: string; path: Path): bool =
       hasString(data, "uuid", path, context, checkIsUuid = true),
       hasString(data, "slug", path, context, maxLen = 255, checkIsKebab = true),
       hasString(data, "name", path, context, maxLen = 255),
+      hasValidTags(data, path),
```
ee7: There's a blocking bug here. There is already a proc in this file named `hasValidTags`, which lints the track-level `tags` (like `typing/static`). It also has the same signature!

```console
$ git grep --break --heading 'proc hasValidTags'
src/lint/track_config.nim
268:proc hasValidTags(data: JsonNode; path: Path): bool =

src/lint/validators.nim
571:proc hasValidTags*(data: JsonNode; path: Path): bool =
```

Could you rename the new one in `validators.nim`? `hasValidAnalyzerTags` is fine by me, if that's accurate terminology.

ee7: Nim silently uses the in-file definition:

```console
$ cat a.nim
import b

proc foo =
  echo "a"

foo()
$ cat b.nim
proc foo* =
  echo "b"
$ nim r --hints:off a.nim
/tmp/a.nim(1, 8) Warning: imported and not used: 'b' [UnusedImport]
a
```

This is like variable shadowing, which Nim considers an idiom, but I admit it's much more unfortunate here.

So I believe the the new tags were being linted with the wrong validator.

ErikSchierboom: Good catch!

ee7 (approved):

PR #826, opened 2023-10-06
build: build arm64 macOS binary natively

### LINKED ISSUE #801: build: arm64 macOS executable is built without LTO
Follow-up from:

- https://github.com/exercism/configlet/commit/f280445978c16e9590d3822f5795b1485a64a65f
- https://github.com/exercism/configlet/pull/796#issuecomment-1681845938

ee7:
Apple silicon (M1) macOS runners are [now available in public beta][1].

Closes: #801

[1]: https://github.blog/changelog/2023-10-02-github-actions-apple-silicon-m1-macos-runners-are-now-available-in-public-beta/

---

@ErikSchierboom it looks the new job currently won't run. GitHub [says](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners):

>  GitHub offers customers on GitHub Team and GitHub Enterprise Cloud plans a range of managed virtual machines with more RAM, CPU, and disk space

and

> can now be used by any developer, team, or enterprise! 

I thought that "GitHub Team" was what open-source orgs were. Do you know whether this runner is available for Exercism, or whether there perhaps a setting somewhere that Exercism needs to change to allow using it in this repo?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,6 +34,10 @@ jobs:
             arch: x86-64
 
           - os: macos
+            runs-on: macos-13-xlarge
+            arch: arm64
+
+          - os: macos
             runs-on: macos-12
             arch: x86-64
 
@@ -83,9 +87,6 @@ jobs:
 
           - runs-on: ubuntu-22.04
             zig_target: riscv64-linux-musl
-
-          - runs-on: macos-12
-            zig_target: aarch64-macos-none
 
           - runs-on: ubuntu-22.04
             zig_target: aarch64-windows-gnu


ErikSchierboom:
> I thought that "GitHub Team" was what open-source orgs were. 

GitHub Team is a paid feature, but it is one we _do_ use, so that's fine.

> Do you know whether this runner is available for Exercism, or whether there perhaps a setting somewhere that Exercism needs to change to allow using it in this repo?

I think I would have to enable large runners, but we haven't done that as they have a cost attached. As per the blog post:

> With today’s launch, our macOS larger runners will be priced at $0.16/minute for XL and $0.12/minute for large.

CC @iHiD

ee7:
This job would only run when there is a configlet release. But I think this PR isn't really worth it if we have to pay extra - only if it can use some existing quota of GitHub Actions minutes.

Advantages of this PR:

- Can run tests for arm64 macOS natively at build time
- Can fix the relatively unimportant https://github.com/exercism/configlet/issues/801

But configlet shouldn't have CPU-specific code, so the first point shouldn't matter much. And the latter point will eventually be resolved when the Zig linker gains support for LTO.

ErikSchierboom:
I'm not sure I can only enable this runner for this repo though.

ee7:
From [here](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#included-storage-and-minutes):

> Note: Included minutes cannot be used for larger runners. These runners will always be charged for, including in public repos. 

So I think it's not worth bothering, even if we could do all of these:

- Enable large runners only for configlet
- Make it run for only a few minutes per configlet release
- Add a timeout so it's guaranteed not to rack up a huge bill if something hangs

I think the status quo is good enough. Closing.

PR #835, opened 2023-10-20
lint(validators): refactor linting of analyzer tag

ee7:
Go via strscans, rather than `find` and `substr`.

Follow up from commit b348ff9e3af3.

---

Happy to write like this? I find this a little more readable, though admittedly it requires understanding of strscans.

### DIFF
--- a/src/lint/validators.nim
+++ b/src/lint/validators.nim
@@ -86,21 +86,18 @@ proc hasValidAnalyzerTag(s, key: string; path: Path; context: string;
                          errorAnnotation = ""): bool =
   ## Returns true if `s` is a valid analyzer tag
   result = true
-  let colonIdx = s.find(':')
-  if colonIdx == -1:
-    let msg = &"The {format(context, s)} tag does not match the expected format: \"<category>:<value>\""
-    result.setFalseAndPrint(msg, path, annotation = errorAnnotation)
-  else:
-    let category = s.substr(0, colonIdx - 1)
+  let (isMatch, category, value) = s.scanTuple("$+:$*")
+  if isMatch:
     if category notin analyzerTagCategories:
       let msg = &"The {format(context, s)} tag's category must be one of: " &
                  "`paradigm`, `technique`, `construct` or `uses`"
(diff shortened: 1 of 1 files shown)


src/lint/validators.nim
```diff
@@ -86,21 +86,18 @@ proc hasValidAnalyzerTag(s, key: string; path: Path; context: string;
...
+  let (isMatch, category, value) = s.scanTuple("$+:$*")
+  if isMatch:
     if category notin analyzerTagCategories:
       let msg = &"The {format(context, s)} tag's category must be one of: " &
                  "`paradigm`, `technique`, `construct` or `uses`"
       result.setFalseAndPrint(msg, path, annotation = errorAnnotation)
-
-    let value = s.substr(colonIdx + 1)
     if value.isEmptyOrWhitespace:
       let msg = &"The {format(context, s)} tag's value is a whitespace-only string"
```
ee7, line 96: Maybe we should write "empty or whitespace-only"?

src/lint/validators.nim
```diff
@@ -86,21 +86,18 @@ proc hasValidAnalyzerTag(s, key: string; path: Path; context: string;
                          errorAnnotation = ""): bool =
   ## Returns true if `s` is a valid analyzer tag
   result = true
-  let colonIdx = s.find(':')
-  if colonIdx == -1:
-    let msg = &"The {format(context, s)} tag does not match the expected format: \"<category>:<value>\""
-    result.setFalseAndPrint(msg, path, annotation = errorAnnotation)
-  else:
-    let category = s.substr(0, colonIdx - 1)
+  let (isMatch, category, value) = s.scanTuple("$+:$*")
```
ee7:
```suggestion
  let (isMatch, category, value) = s.scanTuple("$+:$+")
```

src/lint/validators.nim
ee7: I think https://github.com/exercism/configlet/pull/835/commits/8c3354e1a969c6a33201ecd0d6b6a86bb2ef19b9 is better.

ErikSchierboom (approved):
A lot nicer! Thanks

PR #845, opened 2023-12-19
create: allow creating of exercise

ErikSchierboom:
- Setup basic structure
- initial work to update track config
- Create exercise

Refs: https://github.com/exercism/configlet/issues/675

PR #850, opened 2024-01-11
sync: only fetch prob-specs repo once

ErikSchierboom:


ee7:
Can we close this PR iin favor of #845? I think I'd prefer just to include it in that one - it's not logically separate, right?

ErikSchierboom:
> it's not logically separate, right?

Well, depends on your viewpoint. But I'm fine with closing it here.

ee7:
I think it's something that could be a separate commit, I'd just prefer that it not be a separate commit directly on the main branch. Rationale: theoretically, if we wanted to revert the commit that adds the `create` functionality, maybe we should revert this change too (because it only exists to prepare for the upcoming implementation of `create`).

If you want to have it as a separate commit in the history on `main`, I'd argue that it should be a separate commit in a branch for `create` that's merged via merge commit. But I wouldn't bother.

PR #852, opened 2024-01-16
refactor: use `QuitSuccess` and `QuitFailure` in `quit` calls

ErikSchierboom:


ee7 (approved):
LGTM, but I haven't checked that this is all of them.

nit: I'd probably omit the parens here. The often-preceding `echo` and `writeLine` calls do that too.

ErikSchierboom:
I've updated

PR #864, opened 2024-04-22
create: allow specifying `author` and `difficulty`

ErikSchierboom:
Support specifying the author and/or difficulty when creating exercises.

CC @glennj

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Options for create:
   -e, --exercise <slug>           Only operate on this exercise
   -o, --offline                   Do not update the cached 'problem-specifications' data
   -a, --author                    The author of the exercise, approach or article
-  -d, --difficult                 The difficulty of the exercise (default: 1)
+  -d, --difficulty                The difficulty of the exercise (default: 1)
 
 Options for fmt:
   -e, --exercise <slug>           Only operate on this exercise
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -75,6 +75,8 @@ Options for create:
       --concept-exercise <slug>   The slug of the concept exercise
   -e, --exercise <slug>           Only operate on this exercise
   -o, --offline                   Do not update the cached 'problem-specifications' data
+  -a, --author                    The author of the exercise, approach or article
```
ErikSchierboom, line 78: This is deliberately singular, as 99.99% of the time there is only one author.

src/create/create.nim
```diff
@@ -13,6 +13,10 @@ proc create*(conf: Conf) =
         let msg = &"Both --approach and --article were provided. Please specify only one."
         stderr.writeLine msg
         quit QuitFailure
+      if conf.action.difficulty.isSome:
+        let msg = "The difficulty argument is not supported for approaches"
+        stderr.writeLine msg
+        quit QuitFailure
```
glennj, line 19: Does this need to be a failure, or is a warning sufficient?

glennj (approved):

src/create/create.nim
ErikSchierboom: We tend to error for arguments that are not supported.

PR #879, opened 2024-07-10
.github: bump macos-12 to macos-13

ee7:
This is the newer x86_64 macOS image, and is not in beta. See [here](https://github.com/actions/runner-images/blob/1b535372a075188bfce11fbd27d72a254c736b3c/README.md#available-images).

It would now be possible to use `macos-14` somewhere, building a macOS arm64 configlet asset natively rather than cross-compiling via Zig. But let's try to avoid that for now.

---

To-do:

- [ ] Resolve error in the build workflow, in particular the cross-compile job for `aarch64-macos-none`

```text
CC: ../nimdir/lib/std/sysrand.nim
/Users/runner/.cache/nim/configlet_r/@m..@snimdir@slib@sstd@ssysrand.nim.c:9:10: fatal error: 'Security/SecRandom.h' file not found
nimble.nim(304)          buildFromDir
    9 | #include <Security/SecRandom.h>

      |          ^~~~~~~~~~~~~~~~~~~~~~
    Error:  Build failed for the package: configlet
1 error generated.
     Info:  Nimble data file "/Users/runner/.nimble/nimbledata2.json" has been saved.
Error: execution of an external compiler program 'zigcc -c -w -ferror-limit=3 -pthread -target aarch64-macos-none -Os   -I/Users/runner/work/exercism-configlet/exercism-configlet/nimdir/lib -I/Users/runner/work/exercism-configlet/exercism-configlet/src -o /Users/runner/.cache/nim/configlet_r/@m..@snimdir@slib@sstd@ssysrand.nim.c.o /Users/runner/.cache/nim/configlet_r/@m..@snimdir@slib@sstd@ssysrand.nim.c' failed with exit code: 1
```

I think it just needs to pass to the linker:

```text
-framework Security
```

PR #890, opened 2024-09-17
fix: more occurrences of macos12

vaeng:
Missed to files. Thanks @glennj 

> @vaeng Did you miss one? The merge into main is still using v12: https://github.com/exercism/configlet/actions/runs/10903920485/job/30259101645 image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,7 +34,7 @@ jobs:
             arch: x86-64
 
           - os: macos
-            runs-on: macos-12
+            runs-on: macos-14
             arch: x86-64
 
           - os: windows
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ jobs:
             arch: x86-64
 
           - os: macos
-            runs-on: macos-12
+            runs-on: macos-14
             arch: x86-64
 
           - os: windows


vaeng:
Something is up with the nim installation on macOS and something else is happening with the windows tests in the nim scripts oO

ErikSchierboom:
You should just rerun the Windows test. It's flaky

vaeng:
> You should just rerun the Windows test. It's flaky

Thanks, I will remember that. The MacOS tests are still not happy though

ee7:
Thanks, but I think we don't want the changes in this PR. The change from `macos-12` to `macos-14` is a change of architecture from `x86_64` to `arm64`, which means:

- The test workflow fails because Nim doesn't provide prebuilt binaries for arm64 macOS.
- The configlet release process will either fail when producing a x86_64-macOS binary, or produce an arm64 macOS binary that is labelled as x86_64-macOS.

Furthermore, it is likely that commit https://github.com/exercism/configlet/commit/22c4cec3d16d155afef664124003a9026b014264 broke the `build.yml` job, which runs at release time (and not at PR time). Changes to that workflow should be tested before being merged. And if we'll build on arm64-macos, we might as well build natively rather than using the cross-compiling machinery.

I think we should close this in favor of https://github.com/exercism/configlet/pull/879. That is, we should move to the non-deprecated macOS x86_64 runner first.

vaeng:
> Thanks, but I think we don't want the changes in this PR. The change from `macos-12` to `macos-14` is a change of architecture from `x86_64` to `arm64`, which means:
> 
>     * The test workflow fails because Nim doesn't provide prebuilt binaries for arm64 macOS.
> 
>     * The configlet release process will either fail when producing a x86_64-macOS binary, or produce an arm64 macOS binary that is labelled as x86_64-macOS.
> 
> 
> Furthermore, it is likely that commit [22c4cec](https://github.com/exercism/configlet/commit/22c4cec3d16d155afef664124003a9026b014264) broke the `build.yml` job, which runs at release time (and not at PR time). Changes to that workflow should be tested before being merged. And if we'll build on arm64-macos, we might as well build natively rather than using the cross-compiling machinery.
> 
> I think we should close this in favor of #879. That is, we should move to the non-deprecated macOS x86_64 runner first.

I did not consider the architecture change in the different os releases, thank you for that pointer. We should probably roll the changes back first and then move on with your suggestion.

glennj:
My apologies for stepping in without knowing the history and build status of this repo.

ErikSchierboom:
For the reasons mentioned by @ee7, I'm closing this.

I've also created https://github.com/exercism/configlet/pull/899 which fixes the build: https://github.com/exercism/configlet/actions/runs/11812866423

PR #907, opened 2025-07-27
macos-15

keiravillekode:
@ee7 Could you look at this?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,8 +34,7 @@ jobs:
             arch: x86-64
 
           - os: macos
-            runs-on: macos-12
-            arch: x86-64
+            runs-on: macos-15
 
           - os: windows
             runs-on: windows-2022
@@ -84,7 +83,7 @@ jobs:
           - runs-on: ubuntu-22.04
             zig_target: riscv64-linux-musl
 
-          - runs-on: macos-12
+          - runs-on: macos-15
             zig_target: aarch64-macos-none
 
           - runs-on: ubuntu-22.04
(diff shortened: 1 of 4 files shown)


ee7:
> @ee7 Could you look at this?

Hey @ErikSchierboom, sure.

Bumping `macos-12` to `macos-15` changes the architecture from x86_64 to arm64. So I believe this PR is not correct right now, because:

- It would cause a future `configlet_4.x.y_macos_x86-64.tar.gz` release asset to incorrectly contain an arm64 macOS binary, rather than an x86_64 one.
- It might introduce the same error in the build workflow (which does not automatically run in CI on a PR) that ended up blocking https://github.com/exercism/configlet/pull/879.
- Less importantly: it uses the cross-compilation process to build the arm64 macOS release, even though with this PR that release is being built on arm64 macOS.

ErikSchierboom:
I thought so as well. Thanks for looking into it.

ee7:
~I'll try to fix https://github.com/exercism/configlet/pull/879 now that GitHub provides native aarch64 macOS runners.~

Related: https://github.com/exercism/configlet/pull/890

---

Edit: Actually, instead I think we should do it like this new PR: https://github.com/exercism/configlet/pull/908. I should be able to finish that PR later today.

ee7:
I think https://github.com/exercism/configlet/pull/908 should be OK now - I did the usual test of the build workflow by pushing a development tag for that PR's state to the repo. It'd be good if someone could check the now natively built arm64 macOS asset.

Changes that affect the build workflow are currently more tricky, because that workflow is currently only run as part of the release process - it doesn't automatically run in CI on PRs. But getting rid of the `macos-12` in this repo is indeed long overdue, so thanks @keiravillekode.

ee7:
Closing in favor of #908, which was merged.

PR #928, opened 2026-04-09
fix: make compatible with v5 and v7

vaeng:
I had problems using the existing script with windows 11 and powershell 5. I updated my ps version to 7.6 and still had errors.

As I don't know powershell, I used AI to finda  solution and this seems to work. I am worried about backwards comatibility and general AI use for things I don't understand.

Probably best to let knowledgeable people judge this.

### DIFF
--- a/scripts/fetch-configlet.ps1
+++ b/scripts/fetch-configlet.ps1
@@ -6,16 +6,32 @@
 $ErrorActionPreference = "Stop"
 $ProgressPreference = "SilentlyContinue"
 
-$requestOpts = @{
-    Headers           = If ($env:GITHUB_TOKEN) { @{ Authorization = "Bearer ${env:GITHUB_TOKEN}" } } Else { @{ } }
-    MaximumRetryCount = 3
-    RetryIntervalSec  = 1
+$headers = If ($env:GITHUB_TOKEN) { @{ Authorization = "Bearer ${env:GITHUB_TOKEN}" } } Else { @{ } }
+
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
I'm not really that knowledgeable about PowerShell. Maybe @exercism/powershell could take a look?

scripts/fetch-configlet.ps1
```diff
@@ -31,11 +47,16 @@ Write-Output "Fetching configlet..."
...
 
 $configletPath = Join-Path -Path $outputDirectory -ChildPath "configlet.exe"
 if (Test-Path -Path $configletPath) { Remove-Item -Path $configletPath }
-[System.IO.Compression.ZipFile]::ExtractToDirectory($outputPath, $outputDirectory)
+if (Get-Command Expand-Archive -ErrorAction SilentlyContinue) {
+    Expand-Archive -Path $outputPath -DestinationPath $outputDirectory -Force
+} else {
+    Add-Type -AssemblyName System.IO.Compression.FileSystem
+    [System.IO.Compression.ZipFile]::ExtractToDirectory($outputPath, $outputDirectory)
+}
```
ErikSchierboom, line 59: This looks like it silently fails when `Expand-Archive` cannot be found, in which case configlet cannot be extracted. I don't think that's what we want?

ErikSchierboom:
> I updated my ps version to 7.6 and still had errors.

That's weird. What was the error?

PR #934, opened 2026-06-10
Preserve newly added editor and invalidator keys when using `configlet create`

BNAndras:
Related to https://forum.exercism.org/t/configlet-create-doesnt-add-invalidator-or-editor-to-exercise-config-json/.

Currently, `configlet create --practice-exercise <slug>` does not populate the editor and invalidator file paths in the exercise's config.json when the patterns are set in a track's root-level config file. That invariably requires a follow-up `configlet sync --filepaths -uye <slug>` to add the missing entries.


IsaacG (approved):

glennj:
Any new tests for this change?

BNAndras:
Yeah, it's definitely worth adding one. I'll have to poke around and see how to best represent this scenario in test form.

ErikSchierboom (approved):

PR #936, opened 2026-06-30
specification for practice exercises

keiravillekode:
Tracks may need to assign a practice exercise a different slug than the one used in problem-specifications.

We introduce the optional `specification` key.

https://forum.exercism.org/t/factor-track-feedback/60226/19

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,10 @@ Therefore configlet has a `sync` command, which can check that such Practice Exe
 There are three kinds of data that configlet can update from `problem-specifications`: documentation, metadata, and tests.
 There is also one kind of data that configlet can populate from the track-level `config.json` file: filepaths in exercise config files.
 
+By default, configlet looks up the upstream exercise in `problem-specifications` using the Practice Exercise's `slug`.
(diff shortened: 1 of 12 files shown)


IsaacG (approved):
+cc @ErikSchierboom

BNAndras:
This PR was brought after limited discussion in a single track-specific thread.  I'm against this change in the current form.

What is this trying to solve?
Is it a common issue for other threads or just a Factor-specific problem?
What are the ramifications of this change?
Exercism uses exercise slugs in a few locations across the site so what happens if we're now saying `binary-search` and `binary-chop` are the same exercise?
Will this optional key also be documented in the Exercism docs about the track `config.json`?

IsaacG:
> What is this trying to solve?

Remaining an exercise while allowing the configlet to continue to sync it. 

> Is it a common issue for other threads or just a Factor-specific problem?

The haskell track has three exercises that cannot test due to the same issue. (Matrix, binary search and another.)

> What are the ramifications of this change? Exercism uses exercise slugs in a few locations across the site so what happens if we're now saying `binary-search` and `binary-chop` are the same exercise?

The renamed exercises would largely look like track specific exercises to all other tooling. They would be treated like track specific exercises (which are supported) ... but syncing would magically still work. 

> Will this optional key also be documented in the Exercism docs about the track `config.json`?

Hopefully 😁

